PR #12, opened 2019-04-15
add angular devcontainer

bowdenk7:


repository-containers/github.com/angular/angular/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,26 @@
+#-----------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root for license information.
+#-----------------------------------------------------------------------------------------
+
+FROM openjdk:7
```
jamiebarrow, line 6: I was wondering if there was any specific reason to use this as the base image, and not some kind of NodeJS base image  - for example, as in https://github.com/microsoft/vscode-dev-containers/blob/master/containers/javascript-node/.devcontainer/base.Dockerfile?

Chuxel:
Angular actually has its own definition now, so I suspect we should just remove this one. 

https://github.com/angular/angular/tree/master/.devcontainer

jamiebarrow:
@Chuxel thanks for sharing :)

PR #30, opened 2019-05-04
[java] Fix symlink and upgrade to Gradle 5.4.1

brunoborges:
This PR fixes the wrong symlink to the Gradle binary and upgrades to version 5.4.1

### DIFF
--- a/containers/java-11/.devcontainer/Dockerfile
+++ b/containers/java-11/.devcontainer/Dockerfile
@@ -18,14 +18,14 @@ RUN apt-get -y install git procps curl
 
 # Install Gradle
 ENV GRADLE_HOME /opt/gradle
-ENV GRADLE_VERSION 5.4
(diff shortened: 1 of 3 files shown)


Chuxel:
@chrmarti I think this is safe to merge at this point - do you still want to hold off given its a bug fix?

Chuxel:
@brunoborges @testforstephen @testforstephen Actually I just retried and the SHA check failed.  Do we need that given this will be a moving target?

brunoborges:
@Chuxel I've updated the hashes. Pls try again?

Chuxel:
LGTM! @chrmarti - LMK if you want to merge the fix.

chrmarti:
Bug fixes are good to merge. We will follow the same process as with VS Code at this point. That includes a review (can be yourself if the PR is from the community) and verification of the fixes before we publish a new version (we can download the vsix from the build).

Issue #33, opened 2019-05-06
Kubernetes missing certificate


Shikanime:
- VSCode Version: 1.34.0-insider (1.34.0-insider)
- Local OS Version: macOS Mojare 10.14.4 (18E226)
- Name of Dev Container Definition with Issue: Kubernetes & Helm

Steps to Reproduce:

1. Click on "Reopen on Container"
2. New terminal

Context:

Minikube certificates are not pull and paths are not replaced in the kube configuration.

```
- name: minikube
  user:
    client-certificate: /Users/USER_NAME/.minikube/client.crt
    client-key: /Users/USER_NAME/.minikube/client.key
```

Chuxel:
@Shikanime Thanks so much for the PR on this - it is a good catch!  Let's continue the conversation on the PR.

PR #35, opened 2019-05-06
Fix kubernete-helm certificates

Shikanime:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

Bugfix

## What is the current behavior?

Fixes: #33

## What is the new behavior?

Fix kubernetes-helm missing certificate.

## Does this PR introduce a breaking change?

No


Shikanime:
@Chuxel Sorry about that, I found something new in the documentation. It seems that since Docker 17.06, the recommended way to bind a volume is to use the option `--mount` which as far i tried it seem to work on Windows even if the folder doesn't exist (which is a strange behavior).

Chuxel:
@Shikanime Unfortunately, I'm seeing the same error (this is from my Mac, but same problem as Windows). For bind mounts, this makes sense given the Docker docs say this won't create the folder:

> If you use --mount to bind-mount a file or directory that does not yet exist on the Docker host, Docker does not automatically create it for you, but generates an error.

https://docs.docker.com/storage/bind-mounts/

That said, switching to "-v" actually did work on my Mac.  Let me retest on my desktop back at work in the morning.  It's possible that is Windows specific or something strange was going on.

Shikanime:
Oh nvm, I just saw your error on Windows where the driver was not shared on Docker.
```
Error response from daemon: Mount denied
```

Chuxel:
Yeah, unfortunately this still happens on Windows (and volume mounting is working).  I actually wonder if it is a bug with Docker for Windows.  I'm not finding anything but it is strange.

Shikanime:
@Chuxel Did you retried with the last commit where I reverted to the `-v` option for minikube.

Chuxel:
@Shikanime Yeah I just tried it on another machine to make sure it wasn't my config -- it's must be a Docker for Windows thing.  You can repro it just using the docker cli:

Fails:
```
docker run --rm -v c:/Users/doesnotexist:/data alpine ls /data
```

Works:
```
docker run --rm -v c:/Users:/data alpine ls /data
```

Chuxel:
@chrmarti On this one, one thing that strikes me is that the challenge here is primarily on Linux (though obviously people can install minikube on Windows / macOS too).  We'd talked about platform specific config in devcontainer.json.  Am I correct in remembering that we didn't end up integrating this?

If we had it, we could always include the addition to bashrc to copy the minikube contents, but set it to false on Windows/Mac and not include the volume mount by default.  On Linux it would then be true with the volume mount.  We can then doc how to enable it on the other two platforms.

chrmarti:
@Chuxel We didn't add the platform overrides. It looks like the code is (or could be) written such that it does not break when minikube is not installed.

Chuxel:
@chrmarti The problem is that on Windows, if the folder you are trying to mount doesn't exist, it fails.  So, if there's no `.minikube` folder (like when you use Kubernetes in Docker Desktop), it errors.

chrmarti:
@Chuxel Then maybe comment the mount option for `.minikube` with a comment explaining when it should be used?

Chuxel:
@chrmarti @Shikanime Yeah that was kind of where I was thinking we could land as well.  We could have the following in `devcontainer.json`:

```json

    "runArgs": [
        "-v", "/var/run/docker.sock:/var/run/docker.sock",
        "-v", "${env:HOME}${env:USERPROFILE}/.kube:/root/.kube-localhost",
        
        // Uncomment the next line to also sync certs in your .minikube folder
        // "-v", "${env:HOME}${env:USERPROFILE}/.minikube:/root/.minikube-localhost",

        // Uncomment the next line if you will use a ptrace-based debugger like C++, Go, and Rust.
        // "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined"
    ],
```

We can then update the README to mention it.

Shikanime:
@Chuxel LGTM, I will update the documentation accordingly.

Shikanime:
All done 🥳

Chuxel:
LGTM!! Thanks for the contribution again!

PR #36, opened 2019-05-07
Added security configs for the docker-in-docker-compose example

AntonNguyen:
While trying to get this to work for my own project, I realized most of the examples set the security permissions in `devcontainer.json`:
```
	"runArgs": [
		"--cap-add=SYS_PTRACE",
		"--security-opt",
		"seccomp=unconfined"
	]
```

When overriding docker-compose, the `runArgs` field is ignored, preventing vscode's debugger from working. 

To get over this limitation, we need to define the security permissions in the service in the docker-compose file.

I've updated the docs here: https://github.com/microsoft/vscode-docs/pull/2633

### DIFF
--- a/containers/docker-in-docker-compose/.devcontainer/docker-compose.yml
+++ b/containers/docker-in-docker-compose/.devcontainer/docker-compose.yml
@@ -6,7 +6,7 @@
 version: '3'
 services:
   docker-in-docker:
-    build: 
+    build:
       context: .
       dockerfile: Dockerfile
     volumes:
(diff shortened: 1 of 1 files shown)


Chuxel:
@AntonNguyen This is a good suggestion! But, as I mentioned on the docs PR, this is only required for C++, Go, and Rust which is why it's not included everywhere. That said, I think if the comment before this is updated to specifically say this is optional and only required for certain languages, we can likely include it without worrying anyone.

Beyond this one, there's a number of others that likely should include this as an optional reference (docker-in-docker non-compose, kubernetes-helm, existing dockerfile / docker compose, possibly templates).  I can PR those updates separately.

PR #40, opened 2019-05-08
Highlight ptrace settings via commented out snippets in certain definitions 

Chuxel:
Revised version of https://github.com/microsoft/vscode-dev-containers/pull/36

This does not directly add the different ptrace related settings required for Rust, C++, and Go, but includes them as comments in `devcontainer.json` and `docker-compose.yml` in several definitions that likely are going to be used as a base to build from.

//cc: @AntonNguyen


container-templates/docker-compose/.devcontainer/docker-compose.yml
```diff
@@ -23,6 +23,15 @@ services:
...
+      # Uncomment the next line if you plan to install the Docker CLI. See the docker-in-docker-compose definition for details.
+      # - /var/run/docker.sock:/var/run/docker.sock 
+
+    # Uncomment the next four lines if you will use a ptrace-based debugger like C++, Go, and Rust.
+    # cap_add:
+    #   - SYS_PTRACE
+    # security_opt:
+    #   - seccomp:unconfined
+
+    # Overrides default command so things don't shut down after the process ends - useful for debugging.
```
chrmarti: I'd drop 'useful for debugging', the container must not shutdown for any development activity.

container-templates/docker-compose/.devcontainer/devcontainer.json
```diff
@@ -1,11 +1,21 @@
...
+	"service": "your-service-name-here",
+
+	// The optional 'workspaceFolder' property is the path VS Code should open by default when
+	// connected. This is typically a volume mount in .devcontainer/docker-compose.yml
+	"workspaceFolder": "/workspace",
+
+	// Add any extensions you want auto-installed here.
+	"extensions": [],
+
+	// Update the value of this line like to 'none' if you want to keep your containers running after VS Code shuts down.
```
chrmarti: Drop 'like'?

container-templates/dockerfile/.devcontainer/devcontainer.json
```diff
@@ -1,10 +1,23 @@
...
-        "mutantdino.resourcemonitor"
-    ],
-    "runArgs": []
+	// See https://aka.ms/vscode-remote/devcontainer.json for format details.
+	"name": "Dev Container Definition Template - Dockerfile",
+
+	// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
+	"dockerFile": "Dockerfile",
+
+	// Use the optional 'appPort' property to expose ports not already in your Dockerfile
```
chrmarti: We don't automatically expose ports `EXPOSE`d in the Dockerfile yet. So this is promising a little too much.

containers/docker-in-docker-compose/.devcontainer/docker-compose.yml
```diff
@@ -9,17 +9,23 @@ services:
...
     volumes:
       # Update this to wherever you want VS Code to mount the folder of your project
       - ..:/workspace
 
       # This lets you avoid setting up Git again in the container
       - ~/.gitconfig:/root/.gitconfig
 
-      # Forwarding the socket is optional, but lets docker work inside the container if you install the Docker CLI.
-      # See the docker-in-docker-compose definition for details on how to install it.
+      # Forwards the Docker socket from one side ot the other
```
chrmarti: ot > to

containers/docker-existing-dockerfile/.devcontainer/devcontainer.json
```diff
@@ -1,22 +1,26 @@
 {
-    // "See https://aka.ms/vscode-remote/devcontainer.json for format details."
-    "name": "Existing Dockerfile",
+	// See https://aka.ms/vscode-remote/devcontainer.json for format details.
```
chrmarti: We should probably update the automatically added comment to this and we need to make sure we preserve the comments when applying a definition. (I can do that.)

Chuxel:
@chrmarti Made those edits - Thanks! While I was in there, I also raised visibility to appPort and postCreateCommand since I've noticed on vscode-remote-release that people aren't always finding them.  They're now both commented out in each devcontainer.json with a note on what they are for to try to help with that problem. 

See what you think.

chrmarti:
@Chuxel Looks good. I wonder if `postCreateCommand` would benefit from coming with an example. Either on the existing `postCreateCommand` line or on an additional line above that.

Chuxel:
@chrmarti That's a good idea - some of the definitions like node would make that easy. Others the command could just print out the version info for the runtime.  I'll still leave it commented out but it would anchor the idea. I make that edit.

container-templates/docker-compose/.devcontainer/devcontainer.json
```diff
@@ -1,11 +1,25 @@
...
+	"workspaceFolder": "/workspace",
+
+	// Add any extensions you want auto-installed here.
+	"extensions": [],
+
+	// Update the value of this line to 'none' if you want to keep your containers running after VS Code shuts down.
+	"shutdownAction": "stopCompose",
+
+	// Uncomment the next line to run commands after the container is created.
+	// "postCreateCommand": ["uname -a"]
```
chrmarti: @Chuxel `["uname -a"]` won't work. You can use `"uname -a"` or `["uname", "-a"]`. This is similar to `CMD`, `RUN` and others in the Dockerfile. The single string will be parsed by a shell and then executed, the array form is executed directly, without a shell.

PR #44, opened 2019-05-17
Tweaks to Dart container

DanTup:
This makes some tweaks to the Dart container:

- Removes use of pedantic/analysis_options since it wasn't adding much but was causing lint errors at startup (due to the package not being restored before analysis)
- Removes other unused dependencies (globally activated webdev, build_runner, build_web_compiler)
- Removes commands for `pub get` (the Dart extension already has a command for this, and runs it on pubspec changes) and the instructions about it
- Makes the browser open automatically, and removes the need to manually forward port
- Removes unused Chrome extension
- Move script into `bin/` folder (which is the usual location for scripts like this in Dart) and unignore `bin/` for the Dart folder

Based on the history, I think some of the above was because the original intention was for this to be a proper Dart web app, but it ended up becoming just a command-line app that has a HTTP listener.

I'll have a go at building a dart-web container soon too, though I need to try and resolve (or workaround) a bug that results in red squiggles not disappearing when new packages are downloaded after initial analysis (https://github.com/dart-lang/sdk/issues/30845) or need to be able to run `pub get` prior to the package opening (which I think needs https://github.com/microsoft/vscode-remote-release/issues/349).


DanTup:
Great! If you have any issues with it, feel free to ping me.

I've got a Flutter one too - however it's a bit limited because you can't connect to any device or emulator (though you can run tests on the headless device :-)). It might be confusing to merge like this, so I'll do some digging to see if I can find a way to make it work (without requiring things like the Android SDK on the host!) better.

PR #57, opened 2019-05-25
add Sinatra Container

Amblizer:
Add a Ruby Sinatra container with a test project.


Chuxel:
@Amblizer Thanks so much for the PR!  Apologies for the delay looking at it.  I missed it over the long weekend.  

Two things:
1. The only thing I see missing is the `.npmignore` file in the root of the folder.  This is used to tell VS Code what to exclude when it adds the files 
2. Is port 4567 always used for Sinatra applications or is it just for the test project?  If it's just the test project, we've been leaving the appPort property commented out, and then telling people to use the "Forward Port from Container..." command to test.

See the [Node LTS definition](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/javascript-node-lts) for an example that includes a web based test project - you can probably just "cut-and-paste". 😊

Amblizer:
Hi @Chuxel , thank you so much for the indication! I added the `.npmignore` file. The port 4567 is the default port for Sinatra, and I added a comment telling people to change it if they want to use other ones. 

Have a great weekend!

Chuxel:
LGTM!  Thanks so much for the contribution.

PR #60, opened 2019-06-01
Add R devcontainer for statistical computing

kmehant:
# Context
We can have `R devcontainer` added to the present vs code development container ecosystem.

### DIFF
--- a/containers/R/.devcontainer/Dockerfile
+++ b/containers/R/.devcontainer/Dockerfile
@@ -1,0 +1,18 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+FROM r-base:3.6.0
+
+# Install git, process tools, lsb-release (common in install instructions for CLIs) and GPG
+RUN apt-get update && apt-get -y install git procps lsb-release
+
+
+# Clean up
+RUN apt-get autoremove -y \
+    && apt-get clean -y \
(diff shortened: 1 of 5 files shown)


kmehant:
Hello @Chuxel 
We are missing a debugging extension for R so how should we be proceeding with `launch.json` and the `devcontainer`  itself.

Chuxel:
@kmehant Thanks for the PR!  A few comments:

1. One of these extensions appears to require .NET Core which needs to be installed in the `Dockerfile`. Unfortunately, the R image uses the unreleased Debian buster which makes that tricky to do.
2. Is there a reason to install both extensions?

kmehant:
> @kmehant Thanks for the PR! A few comments:
> 
> 1. One of these extensions appears to require .NET Core which needs to be installed in the `Dockerfile`. Unfortunately, the R image uses the unreleased Debian buster which makes that tricky to do.
> 2. Is there a reason to install both extensions?

@Chuxel Thanks for the valuable review! Let me remove **R Tools** extension for now as the other extension satisfies the basic needs (syntax, snippet and run) for programming in R.

Chuxel:
> Thanks for the valuable review! Let me remove R Tools extension for now as the other extension satisfies the basic needs (syntax, snippet and run) for programming in R and will update the readme more aptly to the development 😄

I expect people will want to use the other extension give it provides more features, so ideally we'd figure out how to get .NET Core into the container so it can be used.

kmehant:
@Chuxel I have used **R base image** which is built on top of **Ubuntu:bionic** and installed **.NET core runtime 2.1** and **libzip** as dependencies for **R Tools** extension.

kmehant:
@Chuxel  Can we get this merged 😃?

Chuxel:
LGTM!

Issue #14, opened 2019-04-17
Create devcontainers for popular repos


bowdenk7:
This issue is meant to be a hit list of repos that we want to have devcontainers for and any known issues associated with getting those setup. 

For the most part these are the most forked repos on GitHub as well as some other notable projects.

| Repo | Owner  | Done | Tested | Known Issues  |
|---|---|---|---|---|
| [Angular](https://github.com/angular/angular) | @bowdenk7 | Yes  | No  |   |
| [Bootstrap](https://github.com/twbs/bootstrap)  | @bowdenk7 |  No |   | Getting a headless browser setup in docker is hard |
| [three.js](https://github.com/mrdoob/three.js)  |   |   |   |
| [jekyll-now](https://github.com/barryclark/jekyll-now) | @mjbvz | Yes  |  |
| [TensorFlow-Examples](https://github.com/aymericdamien/TensorFlow-Examples)| @mjbvz | Yes  |  |  |
| [Learn OpenCV](https://github.com/spmallick/learnopencv)| @mjbvz | Yes |  |  |
| [TensorFlow-Tutorials](https://github.com/Hvass-Labs/TensorFlow-Tutorials)| |  |  | Requires jupyter notebooks  |
| [Jekyll](https://github.com/jekyll/jekyll)| @rebornix |  |  |  |
| [D3](https://github.com/d3/d3) |  |  |  |
| [Vue](https://github.com/vuejs/vue) |  |  |  |
| [Spark](https://github.com/apache/spark) |  |  |  | 
| [Animate.css](https://github.com/daneden/animate.css) |  |  |  | 
| [Lantern](https://github.com/getlantern/lantern) |  |  |  | 
| [Golang](https://github.com/golang/go) |  |  |  |  
| [Darknet](https://github.com/pjreddie/darknet) |  |  |  | 
| [Flask](https://github.com/pallets/flask) |  |  |  | 
| [ElasticSearch](https://github.com/elastic/elasticsearch) |  |  |  | 
| [Swift](https://github.com/apple/swift) |  |  |  | 
| [ASP.NET Core](https://github.com/aspnet/AspNetCore) |  |  |  | 
| [Redis](https://github.com/antirez/redis) |  |  |  | 
| [Ghost](https://github.com/TryGhost/Ghost) |  |  |  |

bhack:
I want to add official tensorflow/addon devcontainer cause it is the main official entry point to extend Tensorflow for community contributions. 
[...]

PR #63, opened 2019-06-01
Adding React devcontainer for creating react apps

kmehant:
### Context
Adding React `devcontainer` for creating react apps in Linux containers with `vscode remote`


kmehant:
This **PR** can be part of these #14 accomplishments.
@Chuxel @egamma @qubitron I wish this **PR** gets reviewed 😄

Chuxel:
This one feels a bit too close to other node.js based definitions already in the repository.   These are also intended to be used with existing projects, so calling the create-react-app CLI would actually cause problems.  Thanks for the contribution though!

PR #72, opened 2019-06-08
Added azure-blockchain dev container

cmendible:
Added azure-blockchain dev container

Hope it helps!

### DIFF
--- a/containers/azure-blockchain/.devcontainer/Dockerfile
+++ b/containers/azure-blockchain/.devcontainer/Dockerfile
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 
-FROM node:lts
+FROM python:2.7
 
 # Configure apt
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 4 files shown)


containers/azure-blockchain/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN npm install -g ganache-cli
+
+# Install the Azure CLI
+RUN apt-get install -y apt-transport-https curl gnupg2 lsb-release \
+    && echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $(lsb_release -cs) main" > /etc/apt/sources.list.d/azure-cli.list \
+    && curl -sL https://packages.microsoft.com/keys/microsoft.asc | apt-key add - 2>/dev/null \
+    && apt-get update \
+    && apt-get install -y azure-cli
+	
+# Install Python 2.7.15
```
Chuxel: So, Python 2.7.13 will already on disk at this point.  Do you know if 2.7.15 specifically is required?  If so, I suspect starting from python and then installing Node 10 would be faster since it's smaller e.g.:  

```Dockerfile
FROM python:2.7

RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - \
         && apt-get install -y nodejs
```

cmendible: Hi thanks for pointing this out. Python 2.7.15 is required by the extension. I'll try the suggested change and come back to you  ASAP.

cmendible: Fixed in 4819e6e05cd552368c760dc34418c46965c8d32c

Chuxel:
LGTM!  Thanks.

PR #73, opened 2019-06-09
Mention dependency of app service over db service

kmehant:
A simple fix in docker-compose configuration file.


Chuxel:
Good catch!!

PR #74, opened 2019-06-09
Updated docker-compose configuration

kmehant:
Changed docker-compose configuration for restarting database container when killed.


Chuxel:
This is a good thought, but this would also start the mongo container even if you stopped the Node container.  Perhaps `restart: unless-stopped`?

kmehant:
> This is a good thought, but this would also start the mongo container whenever Docker is restarted even if you stopped it and/or the Node container. Perhaps `restart: unless-stopped`?

Yes, this seems a perfect fix! Thank you @Chuxel

PR #87, opened 2019-06-18
Test template

chrmarti:
Figured out a way to place tests here and use the remote-containers. For discussion.

PR #88, opened 2019-06-20
vscode-dev-containers regression tests

Chuxel:
I managed to convert and cleanup the initial test implementation after the extension refactoring.  Only odd thing about it is it does require the private repo but is in an OSS repo, so anyone cloning the repo wouldn't actually be able to use it. Started with #87 and built up from there.

It also adds in a few utilities to allow you to turn on trace logging and get config either from an env var or a json file -- we could move those into the core test utilities if you think they'd be broadly useful.  Here they're important since these are bulk tests primarily, but you want to be able to scope to one test for debugging of the tests themselves. The only other thing is I am duping the json schema here since cloining the distro would be overkill for that one file.

Thoughts?


Chuxel:
One thing I hit with using package.json to grab the extension repo is that I got a cached copy of the repo since yarn.lock said 0.62.0 -- the latest for the repo is 0.63.0.  We may need to re-add it to ensure the latest is pulled or run `yarn install --no-lockfile` or something like that in CI.

chrmarti:
We could download the schema from the public repo: https://github.com/microsoft/vscode/blob/master/extensions/configuration-editing/schemas/devContainer.schema.json

Maybe `yarn upgrade ...` also works on URLs (not tried yet).

test/src/containerTestBase.ts
```diff
@@ -0,0 +1,139 @@
...
+		log('trace', result.error + ': ' + result.stderr);
+	}
+
+	return result;
+}
+
+export async function cleanUpTest(devContainer: ResolverResult) {
+
+	log('debug', 'Stopping container');
+	devContainer.params.shutdowns.forEach(async (shutdown: (rebuild?: boolean) => {}) => await shutdown(false));
```
xWiiLLz: Is the goal here to await for all of them to shutdown? If so, I don't think awaiting shutdown inside that forEach is what you'd want.

Something like this might have that behaviour though:

<pre>
const shutdownAllContainers = devContainer.params.shutdowns.map(async (shutdown: async (rebuild?: boolean) => Promise<void>) =>  shutdown(false));
await Promise.all(shutdownAllContainers);
</pre>

test/src/containerTestBase.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+export async function testContainerResolver(definitionPath: string): Promise<ResolverResult> {
+
+	log('debug', `Finding and reading devcontainer.json at ${definitionPath}`);
+	const configPath = await getDevContainerConfigPathIn(definitionPath);
+	const config: any = configPath && await readDevContainerConfigFile(configPath);
+
+	const params = await createTestParams();
+	params.getExtensionsToInstall = () => config.extensions;
```
xWiiLLz: If `configPath` is undefined, line 93 will still set config to `undefined`. Should the error being thrown have more meaning than the possible `cannot read property "extensions" of undefined`?

test/src/containerTestUtils.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export function getRootFolderRealPath(type: string) {
+	const rootPath = getConfig('devContainerRoot', path.join(__dirname, '..', '..'));
+	const expectedDevConPath = path.join(rootPath, type);
+	return fs.existsSync(expectedDevConPath) ? fs.realpathSync(expectedDevConPath) : expectedDevConPath;
+}
+
+export function getDefinitionListFromPath(definitionsPath: string, configProperty: string, getFn: (path: string) => {} = defaultGetDefnitionListFn) {
+	const subsetToTest = getConfig(configProperty);
+	if (subsetToTest) {
```
xWiiLLz: Nit but this would reduce nesting:
<pre>
if (!subsetToTest) {
	return getFn(definitionsPath);
}
...
</pre>

test/src/containerTestUtils.ts
```diff
@@ -0,0 +1,105 @@
...
+		}
+	});
+	return definitionList;
+
+}
+
+
+export function findMountDestination(containerInfo: Docker.ContainerInspectInfo, sourceDir: string, vscodeDevConPath: string) {
+	for (let i = 0; i < containerInfo.Mounts.length; i++) {
+		const mount = containerInfo.Mounts[i];
```
xWiiLLz: Since you're using typescript and latest ES, you could shorten this with a nice `for..of`  😄 
<pre>
for (const mount of containerInfo.Mounts) {
</pre>

Chuxel:
I've made the suggested changes, but in thinking about it, it is pretty odd to put code in a public repo that depends on a private repo. Let's move this into a more logical location rather than here.

PR #89, opened 2019-06-21
Alpine definition

Chuxel:
Now that we have Alpine support in Remote - Containers for VS Code Insiders (🎉🎉🎉), it seems to make sense to create a container definition similar to the Debian / Ubuntu.

One of the things I noticed was the C++ extension didn't work out-of-box since it seemed to depend on glibc. So this container definition includes an optional glibc compat library for that reason - it should raise the chances of extensions working unmodified, so it seemed worth including in addition to git.

Anything else that jumps out we should include @chrmarti?

### DIFF
--- a/containers/alpine-3.10-git/.devcontainer/Dockerfile
+++ b/containers/alpine-3.10-git/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+FROM alpine:3.10
+
+ARG GLIBC_VERSION=2.29-r0
+
+RUN apk add --no-cache -q git \
+    #
+    # Install glibc compatibility - optional, but extnesion more likely to work
+    && apk add --no-cache ca-certificates wget libstdc++ libgcc \
+    && wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub \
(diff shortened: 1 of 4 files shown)


kmehant:
@Chuxel @chrmarti Shall we add this command?
``apk add --update alpine-sdk`` 
It installs many important packages for development such as 
* git
* build-base
* gcc 
* g++
* libc-dev
and many other.

containers/alpine-3.10-git/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+FROM alpine:3.10
+
+ARG GLIBC_VERSION=2.29-r0
+
+RUN apk add --no-cache -q git \
+    #
+    # Install glibc compatibility - optional, but extnesion more likely to work
```
chrmarti: Typo: 'extnesion' > 'extensions are'

chrmarti:
I wouldn't gear it too much towards OS native development, that will get in the way of Python / Node.js / etc. developers. Maybe we should have a generic definition and additionally need C++ on Alpine and Rust on Alpine definitions?

Chuxel:
@chrisdias Yes agreed, generic was my thinking here. I'm a bit worried about shifting existing any definitions to Alpine given all the docs are Debian based (as are most install instructions for tools for new Linux users coming from Mac / Windows).  We also know there are extensions that don't work yet (e.g. I tried Live Share and it failed).  I think we want to let this bake a bit and then consider whether we should move that direction broadly.

The question is whether there's anything else "essential" in the generic image we want to include -- I kind of don't think so.

chrmarti:
I was thinking of adding additional definitions, not changing existing ones. Going with a single one makes sense given that we need more time for extensions.

PR #93, opened 2019-07-03
Tweak Python 3 setup

brettcannon:
- Drop an unnecessary `*` at the end of a file name for copying
- Don't have pip bother checking for a newer version of pip (assume it's new enough as it came with the latest version of Python which the container is pulling from)
- Don't cache wheel files (to keep the image smaller)
- Specify the location of the `pylint` executable to turn on Pylint development

### DIFF
--- a/containers/python-3/.devcontainer/Dockerfile
+++ b/containers/python-3/.devcontainer/Dockerfile
@@ -10,7 +10,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # Copy requirements.txt (if found) to a temp locaition so we can install it. Also
 # copy "noop.txt" so the COPY instruction does not fail if no requirements.txt exists.
-COPY requirements.txt* .devcontainer/noop.txt /tmp/pip-tmp/
+COPY requirements.txt .devcontainer/noop.txt /tmp/pip-tmp/
 
 # Configure apt and install packages
 RUN apt-get update \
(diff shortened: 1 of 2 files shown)


containers/python-3/.devcontainer/Dockerfile
```diff
@@ -10,7 +10,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # Copy requirements.txt (if found) to a temp locaition so we can install it. Also
 # copy "noop.txt" so the COPY instruction does not fail if no requirements.txt exists.
-COPY requirements.txt* .devcontainer/noop.txt /tmp/pip-tmp/
+COPY requirements.txt .devcontainer/noop.txt /tmp/pip-tmp/
```
Chuxel: It looks like the `*` is actually needed here (at least in Docker for Windows where I tested this). What I am seeing when I test this is that it bombs on this line if no `requirements.txt` is present.  If `requirements.txt*` is used, it sill still pass if it isn't there as long as one file gets copied. Interesting behavior - not what I expected.

brettcannon: Ah, I kept testing with a `requirements.txt` file since I had to make sure `pip` worked and then I forgot to take it back out. 😁

brettcannon: I edited in-browser and put the `*` at the front as there's a chance people will have a `dev-requirements.txt` file and so this way makes a bit more sense. I haven't tested it yet but next time I'm in the office on my desktop and not my laptop I can give it a go.

brettcannon:
@Chuxel which Python tools team did you give access? It's a bit nebulous as we have several teams on GitHub across us and various companion teams. 😁

Chuxel:
@brettcannon I gave pvsc-team write access - LMK if I should use a different one.

Chuxel:
Tested it out - LGTM!

PR #94, opened 2019-07-03
Puppet PDK Docker Container file

jpogran:
This PR adds a [Puppet Development Kit](https://puppet.com/docs/pdk/1.x/pdk.html) docker container to VS Code Remote Container extension. It contains a Dockerfile that uses a base `ruby:2` docker image and installs PDK, git, and Ruby debugging gems.

This allows a user to create and edit Puppet manifests, modules and other code without installing the PDK or other Ruby tools on their machine. The [Puppet VSCode Extension](https://pup.pt/vscode) is included in the defintion to provide advanced intellisense and other IDE capabilties. The Ruby VS Code extension is also included to help with editing Ruby files.

### DIFF
--- a/containers/puppet/.devcontainer/Dockerfile
+++ b/containers/puppet/.devcontainer/Dockerfile
@@ -1,0 +1,15 @@
+FROM ruby:2
+
+RUN gem install ruby-debug-ide
+RUN gem install debase
+
+ADD https://apt.puppetlabs.com/puppet6-release-xenial.deb /puppet6-release-xenial.deb
+RUN dpkg -i /puppet6-release-xenial.deb
+
+RUN apt-get update \
+    && apt-get -y install git procps pdk
+
+# Clean up
+RUN apt-get autoremove -y \
+    && apt-get clean -y \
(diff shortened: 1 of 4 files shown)


Chuxel:
Apologies for the delay in getting back to you - the US 4th of July holiday slowed down our response times.  😊

Generally this looks great! Appreciate the contribution! The only thing I need to follow up on is that this references a Puppet license and copyright instead of the MIT license for the repo.  Currently these definitions get bundled and with the Remote - Containers extension, so I need to verify that this isn't going to cause any issues.  Just to verify - Is there a reason that it needs to be under the Puppet license?

jpogran:
No reason specifically, just habit and I didn't quite understand that part of the README doesn't change. I can change and rebase.

jpogran:
@Chuxel modified the license line

Chuxel:
@jpogran Thanks! Yes, we include that there just as a reference since each folder is self-contained. That way, anyone linking directly to a container README won't have to look at the repo root if they are wondering. There's a few spots in documentation where we link directly already.

PR #96, opened 2019-07-05
Add Azure Functions Python 3

anthonychu:
A clone of [Azure Functions Node LTS](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/azure-functions-node-lts) except:
* FROM python:3.6-stretch
* Add Docker-in-Docker (currently required for publish to Azure with native dependencies)


Chuxel:
LGTM - One question: Would it make more sense to call this "azure-functions-python-3" instead of 3.6 so the name does not change as Azure Functions adds 3.7 support?

anthonychu:
Good question. I think Python 3 makes sense. I'll change it. Thanks.

Chuxel:
LGTM!  Thanks!

PR #115, opened 2019-08-09
[WIP] Create an AWS Multi Language development container.

hastarin:
This is a rather large template unlike the existing options but it can be paired back by commenting out various bits and I thought it helps demonstrate just how powerful, and flexible, these dev containers can be.


Chuxel:
Thanks for the PR - Will try to carve out some time here to check it out and get back to you in more detail.

One thought is that we've been talking about adding some "Kitchen Sink" dev container definitions for quickly getting started.  The idea would be that you can just install the kitchen sink image to start, and then refine from there - and it would cover a bunch of different languages.  This actually fits that quite well with what you are doing here, but with an AWS spin.  To keep these together, perhaps we should call this:

"Kitchen Sink - AWS" and kitchensink-aws

A non-cloud kitchen sink could just be kitchensink, and we could then have versions for other cloud vendors if appropriate. Common docs could then tell people to look for "Kitchen Sink" images and see all of them.

Chuxel:
A few other questions comments:

1. Right now this doesn't appear to work as expeced on macOS and Linux.  Some of the volume mounts use Windows specific environment variables like USERPROFILE and APPDATA. We'd want to also support the macOS and Linux equivalent locations (presumably somewhere under HOME).  It also uses USERNAME which does not exist on macOS / Linux - it's USER there.
2. How critical is posgres to the definition?  Is this used by some of the AWS tools? I ask because database tech is one of those things that tends to go by community - e.g. Mongo is popular in the Node.js space.

hastarin:
@Chuxel I'm happy to refine/improve/change as needed to what you think is going to offer the most benefit to the wider community.  This is just based off something I'm using for a POC project and it's early days yet so I certainly welcome feedback/suggestions.  

Postgres/pgAdmin4 are not critical at all.  They can certainly be removed if needed or perhaps just commented out by default but left in the docker-compose as an example?

I had included some notes about Mac/Linux in the README.md.  If need be I can try to make some time this weekend to investigate what the mounts would need to be for those and include some further comments/sections.

hastarin:
@Chuxel I've done what I can without VSCode on macOS/Linux to test it further.  Feel free to review when you get the chance.

animateM:
You may want to get rid of the continuation character at the end of line 166 of the dockerfile?

containers/kitchensink-aws/.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+services:
+  ubuntu-dev:
+    user: vscode
+
+    build:
+      context: .
+      dockerfile: Dockerfile
+
+    volumes:
+      - ..:/workspace
```
natemacinnes, line 17: Why mount in the root workspace? Why not in `/home/vscode/workspace`?

derekmurawsky:
> 1. Right now this doesn't appear to work as expeced on macOS and Linux.  Some of the volume mounts use Windows specific environment variables like USERPROFILE and APPDATA. We'd want to also support the macOS and Linux equivalent locations (presumably somewhere under HOME).  It also uses USERNAME which does not exist on macOS / Linux - it's USER there.

Just put the two variables next two eachother. USER and USERNAME should not be available on any one system and if it is not set it will be blank. 

Example from my devcontainer.json file:
```
"mounts": [
		"source=${localEnv:HOME}${localEnv:USERPROFILE}/.aws/credentials,target=/home/vscode/.aws/credentials,type=bind"
	],
```

Note: I did not come up with this but rather found it in a blog post somewhere. It does seem to work well for me, though.

containers/kitchensink-aws/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,181 @@
...
+
+#
+# Install awscli
+RUN apt-get update && apt-get -y install python3-pip \
+    && apt-get autoremove -y \
+    && apt-get clean -y \
+    && pip3 install awscli \
+    && rm -rf /var/lib/apt/lists/*
+#
+# Install Python3
```
derekmurawsky, line 67: Shouldn't this go before you install python3-pip? I believe installing python3 is redundant afterwards. If is it not, why can't they both be installed at the same time in one `apt install`?

hastarin:
Apologies folks I'd forgotten all about this and haven't worked on it any further. We've got our own containers internally and trying to get this updated is not something I have the time/motivation for at this point. Thanks to all those who provided feedback.

PR #117, opened 2019-08-14
Remove redundant "extensions" field.

andyli:
There is already an `"extensions"` field at line 19.


Chuxel:
Good catch!  Actually, I think what we'd want to do is remove the comment line and leave `"extensions":[]`. This last line ensures there isn't a dangling comma so it passes jsonc schema validation checks.  I can make that tweak after merging.

PR #124, opened 2019-08-28
Set java home via settings property

dehru:
This PR updates the devcontainer.json file to use the settings property to point to the container installed path of java home.  Thanks for help @Chuxel


Chuxel:
LGTM!

PR #126, opened 2019-09-06
Alpine 3.10 container

Chuxel:
Holding off merging until https://github.com/microsoft/vscode-remote-release/issues/1026 is resolved.  

Worst case, will back down to Alpine 3.9 for now if the Node and/or Alpine folks can't get us an answer. 

//cc: @chrmarti FYI


Chuxel:
Merging now that 1026 is closed.

PR #138, opened 2019-10-18
Enable optional installation of Node and the Azure CLI

### LINKED ISSUE #133: Include better defaults for .NET containers
The documentation is woefully inadiquate for how to setup the debugging support for say .NET Core applications and these dev containers don't add the right stuff to ensure that the debugger will work.

The entries should be created in the launch.json and whatever ports etc. that are necessary should also be opened.  Failing that the documentation needs to be updated to actually outline how this is supposed to work.

Chuxel:
Resolves #133. 

This PR includes arguments in the Dockerfile that will install Node.js (common need for ASP.NET web apps) and the Azure CLI.

By default, Node.js is installed, but the Azure CLI is not.

READMEs also contain this information.

### DIFF
--- a/containers/dotnetcore-2.1/.devcontainer/Dockerfile
+++ b/containers/dotnetcore-2.1/.devcontainer/Dockerfile
@@ -5,9 +5,6 @@
 
 FROM mcr.microsoft.com/dotnet/core/sdk:2.1
 
-# Avoid warnings by switching to noninteractive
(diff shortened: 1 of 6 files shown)


Chuxel:
@richlander I'm going to merge this in to catch our next release train.  LMK if you end up thinking we should do something else here.  It's easy enough to tweak.

Issue #141, opened 2019-11-04
Provide guidance for dealing with .NET self-signed certs


Chuxel:
This is a fork of #133 given several other topics in that issue have been addressed.

Currently the dotnet dev container definitions and try repos do not provide good instructions for dealing with self signed certificates in a container environment.  

The question is what the best practice should be for adapting to this and we can then at least include a reference to information on it if not update the definitions to have things set up correctly by default (depending on the recommendation).

Chuxel:
Ok, this seems to get the job done:

1. On Windows run:

    ```bat
    dotnet dev-certs https -ep "%USERPROFILE%/.aspnet/https/aspnetapp.pfx" -p "SecurePwdGoesHere"`
    ```
    macOS/Linux: 

    ```bash
    dotnet dev-certs https -ep "~/.aspnet/https/aspnetapp.pfx" -p "SecurePwdGoesHere"`
    ```

2. Add a few env vars and a volume mount into devcontainer.json

```json
// For format details, see https://aka.ms/vscode-remote/devcontainer.json or the definition README at
// https://github.com/microsoft/vscode-dev-containers/tree/master/containers/dotnetcore-2.2
{
	"name": "C# (.NET Core 3.0)",
	"dockerFile": "Dockerfile",
	
	// Use 'settings' to set *default* container specific settings.json values on container create. 
	// You can edit these settings after create using File > Preferences > Settings > Remote.
	"settings": { 
		"terminal.integrated.shell.linux": "/bin/bash"
	},

	// Uncomment the next line if you want to publish any ports.
	"appPort": [5000, 5001],

	"runArgs": [
		// Uncomment the next line to use a non-root user. On Linux, this will prevent
		// new files getting created as root, but you may need to update the USER_UID
		// and USER_GID in .devcontainer/Dockerfile to match your user if not 1000.
		// "-u", "vscode",

		"-e", "ASPNETCORE_Kestrel__Endpoints__Http__Url=https://*:5000",
		"-e", "ASPNETCORE_Kestrel__Endpoints__Https__Url=https://*:5001",
		"-v", "${env:HOME}${env:USERPROFILE}/.aspnet/https:/home/vscode/.aspnet/https",
		"-e", "ASPNETCORE_Kestrel__Certificates__Default__Password=SecurePwdGoesHere",
		"-e", "ASPNETCORE_Kestrel__Certificates__Default__Path=/home/vscode/.aspnet/https/aspnetapp.pfx"
	],

[...]

Issue #140, opened 2019-11-04
Pre-build images for Microsoft maintained container images in this repository


Chuxel:
Currently this repository includes a number of dev containers for people to use, but requires that the images be built.  While this is great for an example, certain definitions like debian-9-git could clearly have a base image with git already installed and then just a sample Dockerfile for using it.

We can cherry pick a subset of these that are maintained by Microsoft and pre-build them much as we recommend for external contributions.

Chuxel:
Currently the vscode-dev-containers repo contains pure Dockerfiles that need to be built when used. While this works well from the perspective of providing samples, some of the images install a significant number of runtimes or tools which can take a long time to build.

The goal of this proposal is to:

1. Ensure vscode-dev-containers continues to be a good source of samples
2. Improve the performance using the most popular (or in some cases slowest to build) container images
3. Make it easy for users to add additional software to the images
4. Make it easy for contributors to build, customize, and contribute new definitions

We won't be able to build all images in the repository or publish them under a Microsoft registry, but we would want to allow contributors to build their own images for contributions if they wanted to do so by keeping the ability to use a stand alone Dockerfile, image reference, or docker-compose file like today.

## Image generation for vscode-dev-containers

In order to meet the above requirements, this first phase keeps the way the dev containers deployed as it is now - an npm package. Future phases could introduce a formal registry, but this is overkill current state.

### Versioning

At this phase, versioning of the image can be tied to the release of the npm package. For example, as of this writing, the current version is 0.35.0. All images that are generated would then inherit this version.

We would follow semver rules for bumping up the repository name - any fix that needs to be deployed immediately would be "fix" level semver bump. When released as latest, the image would be tagged as is done in other Docker images. Using the 0.35.0 example, new images would be tagged as:

- 0
- 0.35
- 0.35.0
- latest

In the unlikely event a break fix needed to be deployed and not tagged latest, we would have a facility to tag as follows:

- 0.35
- 0.35.0

This has a few advantages:

1. Tags are generated for each version of the repository that is cut. This allows us to refer directly to the exact version of the source code used to build the container image to avoid confusion. e.g. `https://github.com/microsoft/vscode-dev-containers/tree/v0.35.0/containers/javascript-node-8`
2. Similarly, as containers are deprecated and removed from the repository, you can still refer back to the container source and README.
3. Upstream changes that break existing images can be handled as needed.
[...]

PR #142, opened 2019-11-06
Remove "lts" / "latest" definitions in favor of version, add eslint to TS definitions

Chuxel:
There have been a few cases where an update to a "latest" or "lts" tag has broken a definition.  Instead of using these tags, we can keep dev containers for current versions of runtimes and deprecate those that are no longer in support.

This change will also be required to implement #140 and also addresses #141 for .NET Core

We will need to adjust comments that are added to devcontainer.json files to that points to the README in the release tag tree instead of master so that people can still get to READMEs for deprecated containers.

This update also adds eslint to the TypeScript definitions now that tslint has been deprecated.


chrmarti:
Makes sense. 👍

PR #143, opened 2019-11-07
Update Docker-in-Docker templates

karolz-ms:
The Docker extension defaults to being a workspace extension now so the previous workaround is unnecessary

### DIFF
--- a/containers/docker-in-docker-compose/.devcontainer/devcontainer.json
+++ b/containers/docker-in-docker-compose/.devcontainer/devcontainer.json
@@ -8,10 +8,7 @@
 	// Use 'settings' to set *default* container specific settings.json values on container create. 
 	// You can edit these settings after create using File > Preferences > Settings > Remote.
 	"settings": { 
-		"terminal.integrated.shell.linux": "/bin/bash",
-		"remote.extensionKind": {
-			"ms-azuretools.vscode-docker": "workspace"
-		}
+		"terminal.integrated.shell.linux": "/bin/bash"
 	},
 
 	// Uncomment the next line if you want start specific services in your Docker Compose config.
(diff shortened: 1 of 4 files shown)


Chuxel:
Right now it looks like `product.json` is still forcing the Docker extension to the `ui` side, so this would be required.  In addition, the setting should be removed from `devcontainer.json` as well once this has changed.

karolz-ms:
@Chuxel where is the `product.json` file that you are referring to?

Chuxel:
Internal vscode distro settings.  You can see it under `extensionKind` in `C:\Users\clantz\AppData\Local\Programs\Microsoft VS Code\resources\app\product.json` on Windows.

karolz-ms:
@Chuxel thanks. Bear with me please&mdash;I am new to VS Code extension development. 

I can see that in `product.json` our extension is marked as `extensionKind=UI`. This is incorrect, see https://github.com/microsoft/vscode-docker/blob/master/package.json#L11 

Do you happen to know where is this `product.json` sourced from and what makes it claim Docker extension is of type UI? I searched https://github.com/microsoft/vscode for `vscode-docker` but could not find anything relevant..

Chuxel:
@karolz-ms Actually - I take that back, it looks like the extension setting take precedent over `product.json`.   I just tried on a remote SSH env and it installed remotely.

We actually need to make some doc updates with this change since I don't think anyone realized it had been changed.

karolz-ms:
@Chuxel thank you, and please let me know if I can help in any way (I am on the team that owns the extension). 

Just FYI we will be updating [the extension wiki](https://github.com/microsoft/vscode-docker/wiki) soon, adding more information about remote usage among other things.

Chuxel:
Thanks so much for raising the PR!  Given this updated info, I'm happy to merge.  I'll also update the Kubernetes-Helm container since it does the same thing offline.

Doc updates are in progress as well: https://github.com/microsoft/vscode-docs/pull/3234

karolz-ms:
Absolutely, any time! Appreciate merging this PR so quickly!

PR #145, opened 2019-11-11
Add Azure CAF rover to build terraform landingzones

LaurentLesle:
This PR adds support to build Azure CAF Terraform landing zones.

https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/expanded-scope/terraform-landing-zone

https://github.com/aztfmod/rover/tree/vscode-remote-container

### DIFF
--- a/containers/azure-caf-terraform-rover/.devcontainer/devcontainer.json
+++ b/containers/azure-caf-terraform-rover/.devcontainer/devcontainer.json
@@ -1,0 +1,36 @@
+{
+    "name": "Azure CAF rover",
+ 
+    // Update the 'dockerComposeFile' list if you have more compose files or use different names.
+    "dockerComposeFile": "docker-compose.yml",
+ 
+    // The 'service' property is the name of the service for the container that VS Code should
+    // use. Update this value and .devcontainer/docker-compose.yml to the real service name.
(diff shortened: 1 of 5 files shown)


Chuxel:
Thanks so much for the contribution!  I've got a few questions having looked over the PR:

1. Is the `workspace.code-workspace` file needed? This would get dropped on top of a developer's project, so it could overwrite something they already have.  If it's just for testing, then a `.npmignore` should get added - I noticed it's missing. That tells VS Code to ignore it when adding content to a user's project.

2. Given there's only one container, could this be done just using `devcontainer.json` rather than docker-compose?  The other downside with using compose is that environments like Visual Studio Online do not support it yet.  Experimental custom Dockerfile/image support is already in there.

3. Similarly, there's a number of volume mounts.  Some of these look like optimizations that are being investigated for the core product (See https://github.com/microsoft/vscode-remote-release/issues/421 and the November plan).  Others like SSH keys we aren't doing by default given the security implications - and have considered adding a first class property to allow you to quickly enable it.  Is it required in this case?  Typically we've been including workarounds in READMEs or docs to deal with shortcomings rather than adding them to files that go into the core product.

Chuxel:

PR #150, opened 2019-11-21
docker-in-docker: add missing "\" in README.md

ggilmore:
Without the `\`, "docker build" will fail since it thinks that `&&` is a directive on its own (instead of being a part of the previous `RUN` directive).


Chuxel:
Good catch!  Thanks for finding it.

PR #159, opened 2019-11-23
netstandard reference shouldn't be necessary in F# projects

cartermp:
This comes in the shared framework pulled in by the SDK in the project file, so it should be fine to remove. We don't include it in new templates in the .NET SDK.


Chuxel:
Sounds good.  Thanks for the contribution!

PR #163, opened 2019-12-09
Updates to take advantage of new devcontainer.json properties in Remote - Containers

Chuxel:
The release of 1.41 to stable will also release the current insiders version of the Remote - Containers extension which includes a number of properties to address common issues (e.g. `remoteUser`, `containerUser`, `remoteEnv`, `containerEnv`, `mounts`).  This PR updates existing definitions to use these properties or simply reference them in comments.


Chuxel:
All tests appear to pass.

PR #164, opened 2019-12-09
Typo: "presenet" -> "present"

forana:


Chuxel:
Good catch. Thanks for the contribution!

PR #173, opened 2019-12-27
Fixed gopls package path

FrancescoIlario:


Chuxel:
Thanks for the PR! Looks like the it changed upstream. Appreciate it.

PR #177, opened 2020-01-01
Added Jekyll developer container

cmendible:
Added a Jekyll developer container

Hope it helps!

### DIFF
--- a/containers/jekyll/.devcontainer/Dockerfile
+++ b/containers/jekyll/.devcontainer/Dockerfile
@@ -1,0 +1,48 @@
+FROM debian:latest
+
+# Avoid warnings by switching to noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
+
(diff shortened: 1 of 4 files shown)


Chuxel:
Looks good!  Quick question - Should this be prefixed with "ruby"?  For example, we have `ruby-5-rails-5` and `ruby-5-sinatra`.  Mainly helps with sorting like-ecosystem things together when people are looking for something.

cmendible:
Well the way I see it this container is just for developing Jekyll sites and not also for ruby development. Ruby in this case is [requirement v 2.4.0 or above](https://jekyllrb.com/docs/installation/) 

Perhaps I should remove Ruby from the language platform row in the README.md?

Chuxel:
I think as long as people will understand and be able to find it, it's probably fine as is. Otherwise LGTM, so merging!

PR #179, opened 2020-01-04
Fix variable name typo (DOTNETCORE_VESRION -> DOTNETCORE_VERSION)

bergmeister:
Please propagate this change to the remote dev container extension as well please


Chuxel:
Good catch!  Thanks so much!

PR #188, opened 2020-01-10
Add a Dev Container for developing Dapr with Node.js 12 & TypeScript

philliphoff:
Adds a definition for a dev container for using Dapr with Node.js 12 and TypeScript. This definition combines the definition for Node.js 12 and TypeScript with the Docker in Docker Compose definition, plus installs the latest version of the Dapr CLI (which requires the latter). When started, the dev container automatically initializes the Dapr runtime, using a separate Docker network to isolate it from the local environment.

This definition is analogous to the definition for [Dapr with C# and .NET Core 3.0](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/dapr-dotnetcore-3.0).


Chuxel:
Thanks for the contribution!! LGTM!

PR #192, opened 2020-01-15
Use a common Docker network across Dapr Dev Containers.

philliphoff:
Currently, each Dapr Dev Container for a given platform use a platform-specific Docker network, the thinking being that developers would want a completely isolated Dapr development environment.  However, in multi-service application scenarios, developers are likely to want the services to interact, even if developed across different platforms.

This change makes Dapr Dev Containers use a common Docker network by default, to enable such scenarios without additional configuration.  In practice, even for users *not* requiring interacting services (e.g. those new to Dapr), using a common network likely has no real impact on their development.  (And, if it does the user can always change the configuration to use a unique network.)


Chuxel:
LGTM! Thanks for the contribution!

PR #207, opened 2020-01-22
Update Dockerfile

spaceneenja:
fixed typo


Chuxel:
Good catch! Thanks for the contribution!

Issue #226, opened 2020-02-04
Cant run 'go mod init' as non-root-user in Go


whesleybarnard:
- VSCode Version:
```
Version: 1.41.1
Commit: 26076a4de974ead31f97692a0d32f90d735645c0
Date: 2019-12-18T15:04:31.999Z
Electron: 6.1.5
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Linux x64 4.19.99-1-MANJARO
```
- Local OS Version:
```
Linux manjaro 4.19.99-1-MANJARO #1 SMP Mon Jan 27 16:36:28 UTC 2020 x86_64 GNU/Linux
```
- Name of Dev Container Definition with Issue:
```
vscode-dev-containers/containers/go/.devcontainer/Dockerfile
```

Steps to Reproduce:

1. Set remoteUser as vscode.
2. Open vscode inside the go dev container.
3. Try to run go mod init in a non gopath folder.

Example command and error:
```
go mod init test.com/test/tmp

go: creating new go.mod: module test.com/test/tmp
go: failed to create cache directory /go/pkg/mod/cache: mkdir /go/pkg/mod: permission denied
```

I have noticed that the /go/pkg folder only has write permissions for root. I think this is the issue.

Chuxel:
Should be resolved with your PR (#227)!

PR #227, opened 2020-02-04
Added write permission for /go/pkg in go container.

whesleybarnard:
This tries to solve the problem of not being able to run go mod init as a non-root-user in the go dev container.

See github [issue](https://github.com/microsoft/vscode-dev-containers/issues/226).

This is my first pull request. Not sure if I am doing it correctly...


Chuxel:
Good catch! LGTM!

Issue #223, opened 2020-01-31
Parallelize image generation to speed up CI process


Chuxel:
Currently there are not that many images to build, but as time goes on we will want to be able to spin up concurrent image builds to speed up the process. 

This will require adding the idea of "paging" into the build CLI and then some tweaks to the GitHub Actions workflow files.

Chuxel:
One significant challenge here is that definitions can depend on each other, so all definitions for a given dependency tree need to be done in the same page to be sure the updated image is available first.

Chuxel:
Resolved with #228.

PR #228, opened 2020-02-05
Parallelize image generation to speed up CI process

Chuxel:
This implements #223. The core build time is now approx 2-3 mins with three "pages" that are built concurrently.  As we onboard more images, we can add more pages to speed things up further. Packaging then happens after all the builds have completed though in concept it could happen in parallel.

The logic uses one level of image dependencies to do the paging so that, if an image depends on another image, the parent image is built first.  This logic does not extend to multiple parents currently. (Parents can't have parents yet, but we have no definitions in this scenario today.)

All of this required some refactoring of the build CLI commands to get working.

cgmanifest.json generation is also automatically triggered once the build completes but as a separate workflow via the repository_dispatch event. The manifest generation has to be across all images and therefore will generally be slower because, as a separate job, it has to download all built images before inspecting their contents.  By having it as a separate workflow, this allows the CI build to complete quickly.

See the following runs for examples of results:
Release: https://github.com/Chuxel/vscode-dev-containers/actions/runs/35231514
CI: https://github.com/Chuxel/vscode-dev-containers/actions/runs/35228630
CG Manifest triggered by CI: https://github.com/Chuxel/vscode-dev-containers/actions/runs/35229616

//cc: @chrmarti @egamma

### DIFF
--- a/.github/workflows/push-and-package.yml
+++ b/.github/workflows/push-and-package.yml
@@ -7,8 +7,13 @@ on:
     - 'v*'
 
 jobs:
-  build-page-1:
-    name: Build and push (page 1/3)
+  build-and-push:
+    name: Build and push
+    strategy:
+      matrix:
+        page: [1, 2, 3]
+        page-total: [3]
+      fail-fast: true
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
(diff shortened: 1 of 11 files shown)


.github/workflows/push-and-package.yml
```diff
@@ -7,11 +7,13 @@ on:
     - 'v*'
 
 jobs:
-  build:
-    
+  build-page-1:
```
chrmarti: Looks good. I guess the jobs can't be dynamically generated?

Chuxel: Initially I didn't think so because all the examples for the [matrix](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) property are about the OS, but I took a look again just now and it appears I can add any variable to the mix.  I'm testing a variant that uses it and will move that here assuming it works.

Chuxel: Yep!  It worked.  🎉

Chuxel:
I'll merge this after we release just in case I messed something up I'm not seeing.

PR #229, opened 2020-02-06
install `less` to do `git log` without error msg

and-y:
Currently when I am attached to the `python-3` container perform a `git log`, the
command complains that `less` is not available.
So I added it to the same `apt-get` where `git` is installed as well.


Chuxel:
Good suggestion.  I'll add this to the new common scripts as well.

PR #231, opened 2020-02-06
Add less to definitions for git log support

### LINKED ISSUE #230: Install less in all definitions for git log
Currently "less" is not present in all scenarios which can cause git log to error. This should be added to the common script and any definition that does not use it.

Chuxel:
Resolves #230 for all definitions, #229 was for python.

Full regressions are running.  Will merge once I know they pass.


Chuxel:
All tests pass.

PR #232, opened 2020-02-07
Automatically update tags to use version specific script source

Chuxel:
This PR resolves a current issue where Dockerfiles tied to a release tag do not reference version specific links to common scripts.  The resulting `.tgz` package does, however, so this is not P0.  However, this update makes it easier to reproduce a build from a tag.

One downside of the approach, however, is that it deletes and updates an existing tag to maintain the current workflow where just tagging triggers a release.  As a result, whoever triggers a release will need to do a `git fetch --tags --force` to get the updated tag. Otherwise Git will complain about clobbering an existing local one.

### DIFF
--- a/.github/workflows/push-and-package.yml
+++ b/.github/workflows/push-and-package.yml
@@ -18,13 +18,13 @@ jobs:
     steps:
     - name: Checkout
       id: checkout
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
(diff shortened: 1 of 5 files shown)


Chuxel:
The problem with that is that people would never be able to use the contents of the repository directly.  One of the things we'll want to retain here is the simplicity of just grabbing the contents of a folder to get the latest.  Copying files around prevents that.

PR #234, opened 2020-02-09
Fix duplicate ENV statement in several dockerfiles

hcaw:
Several of the Dockerfiles contained the following line in duplicate:
```
ENV DEBIAN_FRONTEND=noninteractive
```
As such, I removed the second instance of each line, including the preceding comment.
I think this occurred because the mistake was included in the template Dockerfile.


Chuxel:
The reason the first line is present is to eliminate errors during the docker build.  However if you leave this on, then if you manually install anything once in the container, you won't be prompted, so it reverts it back to dialog when done.

Chuxel:
Given my explanation above,  I don't think we'll end up being able to take this PR.  Thanks so much for the contribution though and let me know if you don't think I've understood what you were trying to say here.  Thanks!

PR #238, opened 2020-02-14
Image build system "variant" support

Chuxel:
Historically dev container definitions used a Dockerfile to pull a base image. This made it easy to reference different versions of something a given runtime. For example, `python-3` can be tweaked to support 3.5, 3.6., 3.7, 3.8, etc.  However, if we pre-build these images, we loose this ability without creating a separate dev container for each version (`python-3.5`).

To avoid this problem, this PR updates the build system to support the idea of a "variant".  Consider this in the `definition-build.json` file for Python that is included in this PR:

```json
	"build": {
		"rootDistro": "debian",
		"tags": [
			"python:${VERSION}-${VARIANT}"
		],
		"variants": ["3", "3.8", "3.7", "3.6", "3.5"]
	},
	"dependencies": {
		"image": "python:${VARIANT}",
```

The Dockerfile then just references the VARIANT build arg in the correct location.

```Dockerfile
ARG VARIANT=3
FROM python:${VARIANT}
```

The build system then generates the following:
- vscode/devcontainers/python:3
- vscode/devcontainers/python:3.8
- vscode/devcontainers/python:3.7
- vscode/devcontainers/python:3.6
- vscode/devcontainers/python:3.5

The first item in the variant array is then used for `latest` tag (in this case python:3).

//cc: @chrmarti @brettcannon @egamma


chrmarti:
Sounds good. One question that has come up with Node.js versions too: What about having a single image with multiple Python versions installed that can be switched between by using Python environments? Having less images might be easier to maintain and handle in the UI (think growing lists).

Chuxel:
@chrmarti @brettcannon Setting aside Node (which really wasn't want drove this PR), for Python, I think that will result in a large image (with 3.5 through 3.8 installed) and then require the use of something like virtual env.  Feels like this breaks the "get going quick" promise - It could work, but I'm unsure that it is the right pattern to recommend here.

@brettcannon For Python, what are your thoughts on multiple images vs one with multiple versions in it?

Chuxel:
@chrmarti FYI - Brett is away until the 26th. One other downside in having a unified image is that we'd end up needing to do a major version rev each time we deprecated a runtime version.  Right now, the image would continue to live for python 3.5 even after we deprecated it. If we merged into one image, that would not be the case.  There's a similar issue around the VSO image we need to think through.

Chuxel:
Given the discussion here, I'll split this PR into two - one with the core variant support (this PR) and then a secondary PR for how we want to move forward with the Python definition specifically.

PR #241, opened 2020-02-21
Onboard python-3 to image build

Chuxel:
This is a continuation of PR #238 which originally included on-boarding Python into the image build process for #154.

To recap, this PR covers on-boarding Python images into vscode/devcontainers/python repository in MCR.  Since Python's currently in support versions are 3.5 through 3.8, this PR generates separate images per version from the same python-3 definition.

So you get:

- vscode/devcontainers/python:3
- vscode/devcontainers/python:3.8
- vscode/devcontainers/python:3.7
- vscode/devcontainers/python:3.6
- vscode/devcontainers/python:3.5

...along with a "latest" that ties to python:3 and then version specific tags (e.g. python:0.102.0-3.6)

Discussion: The discussion in #238 is whether we would be better off having a single "python" image that includes all supported versions rather than separate images per version.  

Pro: The upside is that developers can just grab a single image and then use virtual env to pick their version.
Con: We would end up with a larger image and people practically speaking would be forced into using virtual env. We would also need to implement separate image versioning for the python image so that we could increment the major version number any time we removed a python version since this would be breaking.  That makes finding the version number harder since it wouldn't mirror the VS Code extension anymore.

The question is which approach is best for Python.

//cc: @chrmarti @brettcannon @egamma


brettcannon:
If you're developing an app you're only going to care about a single Python version, but if you're developing a library you're going to want a bunch of versions.

I say do what makes your life easier which sounds like individual images. For those of us who need more than that we can edit the Dockerfile to pull in e.g. [deadsnakes](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) if the base image is Ubuntu (in which case maybe include a comment about this?).

Chuxel:
Apologies for the delay here - had some unexpected OOF'ness last week.  Makes sense.  Adding the comment also is a good idea.  I can tweak the PR.

Chuxel:
I've adapted to the feedback.  This will now generate separate images, but also includes instructions in `base.Dockerfile` for building one with multiple version of Python inside it.

In VS Code, the `Dockerfile` in this folder is used (though the image tag is updated) which includes some customizations not in the generic Debian Dockerfile (under `build/assets/debian.Dockerfile`).

The resulting output looks like this:
```json
{
    "name": "Python 3",
    "context": "..",

    // Update the FROM statement in this Dockerfile to pick a version of Python.
    "dockerFile": "Dockerfile",

    // Set *default* container specific settings.json values on container create.
    "settings": {
        "terminal.integrated.shell.linux": "/bin/bash",
        "python.pythonPath": "/usr/local/bin/python",
        "python.linting.enabled": true,
        "python.linting.pylintEnabled": true,
        "python.linting.pylintPath": "/usr/local/bin/pylint"
    },

    // Add the IDs of extensions you want installed when the container is created.
    "extensions": [
        "ms-python.python"
    ]

    // Use 'forwardPorts' to make a list of ports inside the container available locally.
    // "forwardPorts": [],

    // Use 'postCreateCommand' to run commands after the container is created.
    // "postCreateCommand": "pip3 install -r requirements.txt",

    // Uncomment to connect as a non-root user. See https://aka.ms/vscode-remote/containers/non-root.
    // "remoteUser": "vscode"
}
```

```Dockerfile
# Copyright (c) Microsoft Corporation. All rights reserved.

# Available Python image versions: python:3, python:3.8, python:3.7, python:3.6, python:3.5
# To fully customize the contents of this image, use the following Dockerfile instead:
# https://github.com/microsoft/vscode-dev-containers/tree/clantz/python-onboard/containers/python-3/.devcontainer/base.Dockerfile
FROM mcr.microsoft.com/vscode/devcontainers/python:0-3

# [Optional] If your requirements rarely change, uncomment this section to add them to the image.
#
# COPY requirements.txt /tmp/pip-tmp/
# RUN pip3 --disable-pip-version-check --no-cache-dir install -r /tmp/pip-tmp/requirements.txt \
#    && rm -rf /tmp/pip-tmp

# [Optional] Uncomment this section to install additional packages.
#
# ENV DEBIAN_FRONTEND=noninteractive
# RUN apt-get update \
#    && apt-get -y install --no-install-recommends <your-package-list-here> \
#    #
#    # Clean up
#    && apt-get autoremove -y \
#    && apt-get clean -y \
#    && rm -rf /var/lib/apt/lists/*
# ENV DEBIAN_FRONTEND=dialog
```

containers/python-3/.devcontainer/base.Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+    && apt-get -y install --no-install-recommends apt-utils dialog wget ca-certificates 2>&1 \
+    #
+    # Verify git, common tools / libs installed, add/modify non-root user, optionally install zsh
+    && wget -q -O /tmp/common-setup.sh $COMMON_SCRIPT_SOURCE \
+    && if [ "$COMMON_SCRIPT_SHA" != "dev-mode" ]; then echo "$COMMON_SCRIPT_SHA /tmp/common-setup.sh" | sha256sum -c - ; fi \
+    && /bin/bash /tmp/common-setup.sh "$INSTALL_ZSH" "$USERNAME" "$USER_UID" "$USER_GID" \
+    && rm /tmp/common-setup.sh \
+    #
+    # Install pylint
+    && pip3 --disable-pip-version-check --no-cache-dir install pylint \
```
brettcannon: If you install this globally as well as what's in `requirements.txt` there is the possibility that there will be a conflict between what the user wants to install and what pylint requires.

If you're installing pylint just to make development under VS Code easier by pre-installing the tool you could install this and other tools as supported by the Python extension into their own virtual environments and then add them as appropriate to `PATH` so they are usable but isolated from each other (people typically use [pipx](https://pipxproject.github.io/pipx/) for that sort of thing which would mean you could install _all_ supported tools and add a single `PATH` entry for pipx-installed tools).

Chuxel:
@brettcannon Since pipx requires Python 3.6, I stuck with a virtualenv, but that update is in now too.

While we are at it, are there other utilities that should be included aside from pip, virtualenv, and pylint?

brettcannon:
> Since pipx requires Python 3.6, I stuck with a virtualenv

Or just don't ship a Python 3.5 image. 😁 I mean it reaches EOL in September so it isn't long for this world and hopefully your work here will outlast September.

The other issue is you are putting **all** the tools in a single virtual environment, which means they will potentially conflict with each other due to their dependencies (we have this problem in our testing of the extension as some tools require older versions of things that newer tools don't like).

> are there other utilities that should be included aside from pip, virtualenv, and pylint?

It depends on how far you want to go with this. 😄 If you just want the popular ones, then in no particular order:
1. pylint
2. flake8
3. autopep8
4. black
5. pytest
6. virtualenv

If you want the complete list of tools that the Python extension supports, that adds in no particular order:
1. yapf
2. mypy
3. pydocstyle
4. pycodestyle
5. prospector
6. pylama
7. bandit

containers/python-3/.devcontainer/base.Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+    && if [ "$COMMON_SCRIPT_SHA" != "dev-mode" ]; then echo "$COMMON_SCRIPT_SHA /tmp/common-setup.sh" | sha256sum -c - ; fi \
+    && /bin/bash /tmp/common-setup.sh "$INSTALL_ZSH" "$USERNAME" "$USER_UID" "$USER_GID" \
+    && rm /tmp/common-setup.sh \
+    #
+    # Install default utilities
+    && mkdir -p ${UTILS_PATH} \
+    && chown ${USER_UID}:${USER_GID} ${UTILS_PATH} \
+    && pip3 --disable-pip-version-check --no-cache-dir install virtualenv \
+    && virtualenv ${UTILS_PATH} \
+    && ${UTILS_PATH}/bin/pip3 install ${DEFAULT_UTILITIES} \
```
brettcannon: Unfortunately this installs **all** tools into the same virtual environment which could lead to a conflict among the tools which unfortunately doesn't improve the situation.

Chuxel:
> Or just don't ship a Python 3.5 image. grin I mean it reaches EOL in September so it isn't long for this world and hopefully your work here will outlast September.
>
> The other issue is you are putting all the tools in a single virtual environment, which means they will potentially conflict with each other due to their dependencies (we have this problem in our testing of the extension as some tools require older versions of things that newer tools don't like).

Ah, that last part is interesting in particular. Ok, sounds like we should just drop 3.5 then - I was trying to include "supported" versions, but 3.5 is pretty old.  I was originally using pipx to install in this same location and then switched - I'll undo that and remove 3.5.

> It depends on how far you want to go with this.

Honestly, it's a mainly an image-size vs benefit question. With the pre-built images, install time isn't as much a factor anymore which was the main reason to keep things tight before.  It sounds like the first set would make sense at a minimum. Folks that were interested in the Go definition ended up submitting a PR to add everything the extension supported, so we could do that here as well if you think that makes sense.

brettcannon:
> it's a mainly an image-size vs benefit question.

None of these tools will be large (they are all pure Python code), so it's probably fine to just go ahead and install all of them.

Chuxel:
@brettcannon Okay, updated. Just installing pylint with pipx results in a 54M layer, with all the tools its 183M. The base image is pretty large to begin with however, so likely that's fine.  See what you think.

containers/python-3/.devcontainer/base.Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+    pytest \
+    yapf \
+    mypy \
+    pydocstyle \
+    pycodestyle \
+    prospector \
+    pylama \
+    bandit \
+    virtualenv \
+    pipx"
```
brettcannon: Should pipx be installed by pipx?

containers/python-3/.devcontainer/base.Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+    #
+    # Setup default python tools in isloated location
+    && mkdir -p ${UTILS_PATH}/bin \
+    && chown -R ${USER_UID}:${USER_GID} ${UTILS_PATH} \
+    && PYTHONUSERBASE=/tmp/pipx-tmp \
+        pip3 --disable-pip-version-check --no-cache-dir install --no-warn-script-location --user pipx \
+    && echo "${DEFAULT_UTILS}" | \
+        PYTHONUSERBASE=/tmp/pipx-tmp \
+        PIPX_HOME=${UTILS_PATH} \
+        PIPX_BIN_DIR=${UTILS_PATH}/bin \
```
brettcannon: These seem to be mirroring what's going on with `UTILS_PATH` and the change to `PATH`. Would it be easier to just set these environment variables above and drop `UTILS_PATH`?

containers/python-3/.devcontainer/devcontainer.json
```diff
@@ -1,15 +1,17 @@
...
+	"dockerFile": "base.Dockerfile",
 
 	// Set *default* container specific settings.json values on container create.
 	"settings": { 
 		"terminal.integrated.shell.linux": "/bin/bash",
 		"python.pythonPath": "/usr/local/bin/python",
 		"python.linting.enabled": true,
 		"python.linting.pylintEnabled": true,
-		"python.linting.pylintPath": "/usr/local/bin/pylint"
+		"python.linting.pylintPath": "/usr/local/py-utils/bin/pylint"
```
brettcannon: If we are going to install all the tools, would it be worth setting the paths now? They will be inactive unless the user flips them on, but at least they wouldn't need to specify the paths themselves (although even specifying them isn't necessary if they can be found on `PATH`).

containers/python-3/.devcontainer/base.Dockerfile
Chuxel: The challenge is that there are two users we need to consider - root and vscode.  So, the default install path for pipx wouldn't work well, I isolated pre-installed stuff in one location with the idea anything the user installs could be in a different spot.  That said, given pipx is mainly used for installing python utilities, maybe just setting the default path is better.

containers/python-3/.devcontainer/devcontainer.json
Chuxel: Can do.

containers/python-3/.devcontainer/base.Dockerfile
Chuxel: Here's the question - would installing pipx globally affect dependencies too?  I was instead installing in an alternate location temporarily and then isolating it so it was available for further use.

brettcannon: I guess if you're using pip to install pipx you're right that it should get installed on its own.

Chuxel: Yeah I looked at using apt-get, but this ended up requiring installation venv and other 
OS dependencies as well all wired into the operating system version of python rather than the one from the image.

I could set `VIRTUALENV_PYTHON` to deal with the default there. Maybe that's enough? apt-get is certainly less gymnastics.

brettcannon: Getting tied to the system Python isn't the worst thing in the world. I think it's a 🤷‍♂️ decision at this point. 😉

Issue #155, opened 2019-11-22
Create "kitchen sink" image based on core VS Online image to allow people to build up from it, use it offline


Chuxel:
Related to #140, #154, and https://github.com/MicrosoftDocs/vsonline/issues/205, we can create a dev container based off of the current VS Online image (which itself is based on Oryx).  

This will allow developers to:

1. Simply use it locally if they prefer
2. Make it easier to iterate on new dev containers from it for their own repositories that use it
4. Make it easier to move back-and-forth between VSO and local
5. Improve transparency about its exact contents

//cc: @lostintangent

Chuxel:
This has been resolved with PR #242

Chuxel:
Resolved by #245

PR #245, opened 2020-02-24
Default Linux VS Online Dev Container

Chuxel:
Implements #155. 

Currently uses `default` as the image name and sets the version to `0.1.0`.

### DIFF
--- a/containers/vsonline-linux/.devcontainer/devcontainer.json
+++ b/containers/vsonline-linux/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-	"name": "VS Online (Linux Default)",
+	"name": "VS Online (Linux Universal)",
 	"dockerFile": "Dockerfile",
 	"remoteUser": "vsonline",
 	"workspaceMount": "source=${localWorkspaceFolder},target=/home/vsonline/workspace,type=bind,consistency=cached",
--- a/containers/vsonline-linux/.devcontainer/devcontainer.json
+++ b/containers/vsonline-linux/.devcontainer/devcontainer.json
@@ -1,0 +1,26 @@
+{
+	"name": "VS Online (Linux Default)",
+	"dockerFile": "Dockerfile",
+	"remoteUser": "vsonline",
+	"workspaceMount": "source=${localWorkspaceFolder},target=/home/vsonline/workspace,type=bind,consistency=cached",
+	"workspaceFolder": "/home/vsonline/workspace",
+	"runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],
+
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		"MS-vsliveshare.vsliveshare",
+		"VisualStudioExptTeam.vscodeintellicode"
+	],
+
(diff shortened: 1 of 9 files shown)


Chuxel:
I should also add that right now, this just installs Live Share and IntelliCode.  Do we also want to install all the language extensions supported?

containers/vsonline-linux/test-project/test.sh
```diff
@@ -0,0 +1,54 @@
...
+    checkMultiple "$1" 1 "[ -d ""$HOME/.vscode-server/extensions/$1*"" ]" "[ -d ""$HOME/.vscode-server-insiders/extensions/$1*"" ]" "[ -d ""$HOME/.vscode-test-server/extensions/$1*"" ]"
+}
+
+# -- Actual tests - add more here --
+checkMultiple "vscode-server" 1 "[ -d ""$HOME/.vscode-server/bin"" ]" "[ -d ""$HOME/.vscode-server-insiders/bin"" ]" "[ -d ""$HOME/.vscode-test-server/bin"" ]"
+check "non-root-user" "id vsonline"
+check "/home/vsonline" [ -d "/home/vsonline" ]
+check "sudo" sudo -u vscode echo "sudo works."
+check "git" git --version
+check "command-line-tools" which top ip lsb_release
```
derekbekoe: There are a few other checks we currently run. Would be good to add those as well.

```
# Check platforms
docker run $IMAGE oryx platforms
docker run $IMAGE python --version
docker run $IMAGE dotnet --info
docker run $IMAGE node --version
docker run $IMAGE php --version

# Check expected shells
docker run $IMAGE bash --version
docker run $IMAGE fish --version
docker run $IMAGE zsh --version
```

containers/vsonline-linux/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
...
+	"remoteUser": "vsonline",
+	"workspaceMount": "source=${localWorkspaceFolder},target=/home/vsonline/workspace,type=bind,consistency=cached",
+	"workspaceFolder": "/home/vsonline/workspace",
+	"runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],
+
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		"MS-vsliveshare.vsliveshare",
+		"VisualStudioExptTeam.vscodeintellicode"
+	],
```
derekbekoe: There are a couple other extensions we currently install. Should we include those as well?
```
"ms-vsonline.vsonline",
"GitHub.vscode-pull-request-github"
```

containers/vsonline-linux/test-project/test.sh
Chuxel: Good idea - added!

containers/vsonline-linux/.devcontainer/devcontainer.json
Chuxel: Done

Chuxel:
@chrmarti and I chatted, and I think we've come around to JC's proposal of using "universal" as the image name rather than default.  We'll still have VSO in the description, which is the important part.

PR #248, opened 2020-02-26
Change the Functions 3.1 dev container to pull from apt-get

brandonh-msft:
Functions PG recently started pushing the v3 core tools to apt-get so we can pull from there now! :)

### DIFF
--- a/containers/azure-functions-dotnetcore-3.1/.devcontainer/Dockerfile
+++ b/containers/azure-functions-dotnetcore-3.1/.devcontainer/Dockerfile
@@ -39,24 +39,24 @@ RUN apt-get update \
     # [Optional] Add sudo support for the non-root user
     && apt-get install -y sudo \
     && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME\
-    && chmod 0440 /etc/sudoers.d/$USERNAME 
-
-# Install Azure Functions Core Tools v3
-RUN wget -qO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.asc.gpg \
+    && chmod 0440 /etc/sudoers.d/$USERNAME \
+    #
+    # Install Azure Functions Core Tools v3
+    && wget -qO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.asc.gpg \
     && mv microsoft.asc.gpg /etc/apt/trusted.gpg.d/ \
     && wget -q https://packages.microsoft.com/config/debian/9/prod.list \
     && mv prod.list /etc/apt/sources.list.d/microsoft-prod.list \
     && chown root:root /etc/apt/trusted.gpg.d/microsoft.asc.gpg \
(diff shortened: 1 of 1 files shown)


Chuxel:
Thanks for the PR!  Would you mind condensing the separate RUN commands into the existing one to minimize the number of layers in the generated image?  This is something we're doing in anticipation of publishing images out of the repository (like we have for a number already).

brandonh-msft:
Sure thing - I'm no container expert so this is good feedback. Making the change now.

Just so I might learn, though, the reason I did separate ones is because I viewed them as "checkpoints" for the `build` wherein if things were changed further down it could easily re-use intermediate layers and then just build the changed ones. E.g. if the URL for the core tools changed to `debian/10` in the future (I do expect it to), we wouldn't have to run the whole `install`, `groupadd`, etc above it to consume the change.

Is my understanding right? How does more RUN commands impact the publishing out to a repo? Thanks!

Chuxel:
@brandonh-msft Yes, generally. However, when uploaded to an image registry, each layer is transferred independently.  When another person downloads the image, they get each layer on its own.  Having one layer has a few advantages:

1. It minimizes total number of bytes uploaded/downloaded. Typically you have a cleanup step at the end of a Dockerfile. If this is in a separate layer, then people still download the layers before the cleanup happens.  The result that the total download size is larger.
2. It generally speeds up download time to pull a single layer since there's a one larger download rather than many little ones.

So there's a bit of a balancing act.  The big win is to have any `apt-get` commands in the same layer as the cache cleanup and autoremove of outdated packages.  In cases where a .tgz is downloaded and unzipped and moved, you'd want to have the "rm" for the tgz in the same layer for similar reasons.  That said, if you only ever use the Dockerfile locally, the main thing you save on is disk space. So in those cases, having separate RUN sections can speed the iteration process due to caching as you say.

In the case of these images, we've [started uploading some of them as images to mcr.microsoft.com](https://hub.docker.com/_/microsoft-vscode-devcontainers), so that's where layer minimization really matters.  So, we're typically having images that are pre-defined use a single layer when possible with an eye to the future.

Make sense?

brandonh-msft:
for sure - thanks!

Chuxel:
//cc: @chrmarti Since we're close to release and the existing definition is functionally similar, my thought is we merge this once we've released.  That said, it's a low risk update.  Thoughts?

brandonh-msft:
> //cc: @chrmarti Since we're close to release and the existing definition is functionally similar, my thought is we merge this once we've released. That said, it's a low risk update. Thoughts?

regarding this, two points to note:
1. The current definition (prior to this PR) was pinned to one specific version of core tools and **could not be upgraded** w/o other changes to the dockerfile due to a breaking change in the core tools package (addition of `gozip`)
2. The new definition will pick up updates from func core tools **and** works w/ the new breaking change since it's all encapsulated w/in apt-get

Ultimately the decision rests w/ your team but wanted you to be fully informed :)

Chuxel:
@brandonh-msft Ah got it - so beyond being convenient, it's actually fixing an issue.  In that regard, I think it is safe to merge even in a "point" release.

PR #250, opened 2020-02-27
Cleanup Go package cache

### LINKED ISSUE #249: Go devcontainer fails to build some Go modules when running as a non-root user
## Versions

- VSCode Version: 1.42.1
- Local OS Version: macOS Mojave 10.14.6
- Name of Dev Container Definition with Issue: go

## Steps to Reproduce

1. Create a new empty directory.
2. Open the directory in VS Code.
3. `> Remote-Containers: Add Development Container Configuration Files...`
4. Select `Go` template.
5. Open the `devcontainer/devcontainer.json` file.
6. Uncomment the `"remoteUser": "vscode"` line and put a `,` before it.
7. Create a `go.mod` file with the following content:

    ```go
    module github.com/microsoft/vscode-dev-containers/go-template-bug

    require github.com/mitchellh/go-homedir v1.1.0

    go 1.13
    ```

8. Create a `main.go` file with the following content:

    ```go
    package main

    import (
        "fmt"
        "github.com/mitchellh/go-homedir"
    )

    func main() {
        home, err := homedir.Dir()
        if err == nil {
            fmt.Println("Home directory is", home)
        } else {
            fmt.Println("Unable to resolve the home directory")
        }
    }
    ```

9. `> Remote-Containers: Reopen in Container`
10. Open the `main.go` file.
11. `> Debug: Run (Start Without Debugging)`

## Expected Outcome

Debug Console shows

```
Home directory is /home/vscode
Process exiting with code: 0
```

## Actual Outcome

Debug Console shows

```
build github.com/microsoft/vscode-dev-containers/go-template-bug: cannot load github.com/mitchellh/go-homedir: module github.com/mitchellh/go-homedir@latest found (v1.1.0), but does not contain package github.com/mitchellh/go-homedir
exit status 1
Process exiting with code: 1
```

## Analysis

The error message is quite cryptic. If we check the source code is actually there:

```
$ ls -l /go/pkg/mod/github.com/mitchellh
total 4
drwx-w--w- 2 root root 4096 Feb 27 21:58 go-homedir@v1.1.0
```

But it turns out that `go-homedir@v1.1.0` is neither readable nor openable for a non-root user. It would also be non-writable as well if not for the `chmod -R a+w /go/pkg` in the Dockerfile.

We can find a lot of other modules that would fail to build due to the permission issue:

```sh
find /go -type d -name '*@*' ! -executable 2> /dev/null
```

## Solution

All the packages in `/go` that end up in the image are leftovers from the Go tools installation. The end-user has no use for them.

If we build the Go tools in a throw-away `GOPATH` and only put the resulting binaries in the image this problem will go away.

kaharlichenko:
Installation of the Go tools necessary for VS Code integration
leaves dependency packages in GOPATH (/go).

This leads to two problems:

1. Permission problems when using the image as non-root.
2. Increase of the image size.

This commit fixes both problems by using a throw-away GOPATH
while building the necessary Go tools and moving the resulting
binaries to /usr/local/bin.

Fixes #249

### DIFF
--- a/containers/go/.devcontainer/Dockerfile
+++ b/containers/go/.devcontainer/Dockerfile
@@ -23,11 +23,11 @@ RUN apt-get update \
     # Verify git, process tools, lsb-release (common in install instructions for CLIs) installed
     && apt-get -y install git openssh-client less iproute2 procps lsb-release \
     #
-    # Install Go tools w/module support
+    # Build Go tools w/module support
     && mkdir -p /tmp/gotools \
     && cd /tmp/gotools \
-    && GO111MODULE=on go get -v golang.org/x/tools/gopls@latest 2>&1 \
-    && GO111MODULE=on go get -v \
+    && GOPATH=/tmp/gotools GO111MODULE=on go get -v golang.org/x/tools/gopls@latest 2>&1 \
+    && GOPATH=/tmp/gotools GO111MODULE=on go get -v \
         honnef.co/go/tools/...@latest \
         golang.org/x/tools/cmd/gorename@latest \
         golang.org/x/tools/cmd/goimports@latest \
@@ -49,16 +49,19 @@ RUN apt-get update \
         github.com/mgechev/revive@latest  \
         github.com/go-delve/delve/cmd/dlv@latest 2>&1 \
     #
-    # Install Go tools w/o module support
-    && go get -v github.com/alecthomas/gometalinter 2>&1 \
+    # Build Go tools w/o module support
+    && GOPATH=/tmp/gotools go get -v github.com/alecthomas/gometalinter 2>&1 \
     #
-    # Install gocode-gomod
-    && go get -x -d github.com/stamblerre/gocode 2>&1 \
-    && go build -o gocode-gomod github.com/stamblerre/gocode \
-    && mv gocode-gomod $GOPATH/bin/ \
+    # Build gocode-gomod
+    && GOPATH=/tmp/gotools go get -x -d github.com/stamblerre/gocode 2>&1 \
+    && GOPATH=/tmp/gotools go build -o gocode-gomod github.com/stamblerre/gocode \
+    #
+    # Install Go tools
+    && mv /tmp/gotools/bin/* /usr/local/bin/ \
+    && mv gocode-gomod /usr/local/bin/ \
     #
     # Install golangci-lint
-    && curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin 2>&1 \
+    && curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin 2>&1 \
     #
     # Create a non-root user to use if preferred - see https://aka.ms/vscode-remote/containers/non-root-user.
     && groupadd --gid $USER_GID $USERNAME \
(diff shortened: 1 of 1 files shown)


containers/go/.devcontainer/Dockerfile
```diff
@@ -49,16 +49,19 @@ RUN apt-get update \
...
-    # Install gocode-gomod
-    && go get -x -d github.com/stamblerre/gocode 2>&1 \
-    && go build -o gocode-gomod github.com/stamblerre/gocode \
-    && mv gocode-gomod $GOPATH/bin/ \
+    # Build gocode-gomod
+    && GOPATH=/tmp/gotools go get -x -d github.com/stamblerre/gocode 2>&1 \
+    && GOPATH=/tmp/gotools go build -o gocode-gomod github.com/stamblerre/gocode \
+    #
+    # Install Go tools
+    && mv /tmp/gotools/bin/* /usr/local/bin/ \
```
Chuxel: This seems to be generally a nice way of handling this! However, why not move these to /go/bin ?  One thing I do notice is that the Go extension will update the GOPATH location under certain circumstances, so if it's somewhere else, you'd end up having two different versions on disk.

kaharlichenko: > However, why not move these to /go/bin ?

Leaving anything in `/go/bin` defeats the whole purpose of this PR: to make it possible to work with the image as a non-root user. Consider a situation, when I want to install a newer version of the same tool. If it were in `/go/bin` it would be owned by `root` and so I won't be able to overwrite it. Exactly the thing that happened with the package cache.

Also, `/usr/local/bin` is the [standard FHS location](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html) for installing the binaries which are not maintained by the package manager, while `/go/bin` is some random path that upstream authors decided to use for the `golang:1` image.

> One thing I do notice is that the Go extension will update binaries in GOPATH location under certain circumstances (e.g. if you switch to gopls the first time), so if it's somewhere else, you'd end up having two different versions on disk.

I think having two versions of the same tool is perfectly fine. That is what you usually do when the package manager provides an ancient version of the tool that you need. You just install the new version to your `$HOME/.local/bin` and enjoy the cool features 😄 .

In fact, `$PATH` within the image is set to `/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`. So if we put all the tools to `/usr/local/bin` and VS Code decides to install something to `$GOPATH` it will take precedence and everything will work smoothly.

Chuxel: Apologies for the delayed response.  I was out of office last week.

I think we'll also need to chown or chmod the `/go` folder anyway won't we?  Anything the user installs ad-hoc will still land in that folder.  We could just `chown -R vscode /go`.

kaharlichenko: The changed `Dockerfile` doesn't write anything to the `/go` directory.

Now let's take a look at what upstream `golang:1` image does to permissions for `/go`:

```
$ docker run --rm golang:1 ls -ld /go
drwxrwxrwx 4 root root 4096 Feb 14 01:21 /go
$ docker run --rm golang:1 ls -lR /go
/go:
total 8
drwxrwxrwx 2 root root 4096 Feb 14 01:21 bin
drwxrwxrwx 2 root root 4096 Feb 14 01:21 src

/go/bin:
total 0

/go/src:
total 0
```

As you can see the upstream image allows everyone to write to `/go` and its subdirectories, so no `chrown` or `chmod` is necessary.

Chuxel: Good catch.  Ok, LGTM!

PR #251, opened 2020-03-04
helm 3 support for kubernetes-helm template

sebastianspross:
The container template `kubernetes-helm` is still using helm 2. I changed it using helm 3 and deleted the optional steps from the docu where it says to call `helm init` which was needed to setup the tiller in the K8s cluster


Chuxel:
Apologies for the delay, I was OOF last week.  Thanks for the contribution.  LGTM!

PR #257, opened 2020-03-13
Update references to C# extension.

fiveisprime:
Updated all references to the C# extension, which was recently
transferred to a new publisher ID.


Chuxel:
LGTM!

PR #260, opened 2020-03-16
Remove libssl fixed version

bhack:


bhack:
Also can we remove some other dependencies seems that we are installing openjdk etc..?

bhack:
Ok, thanks probably was another layer to introduce Jdk in my test. 
But this PR is still valid cause fix hardcoded libssl.

Chuxel:
This installs 1.0.x (if available in the distro) and 1.1.  Is there a problem that installing 1.1 is causing?  What advantage does removing it provide?

bhack:
It conflicts with this condition management https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/common-debian.sh#L54. 
See also https://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=libssl

bhack:
@Chuxel so how you are going to support 16.04 in https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/common-debian.sh#L53

bhack:
I meant have you seen my previous [libssl packages list](https://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=libssl) in 16.04?

bhack:
`libssl1.1` was introduced in `18.04 bionic` see https://packages.ubuntu.com/search?suite=bionic&searchon=names&keywords=libssl

Chuxel:
Ah, that's a separate issue.  Nothing that uses the script right now uses Ubuntu 16.04.  That said, we could make the installation of 1.1 conditional like libssl1.0.x to resolve that.

bhack:
Do you like `--ignore-missing` or it is a too bad hack?

bhack:
Can we use `libssl1.[0-1].[0-9]` and remove the bash stuff?

Chuxel:
@bhack There have been problems we've seen in the past with not being specific here particularly if someone adds a backport to install something. This code was introduced to mitigate that problem.  The check is simple and has avoided any issues related to it since.

This should do it.
```bash
if [[ ! -z $(apt-cache --names-only search ^libssl1.1$) ]]; then
        apt-get -y install  --no-install-recommends libssl1.1
 fi
```

bhack:
Do you like now?

Chuxel:
LGTM!

PR #263, opened 2020-03-22
Fix paths

mkontani:
I found some incorrect paths and typo.
- change from lts to version for javascript
- some typofix


Chuxel:
Thanks!!

PR #266, opened 2020-03-28
Move Azure ML to Ubuntu base due to incompat with Debian

### LINKED ISSUE #265: azure-machine-learning-python-3 not supported by azureml-dataprep part of azureml-sdk
- VSCode insider
- Windows 10 Pro
- azure-machine-learning-python-3

The image 'azure-machine-learning-python-3' derives from continuum:anaconda3, which derives from Debian (https://github.com/ContinuumIO/docker-images/blob/master/anaconda3/debian/Dockerfile).

The supported linux distributions for azureml-dataprep are listed here and Debian is not one of them:
https://docs.microsoft.com/en-us/azure/machine-learning/how-to-train-with-datasets#prerequisites

Steps to Reproduce:

1. use the azureml-sdk to register a Dataset in an azure ml workspace (https://docs.microsoft.com/en-us/learn/modules/work-with-data-in-aml/4-datasets)
2. This will cause the following issue: "NotImplementedError: Unsupported Linux distribution"

It seems the Linux distribution check is done very naively and does not allow Debian although Ubuntu is based on it. The same reportedly holds for Linux Mint based on Ubuntu 18:04 (https://github.com/Azure/MachineLearningNotebooks/issues/713) as it too is 'not supported'.

I have copied the contents of continuumio/anaconda3 and moved to Ubuntu:18.04 as a base image. This solves the issue and lets me register an azure-ml Dataset.

I would like to see the official vscode devcontainer for azure-ml actually supporting azureml-sdk and azureml-dataprep specifically. Although I did not test it I checked whether the suggested alternate base image (see comments in Dockerfile)  'python:3' would help and as far as I can see these are Debian only.  For now I would like to see full Conda and azureml-sdk support as these two are very relevant in the machine learning domain.

Chuxel:
Resolves #265

### DIFF
--- a/containers/azure-machine-learning-python-3/.devcontainer/Dockerfile
+++ b/containers/azure-machine-learning-python-3/.devcontainer/Dockerfile
@@ -52,8 +52,6 @@ RUN apt-get update \
     && echo "Downloading Anaconda..." \
     && su --login -c "wget -q https://repo.anaconda.com/archive/Anaconda3-${ANACONDA_VERSION}-Linux-x86_64.sh -O /tmp/anaconda-install.sh \
         && /bin/bash /tmp/anaconda-install.sh -u -b -p ${CONDA_INSTALL_PATH}" ${USERNAME} 2>&1  \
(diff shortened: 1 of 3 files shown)


Chuxel:
Merging since this resolves the immediate issue. Will submit a second PR as needed for #265

PR #273, opened 2020-04-05
Update to Azure Functions Python to 3.8

### LINKED ISSUE #270: Azure Functions & Python 3.8
I just realize that the Azure Functions & Python 3 included in VSCode only has version 2.x of Azure Function core tools and version 3.6 of Python.

It will be usefull to update those containers so it support Python 3.8 and core tools 3.x

I have tried myself doing the upgrade and I was sucessful changing to Python 3.7.7 and core tools 3.x but the switch to 3.8 is a little more tricky because there is no version of python 3.8-stretch only buster or alpine and I tried buster and there was problem without trivial solution.

anthonychu:
* Update to Python 3.8 using `mcr.microsoft.com/vscode/devcontainers/python:3.8`
* Update Azure Functions Core Tools to version 3

Resolves #270


Chuxel:
LGTM! Thanks!

PR #274, opened 2020-04-05
Azure Functions Node 10 - Update to Core Tools 3

anthonychu:


containers/azure-functions-node-10/.devcontainer/Dockerfile
```diff
@@ -37,6 +37,9 @@ RUN apt-get update \
     && apt-get autoremove -y \
     && apt-get clean -y \
     && rm -rf /var/lib/apt/lists/*
-        
+
+# Azure Functions Core Tools needs a place to save data
+ENV XDG_DATA_HOME=/usr/local/share
```
Chuxel, line 42: This should get updated to work with the non-root user like the python definition too, correct?

anthonychu: Yep. I’ll copy what I did in the Python one.

Chuxel:
LGTM!

PR #276, opened 2020-04-05
Azure Functions Python 3 - Remove unnecessary apt source

anthonychu:
Turns out there's no need to add the extra apt-get source.

### DIFF
--- a/containers/azure-functions-python-3/.devcontainer/Dockerfile
+++ b/containers/azure-functions-python-3/.devcontainer/Dockerfile
@@ -28,10 +28,12 @@ RUN \
     # Clean up
     && apt-get autoremove -y \
     && apt-get clean -y \
-    && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


containers/azure-functions-python-3/.devcontainer/Dockerfile
```diff
@@ -30,6 +29,9 @@ RUN \
     && apt-get autoremove -y \
     && apt-get clean -y \
     && rm -rf /var/lib/apt/lists/*
-        
+
+# Azure Functions Core Tools needs a place to save data
+ENV XDG_DATA_HOME=/usr/local/share
```
Chuxel, line 34: Quick question - When connecting as the non-root vscode user, wouldn't this result in an access denied?

anthonychu: Good catch. I've updated it to the a non-root user.

Chuxel:
LGTM!

PR #278, opened 2020-04-08
Migrate publish to service principal, only update MCR if version updates

Chuxel:
PR switches to a service principal to interact with ACR and enables ACR lookups to see if something is already published. If a version is already published, skip it (though this does not apply to dev tag).


Chuxel:
Holding on VS Code release to avoid unexpected impacts.

Chuxel:
Confirmed working in a fork, merging now that release is out.

PR #284, opened 2020-04-15
Added folder for kubernetes helm with docker compose. Include docker-…

amura:
I have created this PR to include a docker-compose configuration for when running Kubernetes and Helm inside docker. This will make it easier for those who are not only using a dockerfile configuration as all the mounts are now done inside the docker-compose. I have also included a post create command that runs the existing test file so it demonstrates successful configuration at the end.


Chuxel:
Thanks for the contribution!  A couple of things that would be great if you could do:

1. The way this repository works is that all the contents for the dev container should be in the `.devcontainer` folder.  This appears to depend on `docker-compose.yml` a level up which the person using the definition may already have on disk and we wouldn't want to overwrite.
2. Could you link to the best way to contact you from the README.md under contributor in case there are questions for you?

amura:
@Chuxel thanks for reviewing this. I have made the updates as you suggested!

amura:
Accidental close

bamurtaugh:
Thanks again for opening this and for the discussion so far.

PR #288, opened 2020-04-17
fix(spelling): remove "s"

gjungb:
the official name is *PostgreSQL* ;-)


Chuxel:
DOH! Thanks for catching it.

PR #289, opened 2020-04-17
Update Dockerfile

edgonmsft:
Update to create predefined docker group in the image.

We can merge it after the changes in the services go in.


Chuxel:
LGTM!  BTW - Just to make sure I mention it, you'll need to bump up the version in the definition-build.json file for when you want to deploy this outside (or other changes) of dev.  Things are setup now so it won't overwrite the same version number to avoid unexpected upates to the VSO image in the service because of upstream patches for things that are installed.

Issue #253, opened 2020-03-09
PHP & MySQL docker-compose.yml


KROSF:
something similar to [devilbox](https://github.com/cytopia/devilbox) could be great (a tiny version)

Chuxel:
Thanks for the feature request!  Happy to take a look at a PR if you'd like to submit one.

Chuxel:
@ThoolooExpress contributed a PHP7 & MariaDB definition! Check it out.

PR #292, opened 2020-04-19
Add PHP 7 & MariaDB

ThoolooExpress:
Add a devcontainer definition for PHP 7 and MariaDB.  This is in response to #253, which requested a PHP and MySQL definition.

This definition is using MariaDB instead of MySQL because it is based off of a configuration I'd already been using, and because the version of default-mysql-client supplied by PHP 7's version of Debian is not compatible with the authentication plugin used by the latest version of MySQL.  The PHP MySQL drivers are also compatible with MariaDB so this shouldn't matter from the PHP side.


Chuxel:
This looks great - thanks so much for contributing?  Would you mind including a link to the best way to contact you if there are questions in the contributor's field?  (If the answer is GitHub, you could just link there.)

Chuxel:
Thanks so much!

PR #293, opened 2020-04-23
Add `Angular with Node 14` container.

jaideepheer:
Signed-off-by: Jaideep Singh Heer <jaideesinghheer@gmail.com>

### DIFF
--- a/containers/angular-node-14/.devcontainer/Dockerfile
+++ b/containers/angular-node-14/.devcontainer/Dockerfile
@@ -1,0 +1,14 @@
+# Common stage
+FROM node:14 AS CommonStage
+WORKDIR /workspace
+USER root
+RUN npm config --global set user root
+
+# Development stage
+FROM CommonStage AS DevStage
+EXPOSE 4000 4200 4433 5000 8080 9876
+RUN yarn global add eslint tslint typescript @angular/cli
+RUN ng config -g cli.packageManager yarn
+
+# Deployment stage <enter your deployment logic>
+FROM CommonStage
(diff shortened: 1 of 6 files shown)


Chuxel:
Thanks for the pull request!  A few comments questions:

1. There's already a definition for Node.js 12 and it definitely makes sense to add one for 14 now that it is out.  However, this definition deviates pretty significantly from the Node 12 version. Any particular reason why?  For example, we include a "vscode" user in all definitions since this is important for Linux users.
2. Naming wise, it seems like this definition is more `typescript-node-14-angluar` isn't it? Angular is the library in use, but TypeScript is more the language.
3. It looks like you are using a multi-stage Dockerfile. Given we're pre-building the Node images to speed up their use, we could tweak this a bit to I think get the same general idea across and also include a "builder" stage.  For example, if we stick with `typescript-node-12`, it could be something along these lines:

```Dockerfile
FROM mcr.microsoft.com/vscode/devcontainers/typescript-node:12 AS development

# Build steps go here
FROM development as builder
WORKDIR /app
COPY src/ *.json ./
RUN yarn install \
    && yarn compile \
    #  Just install prod dependencies
    && yarn install --prod

# Actual production environment setup goes here
FROM node:12-slim AS production
WORKDIR /app
COPY --from=builder /app/out/ ./out/
COPY --from=builder /app/node_modules/ ./node_modules/
COPY --from=builder /app/package.json .
EXPOSE 3000
ENTRYPOINT [ "/bin/bash", "-c" ]
CMD [ "npm start" ]
```

PR #306, opened 2020-04-28
Update oryx version

edgonmsft:
New version of oryx. Updated the symlink file for new dotnet versions.

### DIFF
--- a/containers/vsonline-linux/.devcontainer/Dockerfile
+++ b/containers/vsonline-linux/.devcontainer/Dockerfile
@@ -2,7 +2,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
-FROM buildpack-deps:stretch@sha256:d095aae2ecefdc786823e02b2cb0abc62b432202a49abd716c12cce37aee702b as builddeps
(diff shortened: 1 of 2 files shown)


Chuxel:
Should we also bump the version number in `definition-build.json` so this goes out when @chrmarti cuts another release for the upcoming 1.45 release?

edgonmsft:
> Should we also bump the version number in `definition-build.json` so this goes out when @chrmarti cuts another release for the upcoming 1.45 release?

Yes although I just wanted to test this one as dev first.

Chuxel:
Also, just doing a smoke test build, I'm seeing the following error (I am running on macOS):

```text
Processing triggers for dbus (1.10.28-0+deb9u1) ...
Errors were encountered while processing:
 python3-pycurl
 python3-apt
 python3-gi
 unattended-upgrades
 python3-software-properties
 software-properties-common
 python3-dbus
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

Chuxel:
Same thing happens on Linux - this would fail in CI too I think.

I also actually am noticing the Oryx version in the FROM statement wasn't updated if that was intended... maybe that's why it is failing.  Otherwise, I'm guessing the builddeps hack is still needed since I noticed that got removed. The commit for it would likely just need to be updated to match the new Oryx version. Those packages likely depend on the OS python install.

edgonmsft:
Yep sorry about that, I've updated the pr.

Chuxel:
LGTM!

PR #310, opened 2020-05-01
Security patching for 4/30/2020

Chuxel:
@edgonmsft @jkeech There are a number of security alerts that came across due to new issues with `git`. This affects the VS Codespaces image along several others.  In addition, older versions of the codespaces image have issues with curl and postgres.

This PR triggers patching and adds an upgrade to `git` to the universal image.  All others always upgrade on release and therefore do not need to be updated.

Any objections to moving forward?


jkeech:
@Chuxel, no concerns from me. Thanks for setting this up! For my own reference, how do you determine the set of image ids/SHAs to patch?

VSCS will pick up the latest images soon after they are published (we always reference a specific image version for repeatability).

Chuxel:
The S360 report tells you. These are in the "Asset Id" column. Another column tells you the exact library it found along and another the security alert ID. I export into excel and sort by that, quickly check the report to see what needs to be done ... which is almost always updating the package, then create the Dockerfile to test with a few SHAs.

The utility then pulls down the SHA, determines its tags, updates, retags, and then deletes the old SHA. (MCR would still have it for people pinning, but they're removed out of our own repo so we don't end up using them by accident and to clean up the report.)

There were a lot more here than usual because this affected git.

PR #311, opened 2020-05-01
Update Java8 Container image

TsuyoshiUshio:
I discussed with @Chuxel about the Azure Functions DevContainer images, I create this PR. 

The Azure Functions Core Tools v3 with Java8 and Azure CLI  image is published and continuously publish it now. 

I change the base image to that one. 

## What I did 
* Dockerfile references the new Azure Functions Core Tools with Java images.

## Pros and Cons
* Customer doesn't need to build Dockerfile a lot. The bootup time is faster than the last version.
* Visibility of the Dockerfile is down. I add comment to point the Dockerfile that we use.


Chuxel:
Looks great!  Now that this is pre-built, you can use the library scirpts if you so choose to install things like zsh in the container or not as you see fit since it won't affect download times as much.

This will download and run the script and create the vscode user so you don't even need to do that:

```Dockerfile
ARG INSTALL_ZSH=true
ARG USERNAME=vscode
ARG USER_UID=1000
ARG USER_GID=$USER_UID

RUN curl -ssL https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/script-library/common-debian.sh -o tmp/common-script.sh \
    && /bin/bash /tmp/common-setup.sh "$INSTALL_ZSH" "$USERNAME" "$USER_UID" "$USER_GID" \
    && rm /tmp/common-setup.sh
```

TsuyoshiUshio:
Hi @Chuxel 
It is beautiful. I'll absolutely consider that! :)

Issue #294, opened 2020-04-24
Consolidate version specific definitions with a build arg, enable expansion in VS Code


Chuxel:
Now that build args are supported in devcontainer.json we can consolidate multiple versions of the same definition that vary only in the version of one piece of software.  Examples include the donetcore definitions, javascript/typescript-node, azure-terraform and more.

For those that include image builds, we can follow the model that is used in python-3 with a VARIANT arg that results in multiple image builds.

To round this out, we can expand these variants in the list that appears in VS Code when using commands like **Reopen in container** or **Add Development Container Configuration Files**.

To-do list:

- [x] Ubuntu 18.04, 20.04 (`base` image)
- [x] Debian 9, 10 (`base` image)
- [x] JS and TS Node.js 10, 12, 14 (`javascript-node`, `typescript-node` image)
- [x] .NET Core 2.1, 3.0, 3.1 (C# and F# definitions)
- [x] Java
- [x] Terraform 0.11, 0.12 (Azure Terraform)

Python is already set up this way.

Chuxel:
Started with Ubuntu 20.04 for #303 in this branch: https://github.com/microsoft/vscode-dev-containers/tree/clantz/multi-tag-and-variant-expansion/containers

Changes were needed to the CI system to support version specific tags (e.g. "bionic"). This branch also gets metadata into the npm package for use in VS Code or Codespaces or wherever.

felipecassiors:
I submitted a pull request, wasn't aware of your current work. Should it be closed? 

I would also want to say that might be also a good idea to change the base image from `ubuntu:20.04` to `buildpack-deps:focal` instead. And maybe reconsider the branding "git" on the container, because, isn't this too obvious?

Chuxel:
@felipecassiors Yeah I am looking into some refactoring that will allow us to just change a property to generate another version of a given image.  I plan to start with Ubuntu 20.04 since it's top of mind.

In terms of using `buildpack-deps:focal` - The only downside is that image is 695MB where the fully built image here is around 252MB.  Are there things critically missing you think should be in the image?

I'm fine with removing the `-git`.

felipecassiors:
@Chuxel the new approach sounds cool! 

[...]

PR #317, opened 2020-05-07
Add Ubuntu 20.04 container

### LINKED ISSUE #303: Add Ubuntu 20.04 container
With the release of Ubuntu 20.04 LTS, I believe a good amount of developers would like to move their devcontainer from the 18.04 to this new LTS version as well.

I suggest to use [buildpack-deps:focal](https://hub.docker.com/_/buildpack-deps/) as the base, since it is an official Docker image based on ubuntu:20.04 with many common and basic dependencies used by developers already installed, such as `git` and `curl`.

felipecassiors:
This new container uses `buildpack-deps:focal` as base, and this
commit duplicates the `ubuntu-18.04-git` folder and rename every
reference of the old container. Plus, it also rebrand the image
without the `-git`, because is more than obvious that it will have
git installed.

Fixes #303


felipecassiors:
Didn't notice, but this seems to be already included in #294 and its branch (without pull request): clantz/multi-tag-and-variant-expansion/containers.

PR #320, opened 2020-05-09
Missing multi-line continuation characters

benc-uk:
The last part of the command fails due to missing back-slashes


Chuxel:
Good catch!  Thanks much!

PR #321, opened 2020-05-10
Updating Core Tools to v3 in PS Functions image

mattchenderson:
I noticed an issue when using this devcontainer in VS Codespaces. I created a new Functions project and HTTP-triggered function. The functions worker process gets into a bad state, prompting retries from the host, which eventually gives up can exits: `Exceeded language worker restart retry count for runtime:powershell. Shutting down Functions Host`

Making this change to update to v3 fixes the issue.


Chuxel:
Thanks for the PR!  LGTM!

PR #332, opened 2020-05-14
Fix missing URL and update comment Development Container Script

TsuyoshiUshio:
Hi @Chuxel 

I update the java8 one with two things. 

* Fix missing URL for the Dockerfile (someone change the URL for the target repo)
* Add comment for the Development Container Script 

Our base image is for Azure Functions Core Tools and Azure CLI.  I wanted to separate the concern. 
I just add the comment for the Development Container Script so that  customer can comment out. 

I also consider if we can do it with a switch instead of the comment. However, it can hit a cache and ignore it. For this reason, I make it as comment. 

Any feedback is welcome. 

# I'll update the azure functions images one by one. :)  Node, Python, Dotnet 3 will be updated by the other PRs. Java11 coming soon as well. :)

### DIFF
--- a/containers/azure-functions-java-8/.devcontainer/Dockerfile
+++ b/containers/azure-functions-java-8/.devcontainer/Dockerfile
@@ -3,5 +3,23 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 
-# ImageURL: https://github.com/Azure/azure-functions-docker/blob/master/host/3.0/buster/amd64/java/java8-core-tools.Dockerfile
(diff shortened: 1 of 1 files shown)


Chuxel:
LGTM!  One note is you can switch where the DEBIAN_FRONTEND is placed to compress it even further.  This works:

```
# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
#     && curl -ssL https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/script-library/common-debian.sh -o /tmp/common-script.sh \
#     && /bin/bash /tmp/common-script.sh "$INSTALL_ZSH" "$USERNAME" "$USER_UID" "$USER_GID" \
#     && rm /tmp/common-script.sh  
```

That script now allows you to also disable package updating which is an additional arg, but it's optional so it works as is.

PR #333, opened 2020-05-15
Move Azure Functions container to Azure Functions Official Images

TsuyoshiUshio:
Hi @Chuxel , 

I published the following Azure Functions images. So I switch to the Azure Functions Docker images. 

* Dotnet Core 3.1
* Node 10
* Node 12
* Python 3.8

These images are tested by running the Readme process.  (Open the DevContainer then, create project, then debug run. check if the break point works if I send a request). 

I also add a comment of  Development Container Script. It is the same as the Java one.  
https://github.com/microsoft/vscode-dev-containers/pull/332


TsuyoshiUshio:
@Chuxel Thanks! I'll removed the comments. I didn't know that. Thank you for sharing!
If it is ok, please merge this. :)

Chuxel:
LGTM!

PR #335, opened 2020-05-16
Remove vscode user from Puppet devcontainer

jpogran:
This PR removes the reference to the vscode user from the devcontainer, as the [`puppet/pdk`](https://github.com/puppetlabs/pdk-docker/blob/cc4ff84497432057d4ae4a5b2e8f45206b26f4e7/Dockerfile) docker image does not have such a user.

At the time of this commit, the [`puppet/pdk`](https://github.com/puppetlabs/pdk-docker/blob/cc4ff84497432057d4ae4a5b2e8f45206b26f4e7/Dockerfile) image does not create a
non-root user. If/when it does, this can be re-added.


Chuxel:
LGTM!  Thanks.

PR #336, opened 2020-05-18
Non-root Docker in Docker README changes

senglin:
I walked through this tutorial and found that this sample code did not work for me.

`*.socket` should have been `*.sock` 
and
Line 83 is actually ensuring that the `docker-host.sock` file exists before creating a symbolic link.
and
Line 90 is actually removing the symbolic link created above if that script gets executed. `sudo socat` will fail if `docker.sock` already exist.s

Let me know if this is incorrect.

I did not test out the other scenarios (docker compose + kubernetes).


Chuxel:
Thanks!  I must have had autocomplete adding "et" when typing this up.  Appreciate it!

PR #341, opened 2020-05-19
Fix Comments in docker-compose devcontainer.json

0916dhkim:
Fix a typo.
Keep instructions about commented properties consistent.


Chuxel:
Good catch. Thanks!

PR #342, opened 2020-05-19
Java11 Dockerfile to use common-debian

senglin:
This PR fixes errors such as the one below:
```
LC_ALL: cannot change locale (en_US.UTF-8)
```

The PR also brings the Dockerfile towards the standardize convention of using common-debian.sh for all docker images based on the debian images.

### DIFF
--- a/containers/java-11/.devcontainer/Dockerfile
+++ b/containers/java-11/.devcontainer/Dockerfile
@@ -13,21 +13,22 @@ ARG USERNAME=vscode
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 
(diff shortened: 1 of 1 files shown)


Chuxel:
@testforstephen This looks like a good approach since it moves to using the common script. The same thing could be done for the other Java definitions. Thoughts?

testforstephen:
> @testforstephen This looks like a good approach since it moves to using the common script. The same thing could be done for the other Java definitions. Thoughts?

Agree on it.

Chuxel:
LGTM!  Thanks for the contribution!!

PR #347, opened 2020-05-26
Tensorflow addons

bhack:
Add support to Tensorflow addons:

See https://github.com/tensorflow/addons/pull/1309 and https://github.com/tensorflow/addons/pull/1888

### DIFF
--- a/repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
+++ b/repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
@@ -3,7 +3,8 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 
-FROM tfaddons/dev_container:latest
+ARG IMAGE_TYPE=latest-cpu
+FROM tfaddons/dev_container:$IMAGE_TYPE
 
 # This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
 # property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
--- a/repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
+++ b/repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 
-FROM tfaddons/tensorflow:2.1.0-custom-op-gpu-ubuntu16-minimal
+FROM tfaddons/dev_container:latest
 
 # This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
 # property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
--- a/repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
+++ b/repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
@@ -15,7 +15,7 @@ ARG USER_GID=$USER_UID
 
 # Options for common package install script - SHA updated on release
 ARG INSTALL_ZSH="false"
-ARG UPGRADE_PACKAGES="true"
+ARG UPGRADE_PACKAGES="false"
 ARG COMMON_SCRIPT_SOURCE="https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/script-library/common-debian.sh"
 ARG COMMON_SCRIPT_SHA="dev-mode"
 
--- a/repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
+++ b/repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
@@ -1,0 +1,42 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+FROM tfaddons/tensorflow:2.1.0-custom-op-gpu-ubuntu16-minimal
+
+# This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
+# property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
+# will be updated to match your local UID/GID (when using the dockerFile property).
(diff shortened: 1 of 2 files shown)


bhack:
As you have seen in the mentioned ticket we would like to have CPU and GPU separated configs but I think is not possible currently if you check https://github.com/microsoft/vscode-remote-release/issues/2143#issuecomment-600087021

repository-containers/github.com/tensorflow/addons/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
+# will be updated to match your local UID/GID (when using the dockerFile property).
+# See https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Options for common package install script - SHA updated on release
+ARG INSTALL_ZSH="false"
+ARG UPGRADE_PACKAGES="true"
```
bhack, line 18: @seanpmorgan Do we want to update the package by default or not?

seanpmorgan: I'd prefer they not be updated in this Dockerfile but rather done in tfaddons/dev_container if needed

bhack:
We are waiting to check test and test discover to see if we could enable `"python.testing.pytestEnabled": true,`

bhack:
@Chuxel is there a way to differentiate between CPU and GPU `.devcontainer`? Or are we still blocked by https://github.com/microsoft/vscode-remote-release/issues/2143?

Chuxel:
@bhack I would use a build arg if you need to modify the contents of the Dockerfile and add a comment into devcontainer.json about switching between the two.

bhack:
@Chuxel need we to comment and uncomment also https://github.com/microsoft/vscode-remote-release/issues/345#issuecomment-493271078 right?

Chuxel:
@bhack If there's arguments you need to pass in, yes.  The .NET Core definition is an example of this being done - in this case for SSL support: https://github.com/microsoft/vscode-dev-containers/blob/master/containers/dotnetcore-3.1/.devcontainer/devcontainer.json

repository-containers/github.com/tensorflow/addons/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,41 @@
...
+	"dockerFile": "Dockerfile",
+	
+    	// Uncomment to enable Nvidia support
+	// "runArgs": ["--runtime=nvidia"],
+
+	// Set *default* container specific settings.json values on container create.
+	"settings": { 
+		"terminal.integrated.shell.linux": null,
+		"python.formatting.provider": "black",
+		"python.testing.pytestEnabled": true,
```
bhack, line 13: @seanpmorgan Let me know if we could enable pytest by default or the discovery it is not working.

bhack:
@Chuxel I have another question how we could automate this manual commenting/uncommenting activity when we let the user to use this for Github codespaces?

Chuxel:
VS Codespaces (and by extension Codespaces in GitHub) will be implementing the ability to iterate on a devcontainer.json and Dockerfile like the Remote - Containers extension over time. It's just not quite there yet.

bhack:
@Chuxel do you mean that they have in the roadmap multiple alternative devcontainer.json and Dockerfile support for the same repo?

Chuxel:
@bhack You'll be able to uncomment and just rebuild like you can with Remote - Containers. Step one is getting to feature parity.

bhack:
@Chuxel Thanks. We could still host `.devcontainer` here or it will be mandatory to be hosted in our repo root for Codespaces?

Chuxel:
Yeah, currently Codespaces does not pull from this repo, but this is part of the work we'll collaborate with that team on as iterating on a container starts to be available.  We'd want the same list of definitions from this repo there as well, so that would be the natural time to include it.

repository-containers/github.com/tensorflow/addons/.devcontainer/devcontainer.json
bhack: The integration is commented by default cause vscode-python upstream bugs with pytest:
https://github.com/microsoft/vscode-python/issues/8678
https://github.com/microsoft/vscode-python/issues/9553

bhack:
@seanpmorgan I've added the build-arg `IMAGE_TYPE` to switch between CPU and GPU images but we need to push a CPU image on Dockerhub.

bhack:
@Chuxel as you see we need to comment and uncomment two sections to switch from CPU to GPU. IT is not the best UX solution but I don't find something better with the current features.

repository-containers/github.com/tensorflow/addons/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,59 @@
...
+	// "build": { 
+	//		"args": { 
+	//			"IMAGE_TYPE": "latest"
+	//		} 
+	//},
+
+	"dockerFile": "Dockerfile",
+	
+	"remoteEnv": {
+		TF_CPP_MIN_LOG_LEVEL": "1"
```
Chuxel, line 14: Missing quote - `TF_CPP_MIN_LOG_LEVEL"` should be `"TF_CPP_MIN_LOG_LEVEL"`

repository-containers/github.com/tensorflow/addons/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,59 @@
...
+        	"ms-vscode.cpptools"
+    	],
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "uname -a",
+
+	// Uncomment to use Docker from inside the container. See https://aka.ms/vscode-remote/samples/docker-in-docker.
```
Chuxel, line 45: Is this needed in this case?  If not, you can remove these two lines.

bhack:
It is ok for me. @seanpmorgan when you can give a pass with the GPU options we could merge this.

Chuxel:
@seanpmorgan @bhack Let me know when you think this is to the point it should be merged.

bhack:
> Yeah, currently Codespaces does not pull from this repo, but this is part of the work we'll collaborate with that team on as iterating on a container starts to be available. We'd want the same list of definitions from this repo there as well, so that would be the natural time to include it.

@Chuxel any news on this specific point? Where can I track progress on this? Any ticket?

bhack:
@Chuxel gently ping

PR #351, opened 2020-05-28
Upgrade to Java 14

akaroml:
Signed-off-by: Rome Li <rome.li@microsoft.com>

Here to propose the "LTS + Latest" combination for Java containers. Now Java has two LTS versions: 8 + 11, so the corresponding containers won't need any change.

And this changelist is about changing from Java 12 to 14, which is the actual "latest".


Chuxel:
LGTM!

PR #354, opened 2020-05-29
Rename "Docker in Docker" to "Docker from Docker"

Chuxel:
In a recent conversation with @jkeech, he correctly pointed out that technically this is not true "Docker in Docker" as the docker daemon is not running inside the container.  There are also base images like `docker` that actually do full dind.  

To avoid confusion, we can rename these to "Docker from Docker" and possibly add one that does full docker-in-docker even if it's generally not recommended due to the perf implications ... particularly when using Docker Desktop.

Docs will also be updated to reflect this change.


jkeech:
@chuxel, do the aka.ms links update automatically through some script, or do you have to tweak those manually?

Chuxel:
@jkeech I created new links and I'll point the docker-in-docker flavors to the docker-from-docker links once I've merged.

PR #355, opened 2020-05-29
Allow use of dotnet tools in default Codespaces 

dmgardiner25:
https://github.com/MicrosoftDocs/vsonline/issues/472


edgonmsft:
Wait a little bit before merging we have another change.

Chuxel:
LGTM!  Let me know if we should bump up the version so this goes with the next release. I assume you want to test first.

edgonmsft:
Yep, I'll bump the version as soon as we test.

PR #356, opened 2020-05-30
Create a container definition for swift-5.2

cloudnull:
This change creates a container definition for swift language version 5.2.

Signed-off-by: Kevin Carter <kevin@cloudnull.com>

### DIFF
--- a/containers/swift-5.2/.devcontainer/Dockerfile
+++ b/containers/swift-5.2/.devcontainer/Dockerfile
@@ -1,0 +1,46 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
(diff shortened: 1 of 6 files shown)


Chuxel:
Thanks for submitting this! Apologies for the delay - it's been a bit of a week on multiple fronts. 🤪

Out of curiosity, one thing we could do here instead is adding a build argument to the Dockerfile that would allow developers to pick their swift version.  We could then just change `swift-4` to `switft`.  e.g.

```Dockerfile
ARG VARIANT="5"
FROM swift:${VARIANT}
```
Then in `devcontainer.json`:

```json
"build": {
    "dockerfile": "Dockerfile",
   "args": {
       "VARIANT": "5"
   }
}
```

Would that potentially be the right thing to do here?  I've made this change for Python and Ubuntu, and plan to do it for others.

cloudnull:
@Chuxel Good suggestion, I'll push an update with that change a little later today.

Chuxel:
I'll merge this in and then make the change I suggest above post-merge.

Issue #322, opened 2020-05-10
Test failing in codespaces-linux - php not in path


Chuxel:
Since updating to the latest oryx image, the `codespaces-linux` image no longer appears to have PHP in the path.  We should either remove it from `test.sh` or correct that.

```text
codespace:~/workspace$ test-project/test.sh 

🧪  Testing vscode-server: [ 1 -ge 1 ]
🏆  Passed!

🧪  Testing non-root-user: id codespace
uid=1000(codespace) gid=1000(codespace) groups=1000(codespace),800(docker)
🏆  Passed!

🧪  Testing /home/codespace: [ -d /home/codespace ]
🏆  Passed!

🧪  Testing sudo: sudo -u codespace echo sudo works.
sudo works.
🏆  Passed!

🧪  Testing git: git --version
git version 2.11.0
🏆  Passed!

🧪  Testing command-line-tools: which top ip lsb_release
/usr/bin/top
/sbin/ip
/usr/bin/lsb_release
🏆  Passed!

🧪  Testing oryx: oryx platforms
Platform  : dotnet
Versions  : 1.0.16
            1.1.13
            2.0.9
            2.1.17
            2.2.7
            3.0.3
            3.1.3
Properties: project - Relative path of the project file to build.

Platform  : nodejs
Versions  : 4.4.7
            4.5.0
            4.8.0
            6.2.2
            6.6.0
            6.9.3
            6.10.3
            6.11.0
            6.17.1
            8.0.0
            8.1.4
            8.2.1
            8.8.1
            8.9.4
            8.11.2
            8.12.0
            8.15.1
            8.17.0
            9.4.0
            10.1.0
            10.10.0
            10.14.2
            10.16.3
            10.19.0
            12.16.1
Properties: npm_registry_url - Custom npm registry URL. Will be written to .npmrc during the build.
            compress_node_modules - Indicates how and if 'node_modules' folder should be compressed into a single file in the output folder. Options are 'zip', and 'tar-gz'. Default is to not compress. If this option is used, when running the app the node_modules folder must be extracted from this file.
            prune_dev_dependencies - When using different intermediate and output folders, only the prod dependencies are copied to the output. Options are 'true', blank (same meaning as 'true'), and 'false'. Default is false.
            require_build - Requires either 'npm run build' or 'yarn run build' or custom run build command  to be run. Default is false. If value is not provided, it is assumed to be 'true'.

Platform: php
Versions: 5.6.40
          7.0.33
[...]

PR #362, opened 2020-06-08
Update oryx version and definition file for codespaces-linux

edgonmsft:


edgonmsft:
@Chuxel We have one more update to a newer oryx image that fixes PHP on the path. Just finished testing.

Thanks!

dmgardiner25:
@edgonmsft Do we want to remove PHP from the `EXTRA_PATHS` variable in the image since it is included in the new Oryx image?

edgonmsft:
Yep, done.

Chuxel:
Doing a local sanity test real quick.  Is this something you want released ASAP?

edgonmsft:
@Chuxel Thanks a lot! 

Yes if possible.

Chuxel:
@edgonmsft  I'm seeing the following warning when building Git from source.  Is this of any concern?
````
/usr/bin/ld: warning: libssl.so.1.0.2, needed by /usr/lib/gcc/x86_64-linux-gnu/6/../../../x86_64-linux-gnu/libcurl.so, may conflict with libssl.so.1.1
/usr/bin/ld: warning: libcrypto.so.1.0.2, needed by /usr/lib/gcc/x86_64-linux-gnu/6/../../../x86_64-linux-gnu/libcurl.so, may conflict with libcrypto.so.1.1
```
Not sure if that is new or not.  Otherwise LGTM.

edgonmsft:
I saw them too but not sure. I tested git inside the container and seems to be working fine.

Chuxel:
Once CI runs successfully, I can cut a 0.122 release so the update for this definition goes out.  Looks like this should also fix https://github.com/microsoft/vscode-dev-containers/issues/322

PR #378, opened 2020-06-20
Added python & device simulator express dev container

cmendible:
Added python & [device simulator express](https://marketplace.visualstudio.com/items?itemName=ms-python.devicesimulatorexpress#user-content-adafruit-circuit-playground-express-cpx-simulator) dev container


Chuxel:
Oh - one more thing, this is missing a `README.md`. Could you add one?

cmendible:
@Chuxel thanks for the tips. Just pushed the request changes.

containers/python-3-device-simulator-express/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+#    && rm -rf /tmp/pip-tmp
+
+# [Optional] Uncomment this section to install additional packages.
+#
+# RUN apt-get update \
+#     && export DEBIAN_FRONTEND=noninteractive \
+#    && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# Create venv for device simulator
+RUN python3 -m venv /root/.vscode-server-insiders/extensions/ms-python.devicesimulatorexpress/venv
```
Chuxel, line 23: You'll also need to do this for the non-root user (`vscode`) and `.vscode-server` for VS Code stable.  If you only think one venv is needed for all of them, you could create it in one spot and then give the `vscode` user access to it and add symlinks.

containers/python-3-device-simulator-express/README.md
```diff
@@ -0,0 +1,140 @@
...
+3. To use the Dockerfile for this definition (*rather than the pre-built image*):
+   1. Clone this repository.
+   2. Copy the contents of `containers/python-3/.devcontainer` to the root of your project folder.
+   3. Start VS Code and open your project folder.
+
+4. After following step 2 or 3, the contents of the `.devcontainer` folder in your project can be adapted to meet your needs.
+
+5. Finally, press <kbd>F1</kbd> and run **Remote-Containers: Reopen Folder in Container** to start using the definition.
+
+## Testing the definition
```
Chuxel, line 124: Can you update this to cover how to test `circuit.py`?

containers/python-3-device-simulator-express/.devcontainer/Dockerfile
cmendible: ok let me check what I can do here.

cmendible: @Chuxel I have to take a closer look cause I noticed the venv folder also depends on the version vs code  in use: `.vscode-server-insiders`

containers/python-3-device-simulator-express/README.md
cmendible: Added in the **Testing the definition** section

containers/python-3-device-simulator-express/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/opt/vscode/extensions/ms-python.devicesimulatorexpress/venv/bin/python"
```
Chuxel, line 2: Not sure if this is required, but FYI, this won't be picked up when added via VS Code, so if this is needed to run, this should be moved as the python.pythonPath in the `settings` section of `devcontainer.json`.

containers/python-3-device-simulator-express/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+#    && rm -rf /tmp/pip-tmp
+
+# [Optional] Uncomment this section to install additional packages.
+#
+# RUN apt-get update \
+#     && export DEBIAN_FRONTEND=noninteractive \
+#    && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# Create venv for device simulator
+RUN python3 -m venv /opt/vscode/extensions/ms-python.devicesimulatorexpress/venv
```
Chuxel, line 23: After running this you'll probably want to make sure that the non-root user can use it w/o `sudo`.  e.g.

`chown -R vscode:root /opt/vscode/extensions/ms-python.devicesimulatorexpress`

Chuxel:
LGTM!

Issue #368, opened 2020-06-11
Update Ansible devcontainer to support Collections


InTheCloudDan:
I tried to use: https://github.com/microsoft/vscode-dev-containers/tree/master/containers/azure-ansible/.devcontainer as basis for Ansible Collection development. An immediate issue I ran into is that the mount is not to the default COLLECTIONS_PATH. It could either mount a volume to that location or set an environment variable for it.

I'm just learning my way around these containers or I'd offer a PR for the functionality.

Chuxel:
> An immediate issue I ran into is that the mount is not to the default COLLECTIONS_PATH. It could either mount a volume to that location or set an environment variable for it.

@cmendible, should we make a tweak here?

cmendible:
@InTheCloudDan check if https://github.com/microsoft/vscode-dev-containers/pull/379 works for you. This will mount the folder **.ansible/collections** from your local profile into the container.

Chuxel:
Closing since #379 should have this covered. Feel free to reopen if that is not the case.  Thanks much!

PR #379, opened 2020-06-20
Added default mount for Ansible Collections

cmendible:
Added default mount for Ansible Collections. Solving #368

### DIFF
--- a/containers/azure-ansible/.devcontainer/devcontainer.json
+++ b/containers/azure-ansible/.devcontainer/devcontainer.json
@@ -1,15 +1,15 @@
 {
 	"name": "Azure Ansible",
 	"dockerFile": "Dockerfile",
-	"mounts": [ 
+	"mounts": [
+		// [Optional] Anisble Collections: Uncomment if you want to mount your local .ansible/collections folder.
+		// "source=${localEnv:HOME}${localEnv:USERPROFILE}/.ansible/collections,target=/root/.ansible/collections,type=bind,consistency=cached",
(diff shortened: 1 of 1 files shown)


Chuxel:
The only downside here is that the container does not start if the path that is mounted is not present. One way to resolve that would be to have the mount commented out.  Another could be to have a couple of scripts that create the directory if it doesn't exist that are referenced from `initializeCommand`.  (You'd could name one `create-ansible-dir` with the execute bit set for linix/macOS and another `create-ansible-dir.cmd` for windows.)

cmendible:
> The only downside here is that the container does not start if the path that is mounted is not present. One way to resolve that would be to have the mount commented out. Another could be to have a couple of scripts that create the directory if it doesn't exist that are referenced from `initializeCommand`. (You'd could name one `create-ansible-dir` with the execute bit set for linix/macOS and another `create-ansible-dir.cmd` for windows.)

I swear I tested that behavior on my windows box and the local folder does get created. I'll try again and come back to you with the results.

cmendible:
> The only downside here is that the container does not start if the path that is mounted is not present. One way to resolve that would be to have the mount commented out. Another could be to have a couple of scripts that create the directory if it doesn't exist that are referenced from `initializeCommand`. (You'd could name one `create-ansible-dir` with the execute bit set for linix/macOS and another `create-ansible-dir.cmd` for windows.)

I commented out the mount as optional.

Chuxel:
LGTM!

PR #381, opened 2020-06-23
Tensorflow devcontainer

bhack:
Add Tensorflow devcontainer

### DIFF
--- a/repository-containers/github.com/tensorflow/tensorflow/.devcontainer/devcontainer.json
+++ b/repository-containers/github.com/tensorflow/tensorflow/.devcontainer/devcontainer.json
@@ -21,7 +21,13 @@
 		"C_Cpp.clang_format_style": "{BasedOnStyle: Google}",
 		"C_Cpp.default.includePath": [
 			"${workspaceFolder}/**",
-        ],
+        	],
+		"files.watcherExclude": {
+			"**/bazel-*/**": true
+		},
+		"search.exclude": {
+			"**/bazel-*/**": true
+		}
 	},
 
 	// Add the IDs of extensions you want installed when the container is created.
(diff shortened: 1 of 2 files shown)


Chuxel:
@bhack Let me know when you think this is to the point it should be reviewed. I assume you're still making edits like you had for the tenserflow_addons PR.

bhack:
No this is totally independent cause it is not waiting for a upstream image. We are wrapping on the official one. So if you want you can make a pass. I suppose that pylint is enabled by default right?

repository-containers/github.com/tensorflow/tensorflow/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,56 @@
...
+      ],
+
+	// Uncomment when using a ptrace-based debugger like C++, Go, and Rust
+	"runArgs": [ 
+		"--cap-add=SYS_PTRACE",
+		// Uncomment this to enable Nvidia support
+		//"--runtime=nvidia",
+		"--security-opt",
+		"seccomp=unconfined" ],
+	"remoteUser": "1000"
```
Chuxel, line 55: I'd recommend the name of the user in the container here (which I assume is vscode).  The UID can be updated change on Linux if your local OS UID is different.

bhack:
@Chuxel If you can try it on GPU we could merge.

Chuxel:
Unfortunately, I don't have a Linux machine with a NVIDIA GPU in it. (My Linux laptop has an Intel GPU.) My Azure VMs aren't using GPU SKUs at the moment either unfortunately.

bhack:
Ok let's merge we will see if somebody with GPU will open a ticket.

PR #388, opened 2020-07-01
Update devcontainer.json

MOrlassino:
Don't watch the rust ./target directory created/used by cargo. Stops cargo locked/missing file warnings.
Added a useful remote extension that allows one to the monitor resource utilization of the container host.


Chuxel:
LGTM!

PR #389, opened 2020-07-01
Create container configuration for Hugo

AarynSmith:
This creates a container configuration for Hugo static site generator

### DIFF
--- a/containers/hugo/.devcontainer/Dockerfile
+++ b/containers/hugo/.devcontainer/Dockerfile
@@ -17,22 +17,8 @@ RUN apk add --update --no-cache ca-certificates openssl git curl && \
     go get github.com/yaegashi/muslstack && \
     muslstack -s 0x800000 /usr/bin/hugo
 
-FROM alpine:edge
-# This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
-# property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
-# will be updated to match your local UID/GID (when using the dockerFile property).
-# See https://aka.ms/vscode-remote/containers/non-root-user for details.
-ARG USERNAME=vscode
-ARG USER_UID=1000
-ARG USER_GID=$USER_UID
-
+FROM mcr.microsoft.com/vscode/devcontainers/base:alpine
 COPY --from=0 /usr/bin/hugo /usr/bin
-RUN apk add --update --no-cache libc6-compat libstdc++ sudo git openssh-client &&\
-    rm -rf /var/cache/apk/*
-RUN addgroup -g ${USER_GID} ${USERNAME} && \
-    adduser -s /bin/ash -u ${USER_UID} -D -G ${USERNAME} ${USERNAME} && \
-    echo ${USERNAME} ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/${USERNAME}&& \
-    chmod 0440 /etc/sudoers.d/${USERNAME}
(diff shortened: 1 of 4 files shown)


containers/hugo/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Summary
+
+*Develop static sites with Hugo, includes everything you need to get up and running.*
+
+|          Metadata           |                    Value                     |
+| --------------------------- | -------------------------------------------- |
+| *Contributors*              | [Aaryn Smith](https://gitlab.com/aarynsmith) |
+| *Definition type*           | Dockerfile                                   |
+| *Works in Codespaces*       | Yes                                          |
```
Chuxel, line 11: I don't think VS Codespaces (or Codespaces in GitHub) works with Alpine based containers at the moment. Were you able to verify this works?

containers/hugo/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,34 @@
...
+			// Rebuild the container if it already exists to update.
+			"VERSION": "latest",
+		}
+	},
+	// Set *default* container specific settings.json values on container create.
+	"settings": {
+		"terminal.integrated.shell.linux": "/bin/ash"
+	},
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [],
```
Chuxel, line 21: Would it make sense to install an extension for toml support for config.toml?

containers/hugo/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+    export VERSION=$(curl -s https://api.github.com/repos/gohugoio/hugo/releases/latest | grep "tag_name" | awk '{print substr($2, 3, length($2)-4)}') ;;\
+    esac && \
+    echo ${VERSION} && \
+    wget -O ${VERSION}.tar.gz https://github.com/gohugoio/hugo/releases/download/v${VERSION}/${VARIANT}_${VERSION}_Linux-64bit.tar.gz && \
+    tar xf ${VERSION}.tar.gz && \
+    mv hugo* /usr/bin/hugo && \
+    go get github.com/yaegashi/muslstack && \
+    muslstack -s 0x800000 /usr/bin/hugo
+
+FROM alpine:edge
```
Chuxel, line 20: Just as a FYI, there's also a base image you could use here that has things like git, even zsh pre-installed.  Just reference `mcr.microsoft.com/vscode/devcontainers/base:alpine`. It's built out of the `alpine` container folder in the vscode-dev-containers repo and mainly runs [this script](https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/common-alpine.sh). 

Not a requirement to use it, just a convenience if it applies.

containers/hugo/README.md
AarynSmith: You are correct, I thought I had worked when I tested it, but I guess it didn't. Updating now.

containers/hugo/.devcontainer/devcontainer.json
AarynSmith: Makes sense. I've added Better-Toml. Updating Now.

containers/hugo/.devcontainer/Dockerfile
AarynSmith: My reference for the Hugo setup said that alpine:edge was required, but after testing the base image appears to work.

Chuxel:
LGTM!

PR #390, opened 2020-07-02
Configure VSCode (or Insiders, if installed) as the default editor for git commits

bratsche:
https://github.com/microsoft/vscs-planning/issues/128


edgonmsft:
LGTM - @Chuxel any comments?

Also update the definition-manifest.json file to bump up the version.

Chuxel:
LGTM!  Is a release needed?

Issue #254, opened 2020-03-09
Bash IDE


KROSF:
provide a full ide for bash with [mads-hartmann.bash-ide-vscode](https://marketplace.visualstudio.com/items?itemName=mads-hartmann.bash-ide-vscode)

Chuxel:
Similar to #253, happy to take a look at a PR if you're so-inclined to create one!

Chuxel:
Resolved with #391

PR #391, opened 2020-07-02
Add devcontainer for Bash

AarynSmith:
Adds a configuration for Bash development for #254. Included Bash IDE and Bash Debug extensions.

### DIFF
--- a/containers/bash/.devcontainer/Dockerfile
+++ b/containers/bash/.devcontainer/Dockerfile
@@ -1,0 +1,22 @@
+FROM alpine:latest
+# This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
+# property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
+# will be updated to match your local UID/GID (when using the dockerFile property).
+# See https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+RUN apk add --no-cache\
+    bash \
+    curl \
+    git \
+    openssh-client \
+    sudo \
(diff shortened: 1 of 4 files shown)


containers/bash/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Summary
+
+*Develop scripts with Bash, includes [Bash IDE](https://marketplace.visualstudio.com/items?itemName=mads-hartmann.bash-ide-vscode), and [Bash Debug](https://github.com/rogalmic/vscode-bash-debug).*
+
+|          Metadata           |                    Value                     |
+| --------------------------- | -------------------------------------------- |
+| *Contributors*              | [Aaryn Smith](https://gitlab.com/aarynsmith) |
+| *Definition type*           | Dockerfile                                   |
+| *Works in Codespaces*       | Yes                                          |
```
Chuxel, line 11: Similar comment on Codespaces here as Hugo

containers/bash/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM alpine:latest
```
Chuxel, line 1: Also same as in the other case, you could use the pre-built `mcr.microsoft.com/vscode/devcontainers/base:alpine` image if you want. I think all the dependencies in the Dockerfile are covered, so you'd be set.

containers/bash/README.md
AarynSmith: Updated

containers/bash/.devcontainer/Dockerfile
AarynSmith: Updated.

Chuxel:
LGTM!

Issue #304, opened 2020-04-28
Document best practice for multi-stage dockerfiles for dev, build, and prod


Chuxel:
The containers in this repository do not make a strong stance on how dev containers should be used for container based applications because we want to:

- Keep them easy to understand for people new to Docker
- Highlight that they can be used with non-container deployed applications as well

However, we can document how the dev containers can be used in a multi-stage dockerfile (with the `target` property in `devcontainer.json`) as a "builder" for creating the production container image as well.

For example:

devcontainer.json snippet:
```json
{
    "build": {
        "dockerfile": "Dockerfile",
        "target": "development"
    }
}
```

Dockerfile:
```Dockerfile
FROM mcr.microsoft.com/vscode/devcontainers/typescript-node:12 AS development

# Build steps go here
FROM development as builder
WORKDIR /app
COPY src/ *.json ./
RUN yarn install \
    && yarn compile \
    #  Just install prod dependencies
    && yarn install --prod

# Actual production environment setup goes here
FROM node:12-slim AS production
WORKDIR /app
COPY --from=builder /app/out/ ./out/
COPY --from=builder /app/node_modules/ ./node_modules/
COPY --from=builder /app/package.json .
EXPOSE 3000
ENTRYPOINT [ "/bin/bash", "-c" ]
CMD [ "npm start" ]
```

This allows the dev container image to be used for development inside the container and in building the application for production, but a "slim" image to be used for production with it's contents copied out of the previous "builder" stage. This creates the smallest possible prod image. 

Furthermore, once this smaller image is deployed, the "attach" workflow can be used if there is something that only appears with this config.

PavelSosin-320:
@Chuxel 
How-to: correctly deploy dev container to managed Swarm cluster(*)
0. Enable Swarm mode in Docker Desktop and init Swarm - docker swarm init
1. Deploy registry;2 image as localhost:5000 from Dockerhub as a service - local images are not supported
2. Clone VSCodeDevContainerbase to the local registry
3 Deploy the cloned image to the swarm as a service:
```
docker service create --detach=false --name=devcontainerVol --network=host --tty --no-healthcheck  \
--mount source=git,target=/var/git localhost:5000/devcontainerVol
```
4. Find the task running locally -  `docker service ps devcontainerVol `
--- memorize task id === container id
5. Test git clone docker exec 9621f482ce96 git clone https://github.com/microsoft/vscode-dev-containers.git`
[...]

PR #395, opened 2020-07-06
Add Devcontainer for Vue.js

AarynSmith:
This adds a config for Vue.js using a node image. 

@Chuxel I debated running the same image from the others and rolling in node, but not sure thats any more efficent, or do you think I should copy the javascript-node-14 config and add in the Vue install?

### DIFF
--- a/containers/vue/.devcontainer/Dockerfile
+++ b/containers/vue/.devcontainer/Dockerfile
@@ -1,16 +1,4 @@
-FROM node:lts-alpine
-
-# This Dockerfile adds a non-root 'vscode' user with sudo access. However, for Linux,
-# this user's GID/UID must match your local user UID/GID to avoid permission issues
-# with bind mounts. Update USER_UID / USER_GID if yours is not 1000. See
-# https://aka.ms/vscode-remote/containers/non-root-user for details.
-ARG USERNAME=node
-ARG USER_UID=1000
-ARG USER_GID=$USER_UID
-
-RUN apk add --no-cache sudo git openssh-client \
-	&& echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME\
-	&& chmod 0440 /etc/sudoers.d/$USERNAME
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:14
 
 RUN npm install -g http-server @vue/cli @vue/cli-service-global
 WORKDIR /app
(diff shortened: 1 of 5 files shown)


Chuxel:
Or you can just reference `mcr.microsoft.com/vscode/devcontainers/javascript-node:14`.  In many ways that would probably be better since the rest of the Node related definitions are Debian based instead of Alpine (and not all extensions will work in an Alpine image due to native dependencies).  Is there a specific reason to use the Alpine image?

AarynSmith:
looks like the javascript-node:14 image works fine. I was trying to minimize sizes and stick to the Alpine image, but if it will mean better compatibility, I have no problem switching to the Debian based. This should make it compatible with Codespaces as well correct?

containers/vue/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:14
+
+RUN npm install -g http-server @vue/cli @vue/cli-service-global
+WORKDIR /app
+COPY . .
```
Chuxel, line 5: This will also copy all local source into the container which is not required for a dev container.  If you want to use the same Dockerfile for deployment and development, you can set up a multi-stage Dockerfile and use the `target` property. (Example is in https://github.com/microsoft/vscode-dev-containers/issues/304)

containers/vue/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
+EXPOSE 8080
+CMD [ "/usr/local/bin/npm run serve" ]
```
Chuxel, line 8: Similarly, this won't be fired unless you set `"overrideCommand": false` in devcontainer.json.

containers/vue/.devcontainer/Dockerfile
AarynSmith: I've removed the copy . . and the npm run serve from the dockerfile.

AarynSmith:
> One thing I just noticed is that this is missing a `.npmignore` file. That's used to exclude folders/files when it is added to a project through VS Code.
> 
> e.g.
> 
> ```
> README.md
> test-project
> definition-manifest.json
> .vscode
> .npmignore
> ```

Added.

Chuxel:
LGTM!  Thanks!

AarynSmith:
Awesome thanks!

PR #396, opened 2020-07-09
Add commented option of embedding python project requirements into image

fonzcastellanos:
requirements

The expected commented lines in the RUN command for installing requirements were missing, so I've added them.


Chuxel:
Whoops!  That's a good catch since the earlier comment is still there. What I'd suggest that we should instead add a comment on it after the main `RUN` at the very end of `base.Dockerfile` that mirrors `Dockerfile`.

```Dockerfile
# [Optional] If your requirements rarely change, uncomment this section to add them to the image.
#
# COPY requirements.txt /tmp/pip-tmp/
# RUN pip3 --disable-pip-version-check --no-cache-dir install -r /tmp/pip-tmp/requirements.txt \
#    && rm -rf /tmp/pip-tmp
```

That way if requirements.txt updates, it will only rebuild this one part rather than the entire previous RUN section. I'll make that update post-merge.

PR #399, opened 2020-07-11
Consolidate version specific javascript-node/typescript-node definitions into two

Chuxel:
This is the next update in https://github.com/microsoft/vscode-dev-containers/issues/294

Moves from a total of 6 definitions with versions for Node.js to 2 - `javascript-node` and `typescript-node` that use a build arg instead. Still generates all 6 images.


chrmarti:
LGTM! (Can't "approve" anymore since it is already merged.)

PR #400, opened 2020-07-11
Create a container definition for JavaScript exercism.io (Node 12)

CyberLight:
This change creates a container definition for Javascript [exercism.io](https://exercism.io/) task solving environment based on Node.js 12

### DIFF
--- a/containers/javascript-exercism-node-12/.devcontainer/Dockerfile
+++ b/containers/javascript-exercism-node-12/.devcontainer/Dockerfile
@@ -1,0 +1,17 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+# To fully customize the contents of this image, use the following Dockerfile instead:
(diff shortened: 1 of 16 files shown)


bamurtaugh:
Thanks for tagging me. I agree with raising awareness from the Exercism docs and/or extension. I did notice the extension appears not to be officially associated with Exercism, though.

I also wonder where it makes the most sense for this container definition to live- it seems like individuals who would like to use it might get the most benefit from it if it could be merged directly as part of an Exercism repo (maybe going one step beyond just referencing it there). Is there a location from which users pull Exercism exercises? It feels like it'd be natural to include a dev container definition there, so that they could pull source code + a dev container all together.

I also agree with Chuck's comments about making the definition as general as possible.

CyberLight:
@Chuxel and @bamurtaugh you are right, earlier I thought that will be nice if this container would be in the VSCode container repository, because it already contains all the necessary tools in order not to look somewhere in the documentation or elsewhere, but you convinced me that it is easier for everyone to create such a container by himself using VSCode Remote Containers.

Alas, I could not write my nickname in history :)

I think that this MR can be closed, but I got experience while creating this container, this is an invaluable experience. Thank you for your time, you are the best!

Chuxel:
Sounds good!  Thanks for the work on it!

bamurtaugh:
Thank you for your time and work in creating this container definition! Glad it could be a good experience :)

PR #401, opened 2020-07-11
Refresh Ruby definitions with latest patterns

Chuxel:
This PR updates the Ruby, Ruby on Rails, and Ruby & Sinatra definitions as follows:

1. Allows the selection of the version of Ruby through a build argument in `devcontainer.json`
2. Uses the common dependency script and adds in arguments for options
3. Uses the Node script to allow the selection of a version of Node.js (or disable its installation)

//cc: @a-oz since this updates two of your definitions.


Chuxel:
I'm merging this in so I can pick it up in a comments reduction pass I am doing.  Please let me know if you see anything you think should be changed.

PR #403, opened 2020-07-12
Add cmake

aszenz:
Cmake is a required dependency for building rust, see https://github.com/rust-lang/rust/blob/dbf3ae7c3beb5b493375bf76152e490b8cc81d1c/README.md#building-on-a-unix-like-system


Chuxel:
Good catch. Thanks!

Issue #398, opened 2020-07-10
.NET 3.1.302 SDK not installed on codespace-linux


timheuer:
- Name of Dev Container Definition with Issue: codespace-linux

Steps to Reproduce:

1. Create a Codespace w/Linux using a Blazor Wasm app repo
2. Attempt to build
EXPECT: Success
ACTUAL: Fail as Blazor Wasm build components in 3.1.302+ SDK for .NET Core which is not present as of the date of this report

Chuxel:
//cc: @edgonmsft

timheuer:
@chuxel - looks like it actually needs to be here: https://github.com/microsoft/Oryx/blob/c2314bce463ceb6c82be1291c1f6f78e5fcab14f/platforms/dotnet/versionsToBuild.txt#L28 as a new line...

Chuxel:
This should be updated in the current `dev` image `mcr.microsoft.com/vscode/devcontainers/universal:dev`

Chuxel:
Closing since this should be resolved and will be deployed into Codespaces soon.

PR #404, opened 2020-07-13
Update oryx image and symlink to newest version.

edgonmsft:
Update to latest oryx, that includes .Net Core 3.1.302 as per https://github.com/microsoft/vscode-dev-containers/issues/398

Also updates the corresponding symlink file.


edgonmsft:
This one includes 3.1.301 maybe we want to wait for them to release 302?

Chuxel:
Running build to verify. Does this need to be released urgently?

edgonmsft:
Probably not since I'm still waiting for oryx to release one with that requested version, we are still one behind

edgonmsft:
Lets wait and let me check with them if they are going to add it soon.

PR #405, opened 2020-07-14
Update Vue definition to match consolidated javascript-node definition, fix non-root scenario

Chuxel:
I just merged in #399 which consolidates the node 10, 12, 14 definitions into one with a build arg. This updates Vue to use the same pattern.

I also noticed that the vue CLI was installed globally as root, which can cause problems with caching, so I prefixed that with `sudo -u node`.

LMK if it looks good @AarynSmith

### DIFF
--- a/containers/python-3/.devcontainer/Dockerfile
+++ b/containers/python-3/.devcontainer/Dockerfile
@@ -7,13 +7,13 @@
 ARG VARIANT=3
 FROM mcr.microsoft.com/vscode/devcontainers/python:${VARIANT}
 
-# [Optional] If your requirements rarely change, uncomment this section to add them to the image.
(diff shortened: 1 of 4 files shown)


AarynSmith:
LGTM. So a single Javascript container with a build argument for the version going forward?

Chuxel:
@AarynSmith Yep, under the hood, unique 10, 12, and 14 images are built and used, but all you have to do now to switch versions is update `devcontainer.json`.  This is how the Python, C#, F#, and soon Swift definitions are setup as well. 

That way differences in definitions can focus on things that are truly unique rather than a single version update.

AarynSmith:
Good deal. Thanks.

PR #411, opened 2020-07-16
Update README.md

ATGardner:
fixed step 4 - RUN echo commands should use " instead of ' for the argument to be replaced


Chuxel:
Good catch!  Thanks!!

Issue #409, opened 2020-07-15
Simplify dev containers that do more than install packages


Chuxel:
Currently the development container definitions in this repository use a single `RUN` to ensure any cleanup that happens is applied to the same layer. This ensures the image size is as small as possible since doing this across RUNs means there is a layer with the contents in the image even after it is removed in a subsequent layer. This is particularly important in pre-built images and published images.

However, this style can be a bit daunting when you first see it. Some parts of these `RUN` statements can be safely separated which can reduce build times if any build args are referenced in the statement. In some cases, it may be better to use a separate shell script when the layer is particularly large. (The downside here being that you have to copy something else when manually copying the contents to your project.) Other older definitions are also not using the common setup scripts which exacerbates this problem.

Finally, Codespaces does not currently have the "Add Development Container Definitions..." flow that Remote - Containers has. As a result, more people are currently looking at `base.Dockerfile` in definitions that have pre-built images. While they're generally intended for "advanced" scenarios, they can use some refinement as well.

Add explanation for those contributing or using definitions:

- [x] CONTRIBUTING
- [x] README

The following definitions could use some layer refinement:

- [x] Python 3
- [x] C++
- [x] Go (related to #375)
- [x] JavaScript / Typescript & Node
- [x] PHP
- [x] Rust #473
- [x] Ruby
- [x] C#, F#
- [x] Docker from Docker / Compose #474
- [x] Kubernetes Helm #474
- [x] Azure Terraform
- [x] Dapr definitions
- [x] Powershell
- [x] Codespaces
- [x] Swift

The following can simplified with the common script:

- [x] Go (related to #375)
- [x] Rust  #473
- [x] Java
- [x] PHP
- [x] Dart
- [x] Jekyll
- [x] Perl  #473
- [x] Bazel  #473
- [x] R  #473
- [x] Azure Ansible #472 
- [x] Azure Blockchain #476
- [x] Azure Functions .NET 2.1
- [x] Azure Functions PowerShell
- [ ] Azure Machine Learning
- [x] Azure Terraform
- [x] Swift

Missing comments about adding your own `RUN` statement:

- [x] Debian base.Dockerfile
- [x] Ubuntu base.Dockerfile
- [x] Go
- [x] Deno
- [x] Bash
- [x] Hugo
- [ ] Elm

PR #412, opened 2020-07-17
Fix error with hugo container

AarynSmith:
Apparently there was a bug with running hugo-extended in alpine containers. I thought it was working, but it wasn't. 

This PR switches from alpine to the base versions to resolve it.


AarynSmith:
Also added the "Add your own run statement" comment for #409.

Chuxel:
LGTM! Thanks!

PR #415, opened 2020-07-18
Python 3 Pipenv dev container

KYDronePilot:
Added a Python 3 Pipenv dev container.

Dockerfile uses the base images from the `python-3` dev container.

`devcontainer.json` is configured by default to install the same Python tools as the `python-3` dev container using Pipenv. It saves them as dev packages.

Had to use a trick of creating a symlink to the Pipenv virtualenv with a glob pattern to get a consistent path to the environment. This symlink path is then used in the `devcontainer.json` config for the Python/tools executable paths.


Chuxel:
Thanks for submitting the PR! Would it make more sense to include this in the python-3 container itself?  How large is the install?

KYDronePilot:
As far as installing Pipenv itself as one of the Python utils, it would add approximately `51.1` MB to the `python-3` container.

We would need to have some way of handling the paths to the Python executable and tools, depending on whether the user decides to use the default tools or Pipenv.

I think this could be accomplished by using static symlinked paths for the Python executable and tools `bin/` in the `devcontainer.json` config such as:

- `/usr/local/python` - Python executable
- `/usr/local/py-static-utils-bin` - Python tools bin

Then, using a Docker arg `"PIPENV": "true"`, there could be a simple bash script in the Dockerfile to determine whether or not the user wants to use Pipenv. If they don't, the directories are symlinked to the paths they currently use for the `python-3` dev container:

- `/usr/local/python` -> `/usr/local/bin/python`
- `/usr/local/py-static-utils-bin/` -> `/usr/local/py-utils/bin/`

If they do, they are symlinked to the Pipenv environment's Python executable and bin:

- `/usr/local/python` -> `~/.local/share/virtualenvs/*-*/bin/python`
- `/usr/local/py-static-utils-bin/` -> `~/.local/share/virtualenvs/*-*/bin/`

The only downside is that the image will have to be built from the Dockerfile and cannot be referenced directly from the registry. Although this should cause no speed hit, since it will just be symlinking some directories.

Does the Pipenv size and process for handling everything sound like it will work? If so, we can close this PR and I can start working on that implementation instead.

Chuxel:
@brettcannon What are your thoughts [here](https://github.com/microsoft/vscode-dev-containers/pull/415#issuecomment-661588323)? Would a separate definition be preferable?

brettcannon:
@Chuxel I think it depends on how much harmony you want in the container definitions. Tools like poetry and pipenv use virtual environments in the end, so they are similar that way. But then they use their own lock file format which you can only install by using the tool (although I may end up working with someone to try and solve that problem). So if you don't want people to uncomment stuff then it will require a different container definition. BTW, I would assume someone will bring up Poetry at some point, so whatever solution is chosen expect one more tool to have to do it for.

@KYDronePilot you could use [`PIPENV_VENV_IN_PROJECT`](https://pipenv.pypa.io/en/latest/advanced/#pipenv.environments.PIPENV_VENV_IN_PROJECT) to get a consistent virtual environment location.

Chuxel:
@brettcannon Got it, this is specifcally about `"python.pipenvPath": "/usr/local/py-utils/bin/pipenv"` and the related python path?

@KYDronePilot I may be missing it, but why is this part of the postCreateCommand needed?

```
pipenv install --dev --pre pylint flake8 autopep8 black yapf mypy pydocstyle pycodestyle bandit
```

These are installed and in the path already via pipx, so they should be available.  Effectively, wouldn't this put two copies on disk since they're not in the global environment?

Using "~" means that you'd technically have to install it twice... once for the root user and one for the VS Code user. That's why utilities aren't installed under a user's folder. So we'd want to get these in a global location anyway or setup symlinks.  Can't the related symlinks be in the Dockerfile?

KYDronePilot:
@brettcannon I had thought about doing that, but was worried that in an unlikely case someone might already have a virtual environment from the host machine in that path. On the upside though, rebuilding the container wouldn't take as long, since the environment would persist across container rebuilds.

@Chuxel Yes, those are already installed, however, because they are separate from the Pipenv environment, the linters complain about imports because they can't see installed packages. While these errors could likely be fixed with some tweaking of the VSCode settings for the linters, it would prevent the tools from being tracked as Pipenv dev packages. This (IMO) is one of the more useful features of Pipenv, since it allows the versions of dev packages to be precisely managed and allows for interoperability with people who don't or can't use dev containers.

Regarding settings related to `~`, things would only have to be installed twice if done at the Docker image build time. Given that the Pipenv environment (with or without the `PIPENV_VENV_IN_PROJECT` env var set) is dependent on the workspace folder (which hasn't yet been attached at build time), the Pipenv environment setup would need to happen using the `postCreateCommand`. Then, regardless of the user chosen, `~` would refer to the correct directory.

Using `PIPENV_VENV_IN_PROJECT` would eliminate having to use `~` entirely, however, in favor of VSCode config variables.

Chuxel:
> Using PIPENV_VENV_IN_PROJECT would eliminate having to use ~ entirely, however, in favor of VSCode config variables.

I would prefer this so that you can to move these contents into the Dockerfile, and to leave the postCreateCommand for users of the definition to use for their own project needs. This also avoids everything happening again every time a developer rebuilds since they'd be able to take advantage of the Docker build cache.

> @Chuxel Yes, those are already installed, however, because they are separate from the Pipenv environment, the linters complain about imports because they can't see installed packages. While these errors could likely be fixed with some tweaking of the VSCode settings for the linters, it would prevent the tools from being tracked as Pipenv dev packages. This (IMO) is one of the more useful features of Pipenv, since it allows the versions of dev packages to be precisely managed and allows for interoperability with people who don't or can't use dev containers.

Are you thinking about these as project dependencies? Generally, if these are intended to be tracked as project assets, they shouldn't be in the definition. There's a similar analog in the Node world with globally installed packages vs locally installed. We only add global installs into the definitions.  That said, if that's not the case, it could be added to the Dockerfile - or you could consider using [base.Dockerfile](https://github.com/microsoft/vscode-dev-containers/blob/master/containers/python-3/.devcontainer/base.Dockerfile) as a starting point. That way you can remove the section with the PIPX tool installs entirely.

KYDronePilot:
> Are you thinking about these as project dependencies? Generally, if these are intended to be tracked as project assets, they shouldn't be in the definition. There's a similar analog in the Node world with globally installed packages vs locally installed. We only add global installs into the definitions. That said, if that's not the case, it could be added to the Dockerfile - or you could consider using [base.Dockerfile](https://github.com/microsoft/vscode-dev-containers/blob/master/containers/python-3/.devcontainer/base.Dockerfile) as a starting point. That way you can remove the section with the PIPX tool installs entirely.

@Chuxel Yes, what I was thinking of was project dependencies. The convenience aspect I had in mind is that having to change the paths for the Python tools was time consuming. But if someone already has to choose which tools to add as local project dev dependencies, then changing the paths isn't that much extra.

> @KYDronePilot I would prefer this so that you can to move these contents into the Dockerfile, and to leave the postCreateCommand for users of the definition to use for their own project needs. This also avoids everything happening again every time a developer rebuilds since they'd be able to take advantage of the Docker build cache.

`pipenv install ...` must be run from within the project directory, which hasn't yet been mounted at image build time. If `PIPENV_VENV_IN_PROJECT` is set though, the environment would live in the mounted project directory on the host system. Then whenever `pipenv install ...` is run using `postCreateCommand`, it should complete quickly, since the previously installed packages are still there.

At this point, I think the most that could be done to add Pipenv support would be adding the following to the existing `python-3` definition:

- The `PIPENV_VENV_IN_PROJECT` env var in `base.Dockerfile`.
- `pipenv` itself as one of the tools to install with PIPX.
- The `python.pipenvPath` setting pointing to where PIPX installs Pipenv.
- A commented out `python.pythonPath` setting pointing to the location of the Python executable if Pipenv is used.
- An additional commented out `postCreateCommand` alongside the existing one, showing how to have Pipenv install dependencies after the container is created:

  ```json
  // "postCreateCommand": "pipenv install --dev",
  ```

Chuxel:
> Then whenever pipenv install ... is run using postCreateCommand, it should complete quickly, since the previously installed packages are still there.

The challenge is that I don't think pipenv will see them - they're in isolated virtual environments so the host python doesn't see them.

One thing that could potentially be done is to exploit path order in devcontainer.json and then add the `bin` folders into the PATH.

e.g. "python.formatting.autopep8Path": null

This means "unset" and should cause it to look in the path (even if you set this in your user settings). 

You could then add the root and vscode user's location into the path using `ENV PATH` in the Dockerfile.

You can then just have the commented out `// "postCreateCommand": "pipenv install --dev",` and mention in the README that people can add things to it.

KYDronePilot:
> The challenge is that I don't think pipenv will see them - they're in isolated virtual environments so the host python doesn't see them.

I'm referring to packages installed specifically using Pipenv, not the default tools installed with PIPX in the existing image. The first time(s) Pipenv is run to setup the environment and install packages chosen by the user (which may happen to be some of the same tools already installed with PIPX), the virtual environment and packages will be located in `/path/to/workspace/.venv` which is mounted to the host machine. In that way, any subsequent `pipenv install` calls when rebuilding the container would complete quickly since the Pipenv virtual environment at `/path/to/workspace/.venv` would still exist.

> One thing that could potentially be done is to exploit path order in devcontainer.json and then add the `bin` folders into the PATH.
>
> e.g. "python.formatting.autopep8Path": null
>
> This means "unset" and should cause it to look in the path (even if you set this in your user settings).
>
> You could then add the root and vscode user's location into the path using `ENV PATH` in the Dockerfile.

This would work if we didn't use `PIPENV_VENV_IN_PROJECT`, since the user-dependent Pipenv virtual environment path of `~/.local/share/virtualenvs/*-*/` could be known at image build time. If we did use `PIPENV_VENV_IN_PROJECT` (which the benefits mentioned above rely on), then the virtualenv path would be workspace-based (`/path/to/workspace/.venv`) which isn't known until the volume is mounted. There would need to be a way of setting the PATH in the container from `devcontainer.json` after mounting the workspace, using something like `${workspaceFolder}/.venv/bin`, while also extending the existing PATH (not sure if this can be done in a clean way).

Chuxel:
> There would need to be a way of setting the PATH in the container from devcontainer.json after mounting the workspace, using something like ${workspaceFolder}/.venv/bin, while also extending the existing PATH (not sure if this can be done in a clean way).

This should do it:

```json
"remoteEnv": {
    "PATH": "${containerWorkspaceFolder}/.venv/bin:${containerEnv:PATH}"
}
```

"remoteEnv" is sets the environment VS Code itself gets after the container is already spun up. See [here](https://code.visualstudio.com/docs/remote/devcontainerjson-reference#_variables-in-devcontainerjson).

KYDronePilot:
Unfortunately, it looks like exploiting path order for the tools won't work. If I set the paths to `null`, the Python extension errors out because it only wants a string. If I set a tool path to just the name of the executable (e.g. `"python.linting.pylintPath": "pylint"`) which is the default value, it then only looks for the tool to be installed as a package in the Python environment (Pipenv or system) and therefore won't check the PATH for it. This is fine when using Pipenv, but not when using the system Python interpreter.

The second issue doesn't seem to effect `python.pythonPath` when set to `python`, so at least the Python interpreter path would work. As for the tools, I don't think there is a way to cleanly make them work with both the system Python interpreter and Pipenv (depending on which the user chooses) without creating a separate definition.

mcsitter:
I would love for this to become a feature! If there is any way i can help, i am happy to do so. It seems as if there are some questions open on how to handle the paths for installed tools ideally?

KYDronePilot:
@mcsitter At the moment, I think the best we can do is to add Pipenv as one of the installed packages in the main `python-3` container template and maybe some commented out settings to give user guidance if they want to use Pipenv. The Python interpreter path should be able to be configured to work with both Pipenv and the standard Python environment in the container.

As far as the Python tool paths are concerned, I don't see any nice way of handling them, other than the user manually editing them in `devcontainer.json` if they want to add Pipenv dev dependencies.

@Chuxel Do you see any other way of handling this? If not, this PR can be closed and a new one opened for changes to the existing `python-3` container template.

Chuxel:
Sorry for the delay - will try to take another look here soon.  Got side tracked.

mcsitter:
@KYDronePilot Having pipenv preinstalled in the container would be a big plus and hopefully does not add too much bulk to it!

Chuxel:
Yeah taking another look at this, I think there's really two possible ways to do this:

1. Include pipenv in python-3 and call it good

2. Move the two separate postCreateCommands into a script that is in the `.devcontainer` folder. This script can then be a bit smarter which of the two listed command should be run.  Detecting if these things are already in place or not.  Since this can be dropped into any project, it's entirely possible that many if not all of these tools are already in the the Pipfile.

Make sense?

mcsitter:
With 2. it seems as if you want to make it as if one is always within the virtualenv without activating it. Why all the hassle?

VSCode can already handle environments well! If you set the python interpreter, it will run code inside it and use the site-pacakges from there for linting etc. as well as automatically activate the virtualenv when opening a terminal.

If we want to support pipenv it should just run "pipenv sync --dev" probably and somewhere other than .vscode/settings.json the correct interpreter is selected.

Doing 1. would probably be a good first step either way, i think.

KYDronePilot:
Apologies for the delayed response.

I agree with @mcsitter that method two would require a lot of extra hassle, since VSCode should ideally know about the paths to the Python tools, presumably through a config file.

While checking out how to implement method one, I noticed that b03e101e97ea79bfe18752ac8f68469d6715e798 adds Pipenv as one of the Python tools, so nothing more should need to be done here. If we want to make any further improvements to Pipenv support, they should be done on the `python-3` container template, as we already discussed, so I'll go ahead and close this PR.

PR #417, opened 2020-07-18
Added script for docker-from-docker in redhat

smankoo:
Added setup script (docker-redhat.sh) for docker-from-docker setups for redhat flavored distros


Chuxel:
Thanks for submitting the PR! Will try to take a look this week!

Chuxel:
Looks great!  Thanks so much for the contribution.  I found one issue in both this and the debian script related to scenarios where the `docker` group already exists, but I'll make that fix for both separately.  This is great otherwise!

PR #424, opened 2020-07-22
containers/go: correct tools installation in modules mode

hyangah:
When installing tools, install them individuallly rather than
with one single `go get` command. Otherwise, the go command
runs MVS for all the listed tools and builds each tool with
unexpected dependencies.

See https://github.com/golang/vscode-go/issues/363#issuecomment-660335139

Also, remove gometalinter - vscode-go no longer uses it.

GO111MODULE=auto is perfectly fine. If `go.mod` exists
in the repo, the go command will automatically switch to the
modules mode. Unsetting is also fine but I will leave it as it is.


Chuxel:
LGTM! Thanks much!

Issue #375, opened 2020-06-17
Simplify Golang definition once gopls is 1.0


lawrencegripper:
Golang has moved toward using `gopls` for nearly all language features. Most of the tooling installed in the current template is no longer needed and adds bloat to the container. 

Would it be alright to push another golang container and have two, one using the go language server and the other one using the existing setup?

With `gopls` the install drops to the following, with a few additional settings. I've also done in these examples is to pin the versions to avoid breaking changes, this feels like the right thing to do to preview breaking changes, for example #173 affecting users, but I understand this would incur more effort to maintain the template. 

```
RUN \
    # --> Delve for debugging
    go get github.com/go-delve/delve/cmd/dlv@v1.4.1 \
    # --> Go language server
    && go get golang.org/x/tools/gopls@v0.4.1 \
    # --> Go symbols and outline for go to symbol support and test support 
    && go get github.com/acroca/go-symbols@v0.1.1 && go get github.com/ramya-rao-a/go-outline@7182a932836a71948db4a81991a494751eccfe77
```
```
"settings": {
		"terminal.integrated.shell.linux": "/bin/bash",
		"go.gopath": "/go",
		"go.useLanguageServer": true,
		"[go]": {
			"editor.snippetSuggestions": "none",
			"editor.formatOnSave": true,
			"editor.codeActionsOnSave": {
				"source.organizeImports": true,
			}
		},
		"gopls": {
			"usePlaceholders": true, // add parameter placeholders when completing a function
			// Experimental settings
			"completeUnimported": true, // autocomplete unimported packages
			"watchFileChanges": true, // watch file changes outside of the editor
			"deepCompletion": true, // enable deep completion
		},
		"files.eol": "\n", // formatting only supports LF line endings		
	},
```

The repository container I pushed in this PR is an example of how much this can be slimmed down. https://github.com/microsoft/vscode-dev-containers/pull/374

Or similar in the [azbrowse devcontainer](https://github.com/lawrencegripper/azbrowse/blob/master/.devcontainer/Dockerfile).

Chuxel:
@lawrencegripper I would prefer not to have two since customers would need to know which one to use in a given case. These show up in VS Code UX, so showing two with the same functionality won't make sense. With Google picking up management of the extension, I'd it also may make sense to have the extension owners weigh in on what is done here at this point. I'll see if I can get a contact.

hyangah:
[...]

PR #426, opened 2020-07-23
Update Go definition to use common script, two layers, simplify a bit

Chuxel:
This PR is part of #409 and related to #375. It does a few things to the Go definition:

1. Switches it to using the common script used in other definitions that installs common utilities and creates the vscode non-root user.
2. Separates out the installation of the Go tools into a separate layer so that any modifications will not require a re-run of the common script while still ensuring each layer cleans up before moving to the next step.
3. Uses xargs to simplify the code that installs all the different Go tools.
4. Adds comments at the end of the Dockerfile about adding your own `RUN` statements for additional installed contents.
5. Adds a build arg that allows you to specify the Go version from devcontainer.json

For #375, we would just need to trim the list in the `GO_TOOLS_WITH_MODULES` arg once we're ready and it's easier for people to opt to do that themselves now.

//cc: @hyangah @lawrencegripper for input

### DIFF
--- a/containers/go/.devcontainer/Dockerfile
+++ b/containers/go/.devcontainer/Dockerfile
@@ -57,8 +57,8 @@ RUN mkdir -p /tmp/gotools \
     && (echo "${GO_TOOLS_WITH_MODULES}" | xargs -n 1 go get -x )2>&1 \
     # gocode-gomod
     && export GO111MODULE=auto \
-    && GOPATH=/tmp/gotools go get -x -d github.com/stamblerre/gocode 2>&1 \
-    && GOPATH=/tmp/gotools go build -o gocode-gomod github.com/stamblerre/gocode \
+    && go get -x -d github.com/stamblerre/gocode 2>&1 \
+    && go build -o gocode-gomod github.com/stamblerre/gocode \
     # golangci-lint
     && curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin 2>&1 \
     # Move Go tools into path and clean up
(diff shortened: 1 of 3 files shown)


containers/go/.devcontainer/Dockerfile
```diff
@@ -1,66 +1,77 @@
...
+    github.com/mgechev/revive \
+    github.com/go-delve/delve/cmd/dlv"
+RUN mkdir -p /tmp/gotools \
+    && cd /tmp/gotools \
+    && export GOPATH=/tmp/gotools \
+    # Go tools w/module support
+    && export GO111MODULE=on \
+    && (echo "${GO_TOOLS_WITH_MODULES}" | xargs -n 1 go get -x )2>&1 \
+    # gocode-gomod
+    && export GO111MODULE=auto \
```
hyangah, line 59: (optional) I think we can utilize GOBIN to build gocode-gomod in modules mode

GOBIN="${GOPATH}" go get -x github.com/stamblerre/gocode 2>&1 \
  && mv "${GOPATH}/gocode" "${GOPATH}/bin/gocode-gomod" 

But it's beyond the scope of this change so I am ok with this.

containers/go/README.md
```diff
@@ -14,7 +14,13 @@
 
 ## Using this definition with an existing folder
 
-This definition does not require any special steps to use. Just follow these steps:
+While the definition itself works unmodified, you can select the version of Go the container uses by updating the `VARIANT` arg in the included `devcontainer.json` (and rebuilding if you've already created the container).
+
+```json
+"args": { "VARIANT": "1.13" }
```
hyangah, line 20: let's suggest 1.14 :-) (soon we will be with 1.15)

containers/go/.devcontainer/Dockerfile
Chuxel: Yeah I wasn't able to get this to work when I tried quickly. Let's separate it out - though I like the idea!

containers/go/README.md
Chuxel: Done!

PR #429, opened 2020-07-25
Alias code-insiders to code for Codespaces image

jkeech:
Fixes https://github.com/MicrosoftDocs/vscodespaces/issues/685

### DIFF
--- a/containers/codespaces-linux/.devcontainer/Dockerfile
+++ b/containers/codespaces-linux/.devcontainer/Dockerfile
@@ -150,6 +150,9 @@ RUN apt-get update -yq\
     && install -o codespace -g codespace -m 755 /tmp/codespace/git-ed.sh /home/codespace/.local/bin/git-ed.sh \
     && sudo -u codespace git config --global core.editor "/home/codespace/.local/bin/git-ed.sh" \
     #
+    # Alias code-insiders to code for easy terminal file editing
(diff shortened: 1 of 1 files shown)


containers/codespaces-linux/.devcontainer/Dockerfile
```diff
@@ -150,6 +150,9 @@ RUN apt-get update -yq\
     && install -o codespace -g codespace -m 755 /tmp/codespace/git-ed.sh /home/codespace/.local/bin/git-ed.sh \
     && sudo -u codespace git config --global core.editor "/home/codespace/.local/bin/git-ed.sh" \
     #
+    # Alias code-insiders to code for easy terminal file editing
```
edgonmsft, line 153: Would it cause problems if a user connects from both at different points in time?

jkeech: I don't think so, since the VS Code remote server sets itself on the PATH as part of spin up, so I'd expect each instance of VS Code's extension host terminals to only have itself on the PATH. But it's not a scenario I've actually tested.

jkeech: I think this should work, so I'm going to go ahead and leave it. If we find out it causes any problems for some edge-case scenarios, we can revisit.

PR #432, opened 2020-07-25
Elixir + Phoenix with PostgreSQL and Node.js

idyll:
This should be the full stack that you need to run Elixir/Phoenix 

It includes Node.js and Postgres as both are required for the Standard Phoenix installation.

### DIFF
--- a/containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
+++ b/containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
@@ -1,0 +1,61 @@
+# Update the VARIANT arg in devcontainer.json to pick a Ruby version: 2, 2.7, 2.6, 2.5
+ARG VARIANT=1.10
+FROM elixir:${VARIANT}
+
+# This Dockerfile adds a non-root user with sudo access. Update the “remoteUser” property in
+# devcontainer.json to use it. More info: https://aka.ms/vscode-remote/containers/non-root-user.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Options for common package install script
+ARG INSTALL_ZSH="true"
+ARG UPGRADE_PACKAGES="true"
+ARG COMMON_SCRIPT_SOURCE="https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/script-library/common-debian.sh"
+ARG COMMON_SCRIPT_SHA="dev-mode"
+
+# Optional Settings for Phoenix
+ARG PHOENIX_VERSION="1.5.4"
+
+# [Optional] Settings for installing Node.js.
+ARG INSTALL_NODE="true"
+ARG NODE_SCRIPT_SOURCE="https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/script-library/node-debian.sh"
+ARG NODE_SCRIPT_SHA="dev-mode"
+ARG NODE_VERSION="lts/*"
+ENV NVM_DIR=/usr/local/share/nvm
+ENV NVM_SYMLINK_CURRENT=true
+ENV PATH=${NVM_DIR}/current/bin:${PATH}
+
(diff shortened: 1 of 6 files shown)


containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,61 @@
+# Update the VARIANT arg in devcontainer.json to pick a Ruby version: 2, 2.7, 2.6, 2.5
```
Chuxel, line 1: You should update this comment to reference Elixr versions rather than Ruby.  You would also need to update this in your `docker-compose.yml` file rather than devcontainer.json. (See other comment on this one.)

containers/elixir-phoenix-postgres/.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: "3"
+
+services:
+  elixir:
+    build: .
```
Chuxel, line 5: If you use the "VARIANT" idea in your Dockerfile, you could add an `arg:` section that references the VARIANT.

containers/elixir-phoenix-postgres/README.md
```diff
@@ -0,0 +1,71 @@
...
+| _Works in Codespaces_       | Yes                   |
+| _Container host OS support_ | Linux, macOS, Windows |
+| _Languages, platforms_      | Elixir                |
+
+## Using this definition with an existing folder
+
+While this definition should work unmodified, you can select the version of Elixir the container uses by updating the `VARIANT` arg in the included `devcontainer.json` (and rebuilding if you've already created the container).
+
+```json
+"args": { "VARIANT": "1.10" }
```
Chuxel, line 20: You'll need to do this in `docker-compose.yml` as mentioned in other comments. Args in devcontainer.json is only for simple single container scenarios.

containers/elixir-phoenix-postgres/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+### Installing Node.js
+
+Given that Phoenix/Elixir web applications use Node.js for compiling assets, this container also includes Node.js. You can change the version of Node.js installed or disable its installation by updating the `args` property in `.devcontainer/devcontainer.json`.
+
+```json
+"args": {
```
Chuxel, line 28: Same comment here as the `VARIANT`. You'll need to do this `docker-compose.yml` since args in devcontainer.json is only for simple single container scenarios.

containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
idyll: Good call. This is what happens when you get something working late and open a PR. ;)

I'll get this all fixed up shortly.

neslinesli93:
Any news on this? It would be awesome to have Elixir supported on Codespaces :heart_eyes:

idyll:
Yeah I will try to fix tonight. :)

idyll:
Sorry this took so long @Chuxel, everything just seems to take longer this year. :(

Chuxel:
I'll try to take a another look at this soon - sorry for the delay.

containers/elixir-phoenix-postgres/.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    args:
+      # Elixir Version: 1.9, 1.10, 1.10.4, ...
+      VARIANT: 1.10
```
Chuxel, line 8: You'll need quote the values of all of these (e.g. "1.10", "1.5.4", "true", "10").  A value of `true` failed to build, and I had to quote the numbers for them to process properly.

Chuxel (changes requested):
Okay, my turn to apologize for not getting back to you faster. Time got away from me.  I ran into some issues when trying to build and noticed a couple of other things on second review.

Chuxel:
Closing in favor of #678.

PR #433, opened 2020-07-25
Created PowerShell 7 devcontainer

brettmillerb:
Created a PowerShell 7 devcontainer.

This uses the 3.1 SDK and ~3 host runtime.

```bash
FROM mcr.microsoft.com/azure-functions/dotnet:3.0-dotnet3-core-tools
```

Changed to using the dotnet image rather than the `dotnet/core/sdk:2.1` as this already contains the Azure CLI, core tools and PowerShell installed with the dotnet global tools.

I kinda followed how the other dotnet azure functions were configured. Let me know if it should be done differently.

```powershell
PS /workspaces/repo> $psversiontable
Name                           Value
----                           -----
PSVersion                      7.0.2
PSEdition                      Core
GitCommitId                    7.0.2
OS                             Linux 4.19.76-linuxkit #1 SMP Tue May 26 11:42:35 UTC 2020
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

```powershell
PS /workspaces/repo> dotnet --info
.NET Core SDK (reflecting any global.json):
 Version:   3.1.302
 Commit:    41faccf259

Runtime Environment:
 OS Name:     debian
 OS Version:  10
 OS Platform: Linux
 RID:         debian.10-x64
 Base Path:   /usr/share/dotnet/sdk/3.1.302/

Host (useful for support):
  Version: 3.1.6
  Commit:  3acd9b0cd1

.NET Core SDKs installed:
  3.1.302 [/usr/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.App 3.1.6 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.6 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET Core runtimes or SDKs:
  https://aka.ms/dotnet-download
```

```powershell
PS /workspaces/LSM_repo> func --version
3.0.2630
```

```powershell
PS /workspaces/repo> lsb_release -a

No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 10 (buster)
Release:        10
Codename:       buster
```


brettmillerb:
@TylerLeonhardt you okay to review?

containers/azure-functions-pwsh-7/README.md
```diff
@@ -0,0 +1,61 @@
+# Azure Functions & PowerShell 7
+
+## Summary
+
+*Develop Azure Functions in PowerShell. Includes .NET Core 3.1, the Azure Functions SDK, and related extensions and dependencies.*
```
TylerLeonhardt, line 5:
```suggestion
*Develop Azure Functions in PowerShell. Includes .NET Core 3.1, PowerShell 7, the Azure Functions SDK, and related extensions and dependencies.*
```

containers/azure-functions-pwsh-7/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM mcr.microsoft.com/azure-functions/dotnet:3.0-dotnet3-core-tools
```
TylerLeonhardt, line 1: There seems to be a `mcr.microsoft.com/azure-functions/powershell:3.0-powershell7-core-tools` container. Have you tried that one?

brettmillerb: Of course there is.

the dotnet one I was using is using 7.0.2 whereas the powershell 7 one is using 7.0.1 but 7.0.3 is tagged as the lts so would that not be a better dependency going forward until 7.1.0?

Updated the dockerfile to use that base image instead.

containers/azure-functions-pwsh-7/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,4 @@
+# Find the Dockerfile for mcr.microsoft.com/azure-functions/powershell:3.0-powershell7-core-tools at this URL
+# https://github.com/Azure/azure-functions-docker/blob/master/host/3.0/buster/amd64/powershell/powershell7-core-tools.Dockerfile
+
+FROM mcr.microsoft.com/azure-functions/powershell:3.0-powershell7-core-tools
```
TylerLeonhardt, line 4: gosh I wish that you could just put the image and tag in the `devcontainer.json` somehow... this `Dockerfile` is kinda silly.

@Chuxel is that possible?

Chuxel: Yes you can do it. Just use the "image" tag instead of "Dockerfile". We tend not to since it doesn't give a place for the customer to add their own software is the challenge.  You'd may want to provide the link to the full Dockerfile for those that want to customize.

TylerLeonhardt:
In this case, I don't see there being a PS8 or PS7.1 verison any time soon so maybe when we get to that point, we can refactor to the build arg you described... unless @brettmillerb wants to take a crack at that.

Chuxel:
Yeah there's an existing one for pwsh-6 is part of what I was curious about.

https://github.com/microsoft/vscode-dev-containers/tree/master/containers/azure-functions-pwsh-6

brettmillerb:
> @brettmillerb @TylerLeonhardt Yeah there's an existing one for pwsh-6 is part of what I was curious about.
> 
> https://github.com/microsoft/vscode-dev-containers/tree/master/containers/azure-functions-pwsh-6

I created that one too.

I can create the build args version but that would mean we'd be pinning the devcontainer to a specific version of PowerShell when the image in use should technically be using the GA/LTS version.

I have opened https://github.com/Azure/azure-functions-docker/pull/306 to bump the image is using to 7.0.3 as that is now the latest stable release. I would have thought the devcontainer should remain static and the version is dictated from the upstream version. Just fewer places to have versions incremented. 🤷‍♂️ 

Let me know your thoughts either way.

brettmillerb:
Thanks @Chuxel I looked at the node one and it's straight forward.

Pushed another commit with those changes. Doing it this way does VSCode unpack those options to the user when they choose an image to build from or will the user have to download the default, change and rebuild the container?

image

Chuxel:
> Pushed another commit with those changes. Doing it this way does VSCode unpack those options to the user when they choose an image to build from or will the user have to download the default, change and rebuild the container?

Not currently. The node one got merged this iteration which is why you still see multiple in the dropdown. Python 3 is an example of this now. It supports 3, 3.8, 3.7, and 3.6.

However, if someone updates the "arg", we will prompt to rebuild because the config changed, so that part is covered.  We add a note in READMEs to this effect.  e.g. here is Ruby:


> ## Using this definition with an existing folder
> 
> While this definition should work unmodified, you can select the version of Ruby the container uses by updating the `VARIANT` arg in the included `devcontainer.json` (and rebuilding if you've already created the container).
> 
> ```json
> "args": { "VARIANT": "2.7" }
> ```

There is a metadata file that could be added and used for this scenario, it's just not there yet. But this definition will be in good company.

Would it make sense to remove the other v6 container and just rename this to `azure-functions-pwsh`?

containers/azure-functions-pwsh-7/README.md
```diff
@@ -0,0 +1,61 @@
+# Azure Functions & PowerShell 7
```
TylerLeonhardt, line 1: You'll need to change this to just "PowerShell"

containers/azure-functions-pwsh-7/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Summary
+
+*Develop Azure Functions in PowerShell. Includes .NET Core 3.1, PowerShell 7, the Azure Functions SDK, and related extensions and dependencies.*
```
TylerLeonhardt, line 5: You'll need to change this to ".NET Core, PowerShell, the Azure Functions SDK...."

containers/azure-functions-pwsh-7/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. If this is your first time using a development container, please follow the [getting started steps](https://aka.ms/vscode-remote/containers/getting-started) to set up your machine.
+
+2. To use VS Code's copy of this definition:
+   1. Start VS Code and open your project folder.
+   2. Press <kbd>F1</kbd> select and **Remote-Containers: Add Development Container Configuration Files...** from the command palette.
+   3. Select the Azure Functions & PowerShell 7 definition.
+
+3. To use latest-and-greatest copy of this definition from the repository:
+   1. Clone this repository.
+   2. Copy the contents of `containers/azure-functions-pwsh-7/.devcontainer` to the root of your project folder.
```
TylerLeonhardt, line 28: This url will change

containers/azure-functions-pwsh-7/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# Find the Dockerfile for mcr.microsoft.com/azure-functions/powershell:3.0-powershell7-core-tools at this URL
+# https://github.com/Azure/azure-functions-docker/blob/master/host/3.0/buster/amd64/powershell/powershell7-core-tools.Dockerfile
```
TylerLeonhardt, line 2: Replace the 7 with ${VARIANT}

containers/azure-functions-pwsh-7/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# Find the Dockerfile for mcr.microsoft.com/azure-functions/powershell:3.0-powershell7-core-tools at this URL
```
TylerLeonhardt, line 1: Replace the 7 with ${VARIANT}

Also don't forget to change the file name now

containers/azure-functions-pwsh-7/README.md
brettmillerb: Done

containers/azure-functions-pwsh-7/README.md
brettmillerb: Done

containers/azure-functions-pwsh-7/.devcontainer/Dockerfile
brettmillerb: Done

brettmillerb:
Thanks @TylerLeonhardt 

Removed any version specific content and tidied up the folders. Tested the new VARIANT layout locally and builds and debugs fine. Might want to give the README a quick glance over in case I have missed anything there @Chuxel

Chuxel:
Ditto!  Thanks so much for the contribution. Merging.

TylerLeonhardt:
@Chuxel when do these show up in vscode?

Chuxel:
Next release. @chrmarti has been on vacation, but next week is endgame, so we should have it in insiders by then at a minimum.

Issue #436, opened 2020-07-26
Could not get jekyll container to work


jongio:
Version: 1.48.0-insider (user setup)
Commit: 731f9c25632dbbf01ee3a7892ad9d2791fe0260c
Date: 2020-07-24T05:57:10.466Z
Electron: 9.1.0
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Windows_NT x64 10.0.19041
**WSL2**

- Name of Dev Container Definition with Issue: jekyll

Steps to Reproduce:

1. Clone https://github.com/azure/azure-sdk
1. Ctrl+P to add Jekyll devcontainer
1. Create .vscode/tasks.json and copy from here: https://github.com/microsoft/vscode-dev-containers/blob/master/containers/jekyll/.vscode/tasks.json
1. Reopen in container
1. Run `bundle exec jekyll serve --livereload`
Get this
![image](https://user-images.githubusercontent.com/2163001/88480188-4c4d2180-cf09-11ea-9e4c-44cac533c0b2.png)
```
root@073cdf38c6ed:/workspaces/azure-sdk# bundle exec jekyll serve --livereload
Could not find concurrent-ruby-1.1.5 in any of the sources
Run `bundle install` to install missing gems
```
1. Run `bundle install`
Get this
![image](https://user-images.githubusercontent.com/2163001/88480224-8dddcc80-cf09-11ea-80d6-3fc4e2a10dd9.png)
```
An error occurred while installing nokogiri (1.10.8), and Bundler cannot continue.
Make sure that `gem install nokogiri -v '1.10.8' --source 'https://rubygems.org/'` succeeds
before bundling.

In Gemfile:
  github-pages was resolved to 202, which depends on
    jekyll-mentions was resolved to 1.4.1, which depends on
      html-pipeline was resolved to 2.12.2, which depends on
        nokogiri
```

1. Run the command suggested in error, get this:
```
root@073cdf38c6ed:/workspaces/azure-sdk# gem install nokogiri -v '1.10.8' --source 'https://rubygems.org/'
Building native extensions. This could take a while...
ERROR:  Error installing nokogiri:
        ERROR: Failed to build gem native extension.

    current directory: /var/lib/gems/2.5.0/gems/nokogiri-1.10.8/ext/nokogiri
/usr/bin/ruby2.5 -r ./siteconf20200726-3813-1izj1o0.rb extconf.rb
[...]

PR #446, opened 2020-07-27
added locales and forwarded port fixing #436

cmendible:
Now after cloning https://github.com/azure/azure-sdk running: 

```bash
bundle install
bundle exec jekyll serve --livereload
```
will allow to build and navigate through the site without issues.

### DIFF
--- a/containers/jekyll/.devcontainer/Dockerfile
+++ b/containers/jekyll/.devcontainer/Dockerfile
@@ -8,6 +8,14 @@ ARG USERNAME=vscode
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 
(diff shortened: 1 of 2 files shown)


jongio:
I get "locales" failed to load when I load it in GH codespaces. But it still seems to work.

Can you put in the devcontainer.json that it should call `bundle install` on open?

jongio:
```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = "en_US",
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
        LANG = "en_US.UTF-8"
```

PR #452, opened 2020-07-29
Fix to avoid writing extra dollar sign to the Dockerfile

eljog:


containers/codespaces-linux/.devcontainer/Dockerfile
```diff
@@ -151,7 +151,7 @@ RUN apt-get update -yq\
     && sudo -u codespace git config --global core.editor "/home/codespace/.local/bin/git-ed.sh" \
     #
     # Alias code-insiders to code for easy terminal file editing
-    && echo $'\nif [[ $(which code-insiders) && ! $(which code) ]]; then alias code=code-insiders; fi' >> /home/codespace/.bashrc \
+    && echo 'if [[ $(which code-insiders) && ! $(which code) ]]; then alias code=code-insiders; fi' >> /home/codespace/.bashrc \
```
jkeech, line 154: The new line makes the `.bashrc` file look a lot nicer. I think a better fix would be to force it to be executed in a bash shell:
```suggestion
    && /bin/bash -c "echo $'\nif [[ $(which code-insiders) && ! $(which code) ]]; then alias code=code-insiders; fi' >> /home/codespace/.bashrc" \
```

jkeech: Actually, the expansion doesn't work right this way :(. Need to try a different way.

containers/codespaces-linux/.devcontainer/Dockerfile
```diff
@@ -151,7 +151,7 @@ RUN apt-get update -yq\
     && sudo -u codespace git config --global core.editor "/home/codespace/.local/bin/git-ed.sh" \
     #
     # Alias code-insiders to code for easy terminal file editing
-    && echo $'\nif [[ $(which code-insiders) && ! $(which code) ]]; then alias code=code-insiders; fi' >> /home/codespace/.bashrc \
+    && /bin/bash -c "echo $'\nif [[ $(which code-insiders) && ! $(which code) ]]; then alias code=code-insiders; fi' >> /home/codespace/.bashrc" \
```
jkeech, line 154: I think this fixes the escaping:
```suggestion
    && /bin/bash -c "echo $'\nif [[ \$(which code-insiders) && ! \$(which code) ]]; then alias code=code-insiders; fi' >> /home/codespace/.bashrc" \
```

PR #455, opened 2020-07-30
Remove buster source list from VSCS image

dmgardiner25:
We had a [user report](https://github.com/MicrosoftDocs/vscodespaces/issues/644) that they were unable to install `ca-certificates` due to conflicting `apt` sources in the Oryx image. I filed an issue with Oryx but I'm not sure how long it'll take for them to fix it so this is the fix in the meantime.

### DIFF
--- a/containers/codespaces-linux/.devcontainer/Dockerfile
+++ b/containers/codespaces-linux/.devcontainer/Dockerfile
@@ -155,6 +155,9 @@ RUN apt-get update -yq\
     #
     # Clean up
     && apt-get autoremove -y \
-    && apt-get clean -y
+    && apt-get clean -y \
+    #
+    # Remove buster source list
+    && rm /etc/apt/sources.list.d/buster.list
 
 USER codespace


dmgardiner25:
@Chuxel do you know if there has been a release since the image version was changed last week?

jkeech:
@dmgardiner25, you don't need to bump the Codespaces image version. We haven't released 0.12.0 yet.

Does this change cause any other side effects or problems?

jkeech:
@Chuxel, FYI, we'll want to cut a new release for 1.130.0 sometime today or tomorrow for deployment of the Codespaces image updates next week.

dmgardiner25:
@jkeech I haven't seen any issues with the change and I don't think that there should be but would you rather we wait to merge this in until I hear back from Oryx?

jkeech:
@dmgardiner25, we should just run a test pass against the change. We can follow up with @edgonmsft offline for validation.

dmgardiner25:
@jkeech Codespaces are being created successfully and packages that have both `stretch` and `buster` versions are correctly pulling from `stretch`.

PR #460, opened 2020-08-01
Consolidate Azure Functions Node defs into one with a build arg

Chuxel:
This PR consolidates the `azure-functions-node-10` and `azure-functions-node-12` into one definition with a build arg in `devcontainer.json`

For Node 10:
```
"args": { "VARIANT": "10" }
```

For Node 12:
```
"args": { "VARIANT": "12" }
```

This will reduce maintenance as new versions come out if the image tag remains the same. We're doing this with other definitions as well.

//cc: @anthonychu @TsuyoshiUshio for input on the approach.


anthonychu:
Sounds good to me. And just in time for Node 14 coming in a couple months. 🙂

Issue #462, opened 2020-08-03
Update dev container definitions to use new method for applying script-library contents


Chuxel:
The build system has a new feature that will automatically update the contents of a folder called `library-scripts` in definitions. This simplifies `Dockerfiles` since you no longer need to use `curl` to download them or worry about a SHA to verify them. See [here for details](https://github.com/microsoft/vscode-dev-containers/tree/master/script-library#copying-the-script-to-devcontainerlibrary-scripts).

This issue tracks updates to definitions to use it:

- [x] container-templates
- [x] alpine
- [x] azure-ansible #472 
- [x] azure-blockchain #476
- [x] azure-cli  #473
- [x] azure-functions-dotnetcore-2.1
- [ ] azure-machine-learning-python-3
- [x] azure-terraform
- [x] bazel  #473
- [x] codespaces-linux (#500)
- [x] dapr-dotnetcore-3.1
- [x] dapr-typescript-node-12
- [x] dart
- [x] debian
- [x] docker-from-docker #474
- [x] docker-from-docker-compose #474
- [x] dotnetcore
- [x] dotnetcore-fsharp
- [x] go
- [x] java
- [x] javascript-node
- [x] jekyll
- [x] kubernetes-helm #474
- [x] perl  #473
- [x] php, php-mariadb
- [x] plantuml
- [x] powershell
- [x] python-3
- [x] python-3-anaconda #466
- [x] python-3-miniconda #466
- [x] r  #473
- [x] ruby
- [x] ruby-rails
- [x] ruby-sinatra
- [x] rust  #473
- [x] swift
- [x] ubuntu

PR #466, opened 2020-08-03
Update python anaconda/miniconda to use common script

Chuxel:
This PR updates the Anaconda and Miniconda definitions to use the common script.  It uses the new approach for doing this outlined in #462.

It otherwise follows the same steps to avoid unexpected impacts.


Chuxel:
@brettcannon No worries! Vacation is important. :) Thanks for the review.

Issue #459, opened 2020-08-01
Java extension no longer supports Java 8, update Azure Functions to JDK 11?


Chuxel:
It appears that Eclipse and by extension the Java VS Code extension no longer supports JDK 8. So the following error appears:

![image](https://user-images.githubusercontent.com/312252/89090534-53848d00-d358-11ea-90fb-48b02cc872fd.png)

I'm thinking we should probably remove `azure-functions-java-8` or update it to JDK 11 even though that is in preview.  Thoughts?

//cc: @anthonychu @TsuyoshiUshio

anthonychu:
@TsuyoshiUshio Is VS Code currently broken for Java 8 function apps?

/cc @apawast

anthonychu:
My read on [this issue](https://github.com/redhat-developer/vscode-java/issues/1543#issuecomment-663382692) is that the _extension_ now requires Java 11, but can be configured to work with Java 8 projects. A Java 8 Functions devcontainer with this pre-configured would actually be useful for helping customers get up and running.

Chuxel:
>My read on this issue is that the extension now requires Java 11, but can be configured to work with Java 8 projects. A Java 8 Functions devcontainer with this pre-configured would actually be useful for helping customers get up and running.

@anthonychu @TsuyoshiUshio @apawast Got it. Quick ping here since this is "endgame" week for the next VS Code release. Should I assume that the image referenced in this definition will get updated to include the Java 11 runtime as well, or should we tactically remove it for this release?

TsuyoshiUshio:
I think we should keep Java 8 and add Java 11. I already has the image. We support both.

anthonychu:
Yes we should have devcontainers for both Java 8 and Java 11. I think the problem here is the latest version of the VS Code Java extension now runs on Java 11, which is missing from the image. Both Java 8 and 11 need to be in the image and there needs to be config in devcontainer.json to set [`java.configuration.runtimes` and `java.home`](https://github.com/redhat-developer/vscode-java/issues/1543#issuecomment-663382692).

TsuyoshiUshio:
Thank you, @anthonychu , I'll reproduce it and fix it.

TsuyoshiUshio:
Hi @anthonychu and @Chuxel , 
I create a quick fix PR. I tested it and make sure it works on Java8 function app. 
I'm on call now, so I'll create java 11 PR soon.

amamounelsayed:
[...]

PR #468, opened 2020-08-04
fix: VS code java extension requires java 11

TsuyoshiUshio:
Fix: https://github.com/microsoft/vscode-dev-containers/issues/459

VSCode Java extension requires Java 11 for the extension. 

I update the image to install not only Java 11 but also Java 8
I tested it with deploying Functions App with java8. 

Hi @Chuxel, @anthonychu , and @amamounelsayed  could you have a look? 

## TODO 
This is a quick fix. For supporting java 8, A lot of people using Java 8.  
I'll also provide java 11 soon.

### DIFF
--- a/containers/azure-functions-java-8/.devcontainer/Dockerfile
+++ b/containers/azure-functions-java-8/.devcontainer/Dockerfile
@@ -1,6 +1,15 @@
 # Find the Dockerfile for mcr.microsoft.com/azure-functions/java:3.0-java8-core-tools at this URL
 # https://github.com/Azure/azure-functions-docker/blob/master/host/3.0/buster/amd64/java/java8/java8-core-tools.Dockerfile
-FROM mcr.microsoft.com/azure-functions/java:3.0-java8-core-tools
+FROM mcr.microsoft.com/azure-functions/java:3.0-java11-core-tools
+
+RUN apt-get -qq update && \
+    apt-get install -y software-properties-common && \
+    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0x219BD9C9 && \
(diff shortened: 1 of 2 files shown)


containers/azure-functions-java-8/.devcontainer/Dockerfile
```diff
@@ -1,6 +1,15 @@
 # Find the Dockerfile for mcr.microsoft.com/azure-functions/java:3.0-java8-core-tools at this URL
 # https://github.com/Azure/azure-functions-docker/blob/master/host/3.0/buster/amd64/java/java8/java8-core-tools.Dockerfile
-FROM mcr.microsoft.com/azure-functions/java:3.0-java8-core-tools
+FROM mcr.microsoft.com/azure-functions/java:3.0-java11-core-tools
```
amamounelsayed, line 3: Why java 8 depends on java11?

TsuyoshiUshio: VSCode java extension is only available on Java 11.  It is workaround, however, we want to keep project as 8 and want to use VSCode extension, we need to install both Java 11 and Java 8. 
I was consider if we change the auzre-functions-docker, however, including both images for the CI user is not good. So I add to install Java 8 later in the Dockerfile.

anthonychu:
Looks great. Would it make more sense if it used `FROM mcr.microsoft.com/azure-functions/java:3.0-java8-core-tools` and installed Java 11 in the Dockerfile?

anthonychu:
Yep I agree if others are good with it too. I think this is a good fix for Java 8. Can sort out how to do a single image for Java 8 and 11 In the next release.

anthonychu:
\+ @apawast PM for Java Functions

PR #469, opened 2020-08-05
Azure Functions Java 11 DevContainer

TsuyoshiUshio:
## A part of Fix 
https://github.com/microsoft/vscode-dev-containers/issues/459

This is the Java Functions Container for Java 11. 
I tested, it works. 

## NOTE
We need to wait until the end of this month.  VSCode Azure Functions extension will estimated that they will support it at the end of this month. 

https://github.com/microsoft/vscode-azurefunctions/issues/2033

Until then, we can use Azure Functions Core Tools to create/deploy the functions. 
For more details, refer to [Quickstart: Create a function in Azure that responds to HTTP requests](
https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-azure-function-azure-cli?tabs=bash%2Cbrowser&pivots=programming-language-java)

### DIFF
--- a/containers/azure-functions-java-11/.devcontainer/Dockerfile
+++ b/containers/azure-functions-java-11/.devcontainer/Dockerfile
@@ -1,0 +1,20 @@
+# Find the Dockerfile for mcr.microsoft.com/azure-functions/java:3.0-java8-core-tools at this URL
+# https://github.com/Azure/azure-functions-docker/blob/master/host/3.0/buster/amd64/java/java8/java8-core-tools.Dockerfile
+FROM mcr.microsoft.com/azure-functions/java:3.0-java11-core-tools
+
+# Uncomment following lines If you want to enable Development Container Script
+# For more details https://github.com/microsoft/vscode-dev-containers/tree/master/script-library
+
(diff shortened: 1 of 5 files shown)


Chuxel:
@anthonychu @TsuyoshiUshio Did you want me to look at merging this or should we wait for Option 2 as described here: https://github.com/microsoft/vscode-dev-containers/issues/459#issuecomment-668979824

anthonychu:
Will leave it up to @TsuyoshiUshio to decide if this is a good path to take right now or if we should take the time to figure out how to use a single container. I think the different settings required for things like java.home could be tricky to configure easily for both.

TsuyoshiUshio:
IMO, I'd keep both Java11 and Java8.  
The reason is, Java 11 image will download faster than an image that contains both SDK.  I believe people eventually move to java 11 soon. 
Java8 image will have two SDK , Java11 and Java8. and will be bigger sized. However, it might also encourage to people to move java 11.  However, it is just my opinion. I also think @anthonychu 's idea got point. I trust your decision. :)

Chuxel:
Sounds good - will merge then.

PR #478, opened 2020-08-06
Update Ruby-based definitions to use new pre-built dev containers Ruby definition

Chuxel:
This PR switches the RoR, Ruby Sinatra, and Jekyll definitions to use the pre-built MCR Ruby image that contains most of what the Dockerfiles in these definitions added. This is related to #154 

//cc: @cmendible as FYI


cmendible:
@chuxel sorry I´m late. I can confirm this one is working!

PR #487, opened 2020-08-12
Add: WordPress container

dinhtungdu:
### Description of the Change

This PR adds a WordPress definition for plugin/theme development:

- Auto-installs and configures WordPress development environment without any configuration.
- Uses `docker-compose` to set up Nginx, PHP-FPM, and MariaDB.
- Auto-installs composer/npm dependencies if needed.
- Preinstalls WP CLI.
- PHP and node.js versions can be changed using config file.
- xDebug is installed and can be toggled.


Chuxel:
Thanks for the contribution!!! I need to take a look at this more deeply, but a few comment / questions:

It would be ideal if we could avoid the use of a 3rd party image since those contents can change without developers knowing about it. Generally images are either from the vendor (e.g. puppet), the language team, or are an official Docker image.  That said, there's actually a PHP image that has Xdebug in it that might do the trick: `mcr.microsoft.com/vscode/devcontainers/php` (which also has NVM and yarn in it) or perhaps https://hub.docker.com/_/wordpress image as a base?

Why use initalizeCommand instead of using a Dockerfile? The Dockerfile would have the advantage of caching and you could use build args in the docker-compose.yml file for options. Similar question for postCreateCommand.  You can see an example of using args in the php-mariadb definition: https://github.com/microsoft/vscode-dev-containers/tree/master/containers/php-mariadb

Chuxel:

Issue #397, opened 2020-07-10
Add Go image to Docker Hub or similar so we can use as a base image for ./devcontainer/Dockerfile, rather than cloning.


leighhopcroft:

Chuxel:
@hyangah Would adding `"go.useGoProxyToCheckForToolUpdates": false` to settings make the tools stable enough that a pre-built image would be viable?  Part of the reason I'd held off was how often the tools updated and that people kept getting toasts to update.  A pre-built image would be even more likely for that to happen.

hyangah:
@Chuxel How frequently does the container image get updated? Since `gopls` is under active development, I think update notification is a feature. But yes, it will be annoying if users have to install the `gopls` whenever starting a new session.

@stamblerre @Chuxel Is it possible to arrange the container image release synchronized with gopls stable version release?

Chuxel:
@hyangah The images in this repo sym-ship with VS Code releases primarily, so I don't know that we can do that from here realistically. What teams that need different release schedules have done is create their own image and just reference it from here.  As you've seen, there's a common script that can be referenced either via curl or directly.

Is the issue just `gopls`? We could have that one in the Dockerfile while the rest are pre-built in the image I suppose. Even that would help quite a bit.

hyangah:
If it's aligned with VS Code releases, it will be updated monthly. That's frequent enough to pick up the recent version. 
So, disabling the tool update check seems ok.

For now, yes, only `gopls` is utilizing this feature.

Chuxel:
@hyangah Great! Do people get notified if the other tools are out of date? Does setting `"go.useLanguageServer": true` change that at all?

hyangah:
@Chuxel We don't have version checks for other tools yet. Only the `gopls`, which is used when `"go.useLanguageServer": true`.

PR #492, opened 2020-08-14
Onboard Go image to build

Chuxel:
This PR will onboard the Go dev container into the image generation process.  It is part of #154 and resolves https://github.com/microsoft/vscode-dev-containers/issues/397 and helps with #375.

Currently it will output the following image variants for Go versions:

```
mcr.microsoft.com/vscode/devcontainers/go:1
mcr.microsoft.com/vscode/devcontainers/go:1.15
mcr.microsoft.com/vscode/devcontainers/go:1.14
mcr.microsoft.com/vscode/devcontainers/go:1.13
```

...based on the [contents of `base.Dockerfile`](https://github.com/microsoft/vscode-dev-containers/blob/clantz/go-onboard/containers/go/.devcontainer/base.Dockerfile). 

I kept 1.13 in the list since 1.15 just came out three days ago @hyangah LMK if you think we should just not bother.

Note that we generally we drop no longer supported versions of go from the build process once the version is out of support. However, existing tags, just with an older image. 

In the next VS Code release, the default Dockerfile in the container will use [this instead](https://github.com/microsoft/vscode-dev-containers/blob/clantz/go-onboard/containers/go/.devcontainer/Dockerfile), but without the "dev-" prefix. The image includes `nvm`, but doesn't install Node in the image since it isn't always needed for Go development, but can be useful in web development involving Go.

The `devcontainer.json` then allows the selection of the Go version and whether no should be installed via build args.

//cc: @hyangah for input

### DIFF
--- a/containers/go/README.md
+++ b/containers/go/README.md
@@ -8,6 +8,9 @@
 |----------|-------|
 | *Contributors* | The VS Code Team |
 | *Definition type* | Dockerfile |
+| *Published images* | mcr.microsoft.com/vscode/devcontainers/go |
+| *Available image variants* | 1, 1.15, 1.14, 1.13 |
+| *Published image architecture(s)* | x86-64 |
 | *Works in Codespaces* | Yes |
 | *Container host OS support* | Linux, macOS, Windows |
 | *Languages, platforms* | Go |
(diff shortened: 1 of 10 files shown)


Chuxel:
//cc: @jkeech since I mentioned this to him as well.

containers/go/README.md
```diff
@@ -20,6 +23,34 @@ While the definition itself works unmodified, you can select the version of Go t
 "args": { "VARIANT": "1.14" }
 ```
 
+You can also directly reference pre-built versions of `.devcontainer/base.Dockerfile` by using the `image` property in `.devcontainer/devcontainer.json` or updating the `FROM` statement in your own  `Dockerfile` to one of the following. An example `Dockerfile` is included in this repository.
+
+- `mcr.microsoft.com/vscode/devcontainers/go` (latest)
+- `mcr.microsoft.com/vscode/devcontainers/go:1`
+- `mcr.microsoft.com/vscode/devcontainers/go:1.15`
+- `mcr.microsoft.com/vscode/devcontainers/ruby:1.14`
```
hyangah, line 31: ruby? is it just a coincidence or a real error?

Chuxel: That's a typo.  Good catch!  Copy-paste from the ruby definition and I missed one. :smile:

I'll remove `1.13` then.

Going forward, we'll just need to update the "variants" property in definition-manifest.json and the README as new versions come out and the previous one is deprecated. The rest should stay the same.

PR #493, opened 2020-08-15
Onboard java to build

Chuxel:
This PR on-boards the java container into the image build process and is part of work for #154.

Currently it will output the following image variants for Java versions:

```
mcr.microsoft.com/vscode/devcontainers/java:11
mcr.microsoft.com/vscode/devcontainers/java:14
```

...based on the [contents of `base.Dockerfile`](https://github.com/microsoft/vscode-dev-containers/blob/clantz/onboard-java/containers/java/.devcontainer/base.Dockerfile). 

Note that we generally we drop no longer supported versions of Java from the build process once the version is out of support. However, existing tags, just with an older image. 

**Note that this does not include Java 8** because the Java extension seems to require it be set to Java home, and there appears to be no good way with symlinks to get this to work consistently because `"java.configuration.runtimes"` appears to force you to enter a specific `"name"` value. It would need its own dedicated build.  (As an aside, I'm really surprised there isn't a `"java.extension.home"` property or something like that added.)

In the next VS Code release, the default Dockerfile in the container will use [this instead](https://github.com/microsoft/vscode-dev-containers/blob/clantz/onboard-java/containers/java/.devcontainer/Dockerfile), but without the "dev-" prefix. The image includes `nvm`, but doesn't install Node in the image since it isn't always needed for Java development, but can be useful in web development involving Java.  

It also has scripts to install Maven and Gradle referenced from both dockerfiles, but they are not in the image given `mvnw` / `gradlew` are often used.

//cc: @testforstephen @akaroml - Let me know what you think.

### DIFF
--- a/containers/java/.devcontainer/Dockerfile
+++ b/containers/java/.devcontainer/Dockerfile
@@ -1,46 +1,32 @@
-# Update the VARIANT arg in devcontainer.json to pick a Java version >= 11
 ARG VARIANT=11
-FROM openjdk:${VARIANT}-jdk-buster
+FROM mcr.microsoft.com/vscode/devcontainers/java:${VARIANT}
 
-# Options for setup script
-ARG INSTALL_ZSH="true"
-ARG UPGRADE_PACKAGES="false"
-ARG USERNAME=vscode
-ARG USER_UID=1000
-ARG USER_GID=$USER_UID
-
-# Install needed packages and setup non-root user. Use a separate RUN statement to add your own dependencies.
-COPY library-scripts/*.sh /tmp/library-scripts/
-RUN /bin/bash /tmp/library-scripts/common-debian.sh "${INSTALL_ZSH}" "${USERNAME}" "${USER_UID}" "${USER_GID}" "${UPGRADE_PACKAGES}" \
-    && if [ ! -d "/docker-java-home" ]; then ln -s "${JAVA_HOME}" /docker-java-home; fi \
-    && apt-get clean -y && rm -rf /var/lib/apt/lists/* /tmp/library-scripts/common-debian.sh
+# [Optional] Install a version of Node.js using nvm for front end dev
+ARG INSTALL_NODE="true"
+ARG NODE_VERSION="lts/*"
+RUN if [ "${INSTALL_NODE}" = "true" ]; then su vscode -c "source /usr/local/share/nvm/nvm.sh && nvm install ${NODE_VERSION} 2>&1"; fi
 
 # [Optional] Install Maven
 ARG INSTALL_MAVEN="false"
 ARG MAVEN_VERSION=3.6.3
 ARG MAVEN_DOWNLOAD_SHA="no-check"
 ENV MAVEN_HOME=/usr/local/share/maven
+COPY library-scripts/maven-debian.sh /tmp/library-scripts/
 RUN if [ "${INSTALL_MAVEN}" = "true" ]; then /bin/bash /tmp/library-scripts/maven-debian.sh ${MAVEN_VERSION} ${MAVEN_HOME} ${USERNAME} ${MAVEN_DOWNLOAD_SHA}; fi \
-    && rm -f /tmp/library-scripts/maven-debian.sh
+    && rm -f /tmp/library-scripts
 
 # [Optional] Install Gradle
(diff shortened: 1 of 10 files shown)


testforstephen:
> Note that this does not include Java 8 because the Java extension seems to require it be set to Java home, and there appears to be no good way with symlinks to get this to work consistently because "java.configuration.runtimes" appears to force you to enter a specific "name" value. It would need its own dedicated build. (As an aside, I'm really surprised there isn't a "java.extension.home" property or something like that added.)

There is a "java.home" user setting exposed by vscode-java extension. And the Java extension will check this setting first, then the JAVA_HOME environment variable.

About running Java 8 application with latest vscode-java extension, i replied a comment at https://github.com/microsoft/vscode-dev-containers/issues/495#issuecomment-674693874.

containers/java/.devcontainer/Dockerfile
```diff
@@ -1,46 +1,32 @@
...
 RUN if [ "${INSTALL_MAVEN}" = "true" ]; then /bin/bash /tmp/library-scripts/maven-debian.sh ${MAVEN_VERSION} ${MAVEN_HOME} ${USERNAME} ${MAVEN_DOWNLOAD_SHA}; fi \
-    && rm -f /tmp/library-scripts/maven-debian.sh
+    && rm -f /tmp/library-scripts
 
 # [Optional] Install Gradle
 ARG INSTALL_GRADLE="false"
 ARG GRADLE_VERSION=5.4.1
 ARG GRADLE_DOWNLOAD_SHA="no-check"
 ENV GRADLE_HOME=/usr/local/share/gradle
+COPY library-scripts/maven-gradle.sh /tmp/library-scripts/
```
testforstephen, line 23: typos?
=> COPY library-scripts/gradle-debian.sh /tmp/library-scripts/

containers/java/.devcontainer/base.Dockerfile
```diff
@@ -0,0 +1,46 @@
+# Update the VARIANT arg in devcontainer.json to pick a Java version: 11, 14
+ARG VARIANT=11
+FROM openjdk:${VARIANT}-jdk-buster
```
testforstephen, line 3: The purpose of adding a base.Dockerfile and Dockerfile is a little confusing to me. I see the main difference is the base image, which one do you recommend the user to use?

Chuxel: The base.Dockerfile is used to generate the image. The Dockerfile is the one that will end up in VS Code. I'll update `devcontainer.json` to point to Dockerfile once the image has been published.  People do come here and just grab the `.devcontainer` folder so I don't want to break that. 😄 

[Lots more detail here](https://github.com/microsoft/vscode-dev-containers/blob/master/build/README.md) if you are interested.

containers/java/.devcontainer/Dockerfile
Chuxel: Ack!  Yeah will fix.

containers/java/.devcontainer/base.Dockerfile
Chuxel: I should also add that the key difference is that all of the additional tools, zsh, etc are built into this image along with nvm and yarn for web development.

I'd wondered about including a version of gradle or maven - but it seems like that would age quickly.  Node.js itself isn't in the image, but nvm is which makes installing an updated version really easy and quick.

containers/java/.devcontainer/base.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+# Options for setup script
+ARG INSTALL_ZSH="true"
+ARG UPGRADE_PACKAGES="false"
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Install needed packages and setup non-root user. Use a separate RUN statement to add your own dependencies.
+COPY library-scripts/*.sh /tmp/library-scripts/
```
chrmarti, line 13: Only copy `common-debian.sh`?

Chuxel: Ack - looks like a missed a couple with a refactor.

Each section should copy its own script so that updating it will only rebuild that section.  If you copy all up front, any change to any script is a full rebuild.  Will fix - good catch!

PR #494, opened 2020-08-15
Codespaces: Adding the "less" tool

lostintangent:
This PR simply adds the `less` utility to the Codespaces universal Linux image, since this is a common/valuable tool for many developers 👍


Chuxel:
LGTM! We probably should switch this definition to using the common script which includes it already (along with other useful things like jq)

PR #499, opened 2020-08-17
Prevent docker scripts to spawn yet another process

felipecassiors:
The exec replaces the current process with the new process, so we don't have yet another process running in the process tree.


Chuxel:
Makes sense!  Thanks!!

Issue #496, opened 2020-08-15
[Discussion] Suggested refactoring of codespaces-linux


Chuxel:
The `codespaces-linux` is a bit of a on-off in this repo which makes it a bit harder to maintain. There's also some things it does that don't seem needed that significantly extend build times. Here's a suggested list of updates:

- [x] Use [common-debian.sh](https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/common-debian.sh) to install common utilities, Live Share dependencies, setup the codespaces user, install zsh, and oh-my-zsh. This is what the other definitions do already.
- [x] Use the [node-debian.sh](https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/node-debian.sh) to install nvm for constancy sake (fewer places to update if we want to adjust)
- [x] Drop the use of the live share dependency script given this would be covered by the above.
- [ ] <strike>No longer build git - this seems redundant and unneeded given it is already in the Oryx image.</strike>
- [x] Move installations into library scripts stored in /library-scripts in this repo so they can be reused across the rest of the repo and updated as needed together.
- [x] Use Oh My Zsh! instead of a custom zsh prompt
- [x] Do a bit more layering to improve download concurrency (being sure to clean up after the end of a layer).

//cc: @edgonmsft @jkeech @lostintangent

Chuxel:
Resolved with #500

Issue #495, opened 2020-08-15
Version of Java in Codespaces image no longer works with the Java extension


Chuxel:
The Java extension now requires JDK 11, while the OS version of Java is only 8.  It is possible to do JDK 8 work if both JDKs are installed and the exact right configuration is done in VS Code.  However, this is requires very specific devcontianer.json config. (See https://github.com/redhat-developer/vscode-java/issues/1543)

To resolve this problem, we either need to:

1. Install OpenJDK 11 and add it as the new JAVA_HOME
2. Or drop Java support in the main Codespaces Linux image in favor of the pre-built Java one being PR'd in #493.

//cc: @edgonmsft @lostintangent @jkeech @nikmd23 @matthewisabel

akaroml:
@Chuxel the Java 11 is only required by the Java Language Server itself. Users can still work with Java 8 or any other Java version in their projects. So the use case of having both Java 8 and 11 installed should work for sure.

@testforstephen should be able to help.

testforstephen:
This requirement only asks you to install JDK 11 to run the vscode-java extension itself, because it's also a Java program and its code is compiled to Java 11 version. If you want to run a Java 8 application in dev container, you need install two JDKs and configure the dual JDKs in the user setting below.
```
{
    "java.home": "C:\\AdoptOpenJDK\\jdk-11.0.8.10-hotspot", // This is used to launch vscode-java extension itself.
    "java.configuration.runtimes": [
        {
          "name": "JavaSE-1.8",
          "path": "C:\\Java\\openjdk8u181-b13" // This tells Java extension to use the specified JDK to run Java 8 application.
        }
    ],
}
```

Chuxel:
Yeah we'll have to create a dedicated definition for it. We can onboard it into image build as well - but I queued up JDK 11/14 for that in the mean time.  The image build system supports multiple definitions contributing to the same repository for this reason.

Chuxel:
I've got a PR out that will tactically add JDK 11 in and make it the default (and leave the existing OS installed JDK 8).  Oryx will be adding JDK 8/11 into that image as well so this would be a point in time thing.

edgonmsft:
By the way another thing is that the git compile/installation we have is because the version of git included in the image is too old. @dmgardiner25 has more details I think.

dmgardiner25:
[...]

PR #500, opened 2020-08-17
Codespaces image refactor, add JDK 11, Rust

Chuxel:
This PR and implements the proposed improvements in https://github.com/microsoft/vscode-dev-containers/issues/496 and picks up some technical debt items (#462, #409).  It also adds Rust and JDK 11 into the image since the Java extension now requires JDK 11 (even to use JDK 8 to develop - see #495).

It:

- [x] Uses existing common scripts like [common-debian.sh](https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/common-debian.sh) to install common utilities, Live Share dependencies, setup the codespaces user, install zsh, and oh-my-zsh. This is what the other definitions do already.
    - [x] Drops the use of the live share dependency script given this would be covered by the above.
    - [x] Uses Oh My Zsh! instead of a custom zsh prompt
- [x] Adds new common scripts for things that will be installed like Rust and Java since these can then be used in other Containers to simplify maintenance.
    - [x] Adds new/updated scripts to /library-scripts in this repo for reuse (and auto-update across defintions via a CI job when changed).
- [x] Does **not** add Docker into the generated image, but wires it in via a custom Dockerfile for local use (via `base.Dockerfile` for build and `Dockerfile` for inclusion in the npm package in this repo).
- [x] No longer builds git - this seems redundant and unneeded given it is already in the Oryx image.
- [x] Does a bit more layering to improve download concurrency (being sure to clean up after the end of a layer).
- [x] Adds JDK 11 and sets it to the default
- [x] Adds Rust and related tools

EDIT: It's probably easiest to review this by pulling down the code. Things moved around enough the diff is pretty useless. :smile:


containers/codespaces-linux/.devcontainer/setup-user.sh
```diff
@@ -0,0 +1,44 @@
...
+{ echo && echo $BASH_PROMPT ; } | tee -a /etc/skel/.bashrc /root/.bashrc /home/${USERNAME}/.bashrc >> /etc/bash.bashrc
+{ echo && echo -e "export PATH=${USER_EXTRA_PATHS_OVERRIDES}:\$PATH:${USER_EXTRA_PATHS}" ; } | tee -a /etc/skel/.bashrc /etc/skel/.zshrc /root/.bashrc /root/.zshrc /home/${USERNAME}/.bashrc /home/${USERNAME}/.zshrc >> /etc/bash.bashrc
+echo "Defaults secure_path=\"/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/bin:${EXTRA_PATHS}\"" >> /etc/sudoers.d/securepath
+
+# Install and setup fish
+apt-get install -yq fish
+printf "$FISH_PROMPT" >> /etc/fish/conf.d/fish_prompt.fish
+
+# Add user to a Docker group
+sudo -u ${USERNAME} mkdir /home/${USERNAME}/.vsonline
```
edgonmsft, line 24: FYI I wanted to remove this line due to the rename of the filepaths, meaning `.vsonline` is going away,

Chuxel: Cool, should it change or just get yanked?

edgonmsft: Yanked, but let me think about the order of the changes first.

Chuxel:
As a FYI - Sounds like Oryx will get Java and Ruby into the VSO image in the next month or so. 🎉   I'll leave Java for the moment, but once we switch to the updated image, we can yank it.

Rust is here, so Go is the main one that would make sense to add then.

Chuxel:
Per https://github.com/microsoft/vscode-dev-containers/issues/495#issuecomment-675641929 we'll want to include the git build sadly.  I'll add that back in.

jkeech:
@Chuxel, what is the size of the base image after these changes? We want to ensure we aren't adding too much where it eats into the storage that each Codespace has available (the bigger the docker images on disk, the less room for the git repo, packages, build artifacts, etc).

Chuxel:
@jkeech Ignoring the JDK 11, Rust, and planned Go additions, it would be slightly less I suspect. 

With those additions, it will get bigger, but I wouldn't expect too much. The JDK wasn't that big, Rust isn't huge, neither is Go. In fact, most of the "heft" was because of installing JDK 8 from the OS --- that seems to drag a ton of OS dependencies in given it was installed w/o `--no-install-recommends`.

That said, the vast majority still comes from the upstream Oryx image > 8GB. It's really all the dozens of versions of .NET and Python under `/opt` that drive most of the size.  I'd love to trim down those honestly.

Will get a final tally and post an update.

The additions are a tradeoff, no doubt - but discussing w/@matthewisabel and @lostintangent, nd given our startup characteristics with custom images right now, we likely need to have a slightly bigger kitchen sink, at least near term despite the downsides.

jkeech:
> The additions are a tradeoff, no doubt - but discussing w/@matthewisabel and @lostintangent, nd given our startup characteristics with custom images right now, we likely need to have a slightly bigger kitchen sink, at least near term despite the downsides.

I agree. Just noting that we are at 10GB today, which is already huge, so part of the consideration of adding new runtimes/tools to the image should include a tradeoff analysis of the size of the addition compared to proportion of the userbase who would benefit from the addition. If we have a small tool that 40% of users would benefit from, it's a clear win. If we have a 1GB runtime that only 2% of users will use, it's less clear that we should add it. We should probably formalize the criteria that we use to decide what is included in the base image so we can evaluate future requests.

Chuxel:
Okay, I'm a good guesser. It's about ~12.2GB (with 12GB the guess I told Matt).  Here's the breakdown:

* Oryx - 8.38GB
* Top section - 962MB. (Most of this is the Azure CLI  **594M**! 😳  - that's so huge I'm removing it to double check.)
* Git from source - 189MB
* PowerShell + .NET Scripts - 162MB
* JDK 8+11 - 698MB (half of which is already in the current image)
* Rust + Tools - 1GB (777M w/o tools)
* Go + Tools - 751MB (377MB w/o tools)

If Docker is added, that's 192MB, but that wouldn't be in the image itself.

Looking at specific folders in the container:
```
310M    /opt/java                         <= JDK 11
1.6G    /opt/nodejs
153M    /opt/microsoft
594M    /opt/az                           <= Azure CLI 😳
231M    /opt/php
47M     /opt/startupcmdgen
80M     /opt/buildscriptgen
5.2M    /opt/yarn
60K     /opt/npm
663M    /opt/python
45M     /opt/hugo
1.9M    /opt/php-composer
1.6G    /opt/dotnet
16K     /opt/oryx
377M    /usr/local/go                    <= Go w/o tools
777M    /usr/local/rustup                <= Rust + tools
12M     /usr/local/cargo                 <= Mostly pointers to rustup
374MB   /usr/local/bin                   <= Lots of utilities (GH CLI, Kubernetes, Helm), Go tools
= ~6.8 GB (/opt is 5.2 GB)
```
JDK 8 and OS installed tools (e.g. OS python, C++, all dev libs, etc) are not in this list.

jkeech:
Thanks for the breakdown! What are your thoughts on the size tradeoff relative to the percent of the users who benefit from it? Where would you draw the line on languages/tools?

Chuxel:
@jkeech I confirmed the Azure CLI really is that huge - that's crazy!

Honestly, I kind of want to yank that and the custom git build -- We'd get Go for free for that amount of storage.  We could potentially do a breaking change version update so anyone using the old version of the base image for a custom container doesn't get busted by it.

@matthewisabel @lostintangent and can chat.  I pushed the updated script that has Go in it as well if you want to try it.

jkeech:
@Chuxel, `az` CLI was explicitly added as part of this issue: https://github.com/MicrosoftDocs/vscodespaces/issues/101. We could of course revisit that decision, but there was a desire to make sure the `az` CLI experience works well in Codespaces to support Azure users.

Chuxel:
@jkeech Yep -  new info have makes me question it a bit more than in the past. 😉 We can offline that part.

Chuxel:
@jkeech @edgonmsft @matthewisabel - @lostintangent and I got together on this quickly.  Considering Ruby and Java will be in the base Oryx image, this is getting pretty tight. Going over 12GB just feels too much.  

What we're thinking is we do the following:

1. Keep Rust and Go out of this image for now
2. However, include the "bootstrap" scripts in the image in a location we can tell customers about in order to add them to the kitchen sink (e.g. `~/bootstrap-scripts`)
    - This scripts will also be used for the Go / Rust images in vscode-dev-containers to help ensure they're kept working 
3. We can then route people down the custom flow as improvements are made there.

We could make sub-images as well easily enough, but with the starting point being 10-12GB, this doesn't seem to have a ton of value over custom images.

Make sense?

Chuxel:
Found some hidden caches and temp files that further reduces the resulting image size.  Also got Xdebug added for PHP. The build system now has flattening support, so this should actually be smaller than the current image with these latest cleanup steps.

@jkeech LMK if you want to move forward and merge.

jkeech:
@chuxel, I haven’t tested your changes, but conceptually everything looks good to me. Feel free to merge anytime! We can pick it up tomorrow and validate and probably ship it this week.

Chuxel:
Sounds good. Let's merge and iterate from here.

edgonmsft:
Checking this one I think PHP is no longer in the path. Looking at where that got lost...

Chuxel:
@edgonmsft It should be there. It's added by `benv` - make sure you pull latest. I fixed this last night.

PR #517, opened 2020-08-24
Add JDK 8 definition

Chuxel:
This PR adds a Java 8 definition.  It builds up from the java definition and reuses the `mcr.microsoft.com/vscode/devcontainers/java:11` image and installs Java 8 using SDKMAN!  The appropriate configuration is then in devcontainer.json to use JDK 11 for the extension but target JDK8.

It will result in an `mcr.microsoft.com/vscode/devcontainers/java:8` image

Note that to test, you will need to update `devcontainer.json` to reference `base.Dockerfile` since it defaults to the Dockerfile that uses the pre-built image for those that copy the files directly out of this repo.


felipecrs:
@Chuxel In case you are interested, this is how I do to automatically download the latest version of Java 8 with SDKMAN!.

https://github.com/felipecrs/dotfiles/blob/master/run_once_setup.sh#L137-L143

Chuxel:
@felipecrs Nice! Will look at incorporating that. I figured I could regex my way there, but this should do the trick!

testforstephen (approved):

Issue #512, opened 2020-08-22
Do not update user's dotfiles


felipecrs:
In the scripts from script's library, often we see the function `updaterc` updating the user's dotfiles such as `~/.bashrc`. I think we should never do that since those files will be later replaced in case the user has a `dotfiles` repository configured. 

So I have two suggestions:

## First 
Instead, perform these changes to the global `bashrc` and `zshrc` files, which are respectively `/etc/bash.bashrc` and `/etc/zsh/zshrc`. Those files gets sourced by bash and zsh automatically, for all the users. 

This means that for free, we would earn synchronization between root and non-root user without having to update both user's dotfiles.

## Second

Update `/etc/bash.bashrc` and `/etc/zsh/zshrc` to include something like `. /etc/profile`, because unfortunately, in Docker, `/etc/profile` does not get read automatically.

Then, if you take a look at `/etc/profile` you'll see something like: 

```sh
if [ -d /etc/profile.d ]; then
  for i in /etc/profile.d/*.sh; do
    if [ -r $i ]; then
      . $i
    fi
  done
  unset i
fi
```

This means that any `.sh` under `/etc/profile.d/` will get sourced by `/etc/profile`, and this is the recommended way to customize the system-wide Linux environment.

Then, we could convert lines such as:

```bash
echo "export SDKMAN_DIR=${SDKMAN_DIR}\nsource \${SDKMAN_DIR}/bin/sdkman-init.sh" >> /home/vscode/.bashrc
echo "export SDKMAN_DIR=${SDKMAN_DIR}\nsource \${SDKMAN_DIR}/bin/sdkman-init.sh" >> /home/vscode/.zshrc
# for root
echo "export SDKMAN_DIR=${SDKMAN_DIR}\nsource \${SDKMAN_DIR}/bin/sdkman-init.sh" >> /root/.bashrc
echo "export SDKMAN_DIR=${SDKMAN_DIR}\nsource \${SDKMAN_DIR}/bin/sdkman-init.sh" >> /root/.zshrc
```

with:

```bash
echo "export SDKMAN_DIR=${SDKMAN_DIR}\nsource \${SDKMAN_DIR}/bin/sdkman-init.sh" >> /etc/profile.d/sdkman.sh
```

Which has the benefits:

1. Works for all users (root and non-root)
2. Works for bash and zsh
3. Much safer than editing the `rc` directly

felipecrs:
Latest news about this:

[...]

PR #518, opened 2020-08-25
Update global dotfiles instead of user's for python and java

felipecrs:
Please note: 

If I start a container with my custom `dotfiles` WITHOUT this change, I'm not able to find the latest Python on my path because of the changes to the `bashrc` or `zshrc` were overridden because of my `dotfiles`.

Doing it in this way it works if I use a custom `dotfiles` or not.

Fixes partially #512.

### DIFF
--- a/script-library/java-debian.sh
+++ b/script-library/java-debian.sh
@@ -13,7 +13,7 @@ UPDATE_RC=${4:-"true"}
 
 set -e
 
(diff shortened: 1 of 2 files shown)


felipecrs:
Already fixed by https://github.com/microsoft/vscode-dev-containers/pull/529.

PR #522, opened 2020-08-27
Adds mongo shell and networks mongo to localhost

burkeholland:


Chuxel:
LGTM!

chrmarti:
@burkeholland @Chuxel The hostname might need an update too (I don't think `mongo` still resolves):
https://github.com/microsoft/vscode-dev-containers/blob/147610ce02c21d7c620b86374d6c2320da43aa18/containers/javascript-node-mongo/test-project/server.js#L13

Chuxel:
@chrmarti Yeah, you are right, and I fixed this in the next PR (#523) that updated the others as well along with readmes. Good catch tho!

PR #525, opened 2020-08-28
fixes #520:

### LINKED ISSUE #520: Remove redundant options in js-node-mongo server.js
Currently, `{useNewUrlParser: true}` is passed as an option in the starter project, which is redundant as the default value for the option is `true`. See [this](http://mongodb.github.io/node-mongodb-native/3.5/api/MongoClient.html)

https://github.com/microsoft/vscode-dev-containers/blob/fe59d117cd79cc99e129b7b58570986b0367d9d4/containers/javascript-node-mongo/test-project/server.js#L19

Also, consider adding `{useUnifiedTopology: true}` to the starter server.

If accepted as valid changes, I'd like to raise a PR for the same.

yashshah1:
* updated redundant options in js-node-mongo server.js

### DIFF
--- a/containers/javascript-node-mongo/test-project/server.js
+++ b/containers/javascript-node-mongo/test-project/server.js
@@ -16,7 +16,7 @@ const HOST = '0.0.0.0';
 
 (async function() {
 	// Use connect to mongo server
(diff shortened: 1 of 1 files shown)


Chuxel:
Thanks!

PR #529, opened 2020-08-30
Resolve pathing issues, move script inits out of user rc files

Chuxel:
There's something odd about the Oryx image that is causing default paths set in the Dockerfile to not appear which prevents things like PHP from working. This PR reworks where global paths are set to attempt to resolve this.  It also moves config file updates into /etc/bash.bashrc and /etc/zsh/zshrc per #512


Chuxel:
//cc: @jkeech @edgonmsft as FYI

PR #540, opened 2020-09-05
Move bash definition to Debian/Ubuntu

Chuxel:
Updates the bash definition to use a choice of Debian / Ubuntu to ensure it works in GitHub Codespaces - technically could set variant to "alpine" as well.

//cc: @AarynSmith given you contributed the definition


AarynSmith:
Lgtm

PR #553, opened 2020-09-12
Add basic sshd boostrapping script

Chuxel:


kingces95:
@Chuxel, hello! Are you also planning to add a `sshd-alpine.sh`?

I tried to spin up an sshd server on top of the alpine container, and after authenticating, got a 
> stderr> Illegal option -p

I could keep hacking at it if that type of work might be merged, or if you are already cooking it then, I'll wait for the drop.

Thanks!

script-library/sshd-debian.sh
```diff
@@ -46,23 +45,27 @@ if ! dpkg -s openssh-server openssh-client > /dev/null 2>&1; then
...
+# Generate password if new password set to the word "random"
 if [ "${NEW_PASSWORD}" = "random" ]; then
     NEW_PASSWORD="$(openssl rand -hex 16)"
     EMIT_PASSWORD="true"
 fi
 
+# If new password not set to skip, set it for the specified user
 if [ "${NEW_PASSWORD}" != "skip" ]; then
     echo "${USERNAME}:${NEW_PASSWORD}" | chpasswd
     if [ "${NEW_PASSWORD}" != "root" ]; then
```
kingces95, line 57: @Chuxel, Should this be `${USERNAME}` instead of `${NEW_PASSWORD}`?

Chuxel: Nope, if the new password is not set to skip, it pipes the username : password into chpasswd. See here for where the password would get set: https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/docs/sshd.md#ad-hoc-usage

Chuxel:
> @Chuxel, hello! Are you also planning to add a `sshd-alpine.sh`?
> 
> I tried to spin up an sshd server on top of the alpine container, and after authenticating, got a
> 
> > stderr> Illegal option -p
> 
> I could keep hacking at it if that type of work might be merged, or if you are already cooking it then, I'll wait for the drop.
> 
> Thanks!

The reason the script is `-debian` is because its for debian/ubuntu.  I could, though at the moment Codespaces doesn't support alpine does it? I'd love to fix that.

script-library/sshd-debian.sh
kingces95: Sorry, I mean on line 57, we compare `"${NEW_PASSWORD}" != "root"`, should that be `"${USERNAME}" != "root"`?

kingces95:
@Chuxel, actually, I was trying to use VSCode + Remote SSH plugin to connect to a sshd demon running in an alpine container. Using your script, it's, of course, trivial to make this scenario work with debian. So I'm wondering if there were plans to also make it work with [the alpine container](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/alpine).

PR #564, opened 2020-09-17
Use bash -c instead of piping in the script-library docs

felipecrs:
This way is simpler (at least for passing parameters) + safer. That's why Homebrew suggests this way (https://brew.sh/).


Chuxel:
LGTM!

Issue #561, opened 2020-09-16
Add SSH script to GH Codespaces Universal image


Chuxel:
We've got [a script](https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/sshd-debian.sh) to add a SSH server into containers for scenarios where users want to take advantage of something like SSHFS to gain complete access to the entire container contents.

We can add this to the "universal" image and look into using spinning up the server by default. At a minimum we could have the sever pre-installed and the script is then in the path start up the server when needed (it already can generate a random password for secure use).

Chuxel:
Resolved in #556.  Will update to port 22 when that option is available in Codespaces.

Issue #560, opened 2020-09-16
[Tracking] Misc script library tasks


Chuxel:
There are a series of improvements that need to be made to the library scripts. This issue is for tracking.

- [x] Add a SSH script (#553)
- [x] Add a light weight Linux desktop script (#554)
- [x] Fix `code` alias in Zsh and Fish (#557)
- [x] Add Oh My Bash! into the common script ([bash-it has no license](https://github.com/Bash-it/bash-it/issues/512), so can't use)
- [x] Default to non-root user (#552)
- [x] Be smarter about identifying a non-root user (look up UID 1000 if no user name specified)
- [ ] Switch to pyenv for Python script (#533)
- [ ] Adapt devcontainer.json and Dockerfile comments to cover using an interactive shell for nvm, sdkman, etc (#559)
- [x] Add documentation for scripts explaining usage #563 
- [ ] <strike>Switch ssh script to default to port 22 in codespaces-linux once this is possible</strike> (Not doing this since it would conflict with local ports anyway.)

PR #566, opened 2020-09-18
Add SSH, Oh My Bash! to codespaces-linux

Chuxel:
For #560, #561

This PR adds an SSH server to the codespaces-linux image. Instructions for use are in the [README](https://github.com/microsoft/vscode-dev-containers/tree/clantz/codespaces-ssh/containers/codespaces-linux#accessing-the-container-using-ssh-scp-or-sshfs).

It also adds Oh My Bash! for shell customization and a default theme. The theme will includes prompt with `gh:@<userid>` if `GITHUB_USER` is set and otherwise uses `whoami` to show the current active user. 

The themes are embedded in [common-debian.sh](https://github.com/microsoft/vscode-dev-containers/blob/clantz/codespaces-ssh/script-library/common-debian.sh) so we can easily update them as we finalize without requiring multiple script files to be copied around.

//cc: @jkeech

### DIFF
--- a/containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
+++ b/containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
@@ -16,7 +16,7 @@ NEW_PASSWORD=${4:-"skip"}
 set -e
 
 if [ "$(id -u)" -ne 0 ]; then
-    echo -e 'Script must be run a root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
     exit 1
 fi
 
--- a/containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
+++ b/containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+# Syntax: ./sshd-debian.sh [SSH Port (don't use 22)] [non-root user] [start sshd now flag] [new password for user]
+
+# Note: You can change your user's password with "sudo passwd $(whoami)" (or just "passwd" if running as root).
+
+SSHD_PORT=${1:-"2222"}
+USERNAME=${2:-"vscode"}
+START_SSHD=${3:-"false"}
+NEW_PASSWORD=${4:-"skip"}
+
+set -e
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo -e 'Script must be run a root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    exit 1
+fi
+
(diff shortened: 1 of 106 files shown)


containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
```diff
@@ -0,0 +1,96 @@
...
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+# Syntax: ./sshd-debian.sh [SSH Port (don't use 22)] [non-root user] [start sshd now flag] [new password for user]
+
+# Note: You can change your user's password with "sudo passwd $(whoami)" (or just "passwd" if running as root).
+
+SSHD_PORT=${1:-"2222"}
```
jkeech, line 11: We're working on freeing up port 22 on the host, so the default SSH port should work in Codespaces in a week or two, at which point we can probably move this back to the default since that will be simpler for folks to use with standard SSH config on the clients.

containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
```diff
@@ -0,0 +1,96 @@
...
+USERNAME=${2:-"vscode"}
+START_SSHD=${3:-"false"}
+NEW_PASSWORD=${4:-"skip"}
+
+set -e
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo -e 'Script must be run a root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
```
jkeech, line 19: Small typo
```suggestion
    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
```

containers/codespaces-linux/definition-manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-	"definitionVersion": "0.16.1",
+	"definitionVersion": "0.17.0",
```
jkeech, line 2: FYI, @dmgardiner25 is going to upgrade `gh` CLI to 1.0.0, so let's not ship this until the GH change is made as well.

jkeech (approved):

containers/codespaces-linux/definition-manifest.json
Chuxel: Actually, good news on that one. The script in kitchen sink should be automatically grabbing the latest release, so that should be done.  

I just noticed the old env var is getting passed in still, but its not set, which results in the latest coming down. I'll remove that bit.

image

containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
Chuxel: Ha - that typo is all over the place due to copy-paste.  Will find and replace. Good catch.

containers/codespaces-linux/definition-manifest.json
jkeech: Ah, cool, I missed that!

containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
Chuxel: Yeah, agreed.  We can PR that separately so this branch doesn't get too far out of date.

PR #569, opened 2020-09-21
Script documentation and comment tweaks

### LINKED ISSUE #563: Properly document the Dockerfile requirements for the scripts
For example, `node-debian.sh` seems to require:

```Dockerfile
ARG NODE_VERSION="lts/*"
ENV NVM_DIR=/usr/local/share/nvm
ENV NVM_SYMLINK_CURRENT=true \
  PATH=${NVM_DIR}/current/bin:${PATH}
```

In order to work properly. 

At least until we have https://github.com/microsoft/vscode-remote-release/issues/3593, otherwise the scripts would take care of it by themselves.

Chuxel:
Also adds "latest" mode into two scripts.

Fixes #563

### DIFF
--- a/containers/codespaces-linux/.devcontainer/base.Dockerfile
+++ b/containers/codespaces-linux/.devcontainer/base.Dockerfile
@@ -48,9 +48,9 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     # Clean up
     && rm -rf /tmp/scripts/ && apt-get autoremove -y && apt-get clean -y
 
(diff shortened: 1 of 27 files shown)


felipecrs:
@Chuxel, I saw that most of the scripts calls in the docs run `apt-get update` before, such as in node.md:

```Dockerfile
RUN apt-get update && bash /tmp/library-scripts/node-debian.sh "${NVM_DIR}"
```

But I believe the apt-get update is already being made in the script itself, so is it really needed? If not, I can open a separate Issue for that.

PR #575, opened 2020-09-23
Use code as shim instead of as an alias

felipecrs:
This is a better approach, since using alias does not make it work when `code` is called by CLIs, such as `git` or `gh`.

PS: this does not take precedence over the normal `code` binary, so it's harmless when it's not useful. See:

```sh-session
$ which -a code
/home/vscode/.vscode-server/bin/58bb7b2331731bf72587010e943852e13e6fd3cf/bin/code
/usr/local/bin/code
```

References https://github.com/MicrosoftDocs/vscodespaces/issues/809.


Chuxel:
LGTM! Thanks!

PR #578, opened 2020-09-25
adds devcontainer for Azure Static Web Apps

videlalvaro:
This adds a new devcontainer that does the setup for the required extensions & ports for running Azure Static Web Apps with an Azure Functions API backend. 

It adds a sample `test-project` to test the container.

The extra extension is Live Server. The test project has sample config for that extension


Chuxel:
Thanks for the contribution!

However, this seems close enough to the existing functions definition that I'm not sure it needs to be separate.  @anthonychu - Would it be better to just add "Live Server" into functions?  We could also add Node.js as an optional install into the .NET and Python ones.

anthonychu:
@videlalvaro Thanks!

@Chuxel That's a good point. I had chatted with @videlalvaro about adding a Static Web Apps devcontainer. What I sort of had in mind was a container definition with these tools:

- Live Server (it would be replaced by an official Static Web Apps CLI/emulator in a couple of months, hopefully)
- The Static Web Apps [VS Code Extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azurestaticwebapps)
- Support for functions in Node, .NET, and Python (thinking use the Node functions container and add Python).

Do you think that would be different enough to warrant a separate container definition? Another reason I was thinking a separate one is for discoverability. Open to suggestions though.

The flow for a dev I was thinking of would be something like this:

1. Open repo with a React app
1. Add Azure Static Web Apps devcontainer
1. Open in Codespaces / VS Code remote for containers
1. Use the extension to create GitHub Actions workflow and deploy
1. Add functions in a language of their choice without having to install more tools

videlalvaro:
@anthonychu AFAIK at the moment the different functions language backends are in separate devcontainers, how would that work @Chuxel ?

Chuxel:
@anthonychu @videlalvaro If it was me, I would start with ` mcr.microsoft.com/azure-functions/python:3.0-python3.8-core-tools` and add Node.js since it's much easier to install than Python.  Installing Python on Linux == compiling which is SLLLLLLOOOOOWWWW.  That would give you .NET Core 3.1, Python 3.8, and Node.js.  

For Node.js, you could use `nvm` via the [node.js in this repo](https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/docs/node.md)) or use one of the [Nodesource installer scripts](https://github.com/nodesource/distributions#debian-and-ubuntu-based-distributions).   

I believe the tools are .NET Core 3.1 based in all scenarios, right @anthonychu ?

Also @burkeholland - FYI since you were asking about Functions + Node the other day.

(FWIW - This is the upstream dockerfile for that image https://github.com/Azure/azure-functions-docker/blob/master/host/3.0/buster/amd64/python/python38/python38-core-tools.Dockerfile)

videlalvaro:
@Chuxel can we push new containers to that MS registry? I assume you mean to create a new container that includes these technologies right?

anthonychu:
I think @Chuxel's suggestion is use `FROM mcr.microsoft.com/azure-functions/python:3.0-python3.8-core-tools` in `.devcontainer/Dockerfile`, then add the instructions to install Node. There's no need to publish a new image to MCR for Static Web Apps (yet). We could decide to do that in the future.

I think if we can make that change and add the Static Web Apps extension, this should be in pretty good shape.

Also note that SWA currently supports Node.js 12, Python 3.8, and .NET 3.1. This should simplify things a bit for now as there's currently no need to use a variant arg.

videlalvaro:
I've updated the PR @Chuxel 

Changed the Dockerfile, and added the missing extensions to the devcontainer

videlalvaro:
bump @Chuxel

Chuxel:
LGTM! @anthonychu ?

PR #579, opened 2020-09-26
Fix schellcheck warning in Debian desktop script

vadi2:
See https://github.com/koalaman/shellcheck/wiki/SC2054 for background. Seems like a reasonable complaint.


Chuxel:
Thanks!  Had this bug elsewhere but missed one!

PR #584, opened 2020-09-27
Clarify terminology for Debian Lite

vadi2:
I was honestly not sure what a "web based desktop viewer" exactly meant given that instructions are also talking about VNC in the same context, but it did turn out to be just a normal web browser.


Chuxel:
Thanks!

PR #585, opened 2020-09-27
config.js file info

LavaAdventures:
File path and file name


Chuxel:
Thanks for the PR.  However, what does doing this enable?  Why just this file?

LavaAdventures:
It shows what file is being edited.

Chuxel:
@LavaAdventures There are many other ways to find that other than adding it to every file in this repository. Some files do not have header contents either. When would this be useful?

Kir-Antipov:
@Chuxel, don't waste your time, it's Shitoberfest's spammer. Just close this PR.

PR #586, opened 2020-09-27
async.js file name and file path info

LavaAdventures:
*  File: async.js
 *  File Path: vscode-dev-containers/build/src/utils/async.js


Chuxel:
Thanks for the PR. Please consolidate these changes into a single PR and explain why it is useful as asked in #585

PR #587, opened 2020-09-27
File path and file name info

LavaAdventures:
*  File: push.js
 *  File Path: vscode-dev-containers/build/src/push.js


Chuxel:
Thanks for the PR. Please consolidate these changes into a single PR and explain why it is useful as asked in #585

PR #594, opened 2020-10-02
Add Azurite Node devcontainer

AlexanderYW:
This is a new devcontainer that setup recommended extensions and ports for running Azurite locally in a devcontainer

It also contains two tests for `blob` and `queue`

### DIFF
--- a/containers/javascript-node-azurite/.devcontainer/docker-compose.yml
+++ b/containers/javascript-node-azurite/.devcontainer/docker-compose.yml
@@ -1,0 +1,40 @@
+version: '3'
+
+services:
+  app:
+    build: 
+      context: .
+      dockerfile: Dockerfile
+      args:
+        # [Choice] Node.js version: 14, 12, 10
+        VARIANT: 14
+        # On Linux, you may need to update USER_UID and USER_GID below if not your local UID is not 1000.
+        USER_UID: 1000
+        USER_GID: 1000
+
+    volumes:
+      - ..:/workspace:cached
+
+    # Overrides default command so things don't shut down after the process ends.
+    command: sleep infinity
+
+    # Runs app on the same network as the database container, allows "forwardPorts" in devcontainer.json function.
+    network_mode: service:storage
+
+    # Uncomment the next line to use a non-root user for all processes.
+    # user: node
+
+    # Use "forwardPorts" in **devcontainer.json** to forward an app port locally. 
+    # (Adding the "ports" property to this file will not forward from a Codespace.)
+
+  storage:
+    image: mcr.microsoft.com/azure-storage/azurite:latest
+    restart: unless-stopped
+    volumes:
+      - storage-data:/data/db
+
+    # Add "forwardPorts": ["10000", "10001"] to **devcontainer.json** to forward Azurite locally.
+    # (Adding the "ports" property to this file will not forward from a Codespace.)
+
+volumes:
+  storage-data:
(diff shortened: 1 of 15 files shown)


containers/javascript-node-azurite/.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+    # Runs app on the same network as the database container, allows "forwardPorts" in devcontainer.json function.
+    network_mode: service:storage
+
+    # Uncomment the next line to use a non-root user for all processes.
+    # user: node
+
+    # Use "forwardPorts" in **devcontainer.json** to forward an app port locally. 
+    # (Adding the "ports" property to this file will not forward from a Codespace.)
+
+  storage:
```
ItalyPaleAle, line 30: Would it make sense to add `container_name: storage` (or `azurite`) so the other services can connect to Azurite via `http://storage:10000` directly?

containers/javascript-node-azurite/.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+    image: mcr.microsoft.com/azure-storage/azurite:latest
+    restart: unless-stopped
+    volumes:
+      - storage-data:/data/db
```
ItalyPaleAle, line 34: Should the entire `/data` folder be mounted in the container?  As per docs: https://github.com/Azure/Azurite#run-azurite-v3-docker-image-with-customized-persisted-data-location

ItalyPaleAle (commented):
I've used Azurite in a dev container before and I think this is a good example, but I've proposed a few tweaks (I like how this should work through Codespaces too!)

containers/javascript-node-azurite/.devcontainer/docker-compose.yml
AlexanderYW: Thanks for pointing that out, I will correct that soon :)

containers/javascript-node-azurite/.devcontainer/docker-compose.yml
AlexanderYW: Sure, I can add that :)

AlexanderYW:
> I've used Azurite in a dev container before and I think this is a good example, but I've proposed a few tweaks (I like how this should work through Codespaces too!)

To be honest with you, I tried reaching out to a few friends if they had gotten early access to codespaces, so they could test it for me since I haven't, but no one had, if you have, I would love if you could test it out, but based on my assumptions it should 🙃

containers/javascript-node-azurite/README.md
```diff
@@ -0,0 +1,64 @@
+# Node.js & Azurite
```
2percentsilk, line 1:
```suggestion
# Node.js & Azurite (Community)
```

containers/javascript-node-azurite/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Summary
+
+Everything you need to get started using Node with Azurite.
+
+| Metadata | Value |  
+|----------|-------|
+| *Contributors* | [Alexander Wennerstrøm](https://github.com/AlexanderYW)   |
+| *Definition type* | Docker Compose |
```
2percentsilk, line 10: We've added a new piece of metadata: Categories! 

```suggestion
| *Categories* | Community, Azure
| *Definition type* | Docker Compose |
```

2percentsilk (changes requested):
Apologies for the delay here. This addition looks great. I suggested two small changes since we've updated how we title and tag the definitions so you'll need to please accept those.

As the final piece, I want to confirm that you're willing to be the maintainer for this definition - generally this means we'll loop you in for comment/approval for any other PRs that the community may make against this definition. Once you confirm this, I'll be happy to merge!

AlexanderYW:
Sure, I can confirm that I'm willing to maintain this 🙂

PR #595, opened 2020-10-03
Add: Chef Workstation container

### LINKED ISSUE #256: add chef-workstation
Please add a [Chef Workstation](https://docs.chef.io/about_workstation/) Dev Container.

They have an [official docker image](https://hub.docker.com/r/chef/chefworkstation) built on [this`Dockerfile`](https://github.com/chef/chef-workstation/blob/master/Dockerfile).

ehanlon:
Resolves #256


ehanlon:
@Chuxel is there a reviewer available for this one?

Chuxel:
Apologies for the delays here. This one got away from us.  Generally looks good!  Most definitions in this repository do include a non-root user and the other contents we have in the "common-debian.sh" script but if you don't think commands like "lsof" or "htop" are needed, that's not strictly required.  I did see git was there which is the big one we always want to be sure is present.

I assume you're good with maintaining this definition going forward? We'll @ you in as new issues appear.

Chuxel (approved):

ehanlon:
> Apologies for the delays here. This one got away from us. Generally looks good! Most definitions in this repository do include a non-root user and the other contents we have in the "common-debian.sh" script but if you don't think commands like "lsof" or "htop" are needed, that's not strictly required. I did see git was there which is the big one we always want to be sure is present.
> 
> I assume you're good with maintaining this definition going forward? We'll @ you in as new issues appear.

No worries. I'd be happy to maintain this.

Chuxel:
Sounds good.  There's one tweak since this was submitted where we no longer need to set the default terminal because of VS Code improvements but I can do that quickly post-merge.

PR #612, opened 2020-10-15
[Codespaces] Installing libsecret-1-dev by default

lostintangent:
This PR simply installs the `libsecret-1-dev` package in the universal image, which ensures that common app scenarios work without any further modification (e.g. Node apps that depend on `keytar`). Without this, developers that are using macOS might be surprised to see that their environment no longer works, because they aren't used to the extra requirements for Linux.


Chuxel:
LGTM - I kept this out of the core images since it adds 80MB to something like the 382MB base image. In this case, I don't think it will have a material effect since I'd expect most dependencies like that to already be in this giant image.

PR #613, opened 2020-10-15
chore: changing "restart" to "start" of sshd server

rajkumar42:
pid file of the sshd server gets saved in the container. When it is restarted on a different VM, the pid is different and that process gets killed.
This results in random process kills and failure of about 10% of the resumes.
Fix is to change "restart" to "start" so that no process gets killed as container doesn't have any processes at that time.

### DIFF
--- a/containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
+++ b/containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
@@ -86,9 +86,9 @@ tee /usr/local/share/ssh-init.sh > /dev/null \
 set -e 
 
 if [ "\$(id -u)" -ne 0 ]; then
-    sudo /etc/init.d/ssh restart
+    sudo /etc/init.d/ssh start
 else
-    /etc/init.d/ssh restart
+    /etc/init.d/ssh start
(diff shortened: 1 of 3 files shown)


rajkumar42:
@Chuxel , there is also 
`vscode-dev-containers\script-library\sshd-debian.sh` 

i'm keeping restart as is. 

```
if [ "\$(id -u)" -ne 0 ]; then
    sudo /etc/init.d/ssh restart
else
    /etc/init.d/ssh restart
fi
```

Do you think this would have the same use-case as codespaces? And needs changing?

Chuxel:
@rajkumar42 Yes I was just typing that actually. 😀  To avoid this getting overwritten, make this same change here: https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/sshd-debian.sh

There's a process that mass-updates changes that are made in the script-library folder to reduce maintenence.

You can then bump up the last digit of the version here and we can cut a release: https://github.com/microsoft/vscode-dev-containers/blob/master/containers/codespaces-linux/definition-manifest.json

//cc: @chrmarti as FYI

Chuxel:
LGTM! Merging.

PR #614, opened 2020-10-18
Add trailing comma after extensions array

nguyenyou:
This would be convenience since we don't have to add a comma after the `extensions` array whenever we need to use an option below this `entensions` option. 

We already did that with the following devcontainer:

- containers/azure-functions-node/.devcontainer/devcontainer.json
- containers/azure-functions-python-3/.devcontainer/devcontainer.json
- containers/bash/.devcontainer/devcontainer.json
- containers/deno/.devcontainer/devcontainer.json
- containers/hugo/.devcontainer/devcontainer.json
- containers/jekyll/.devcontainer/devcontainer.json
- containers/php/.devcontainer/devcontainer.json
- containers/php-mariadb/.devcontainer/devcontainer.json
- containers/vue/.devcontainer/devcontainer.json

### DIFF
--- a/container-templates/docker-compose/.devcontainer/devcontainer.json
+++ b/container-templates/docker-compose/.devcontainer/devcontainer.json
@@ -18,7 +18,7 @@
 	},
 
 	// Add the IDs of extensions you want installed when the container is created.
-	"extensions": []
+	"extensions": [],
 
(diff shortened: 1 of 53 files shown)


Chuxel:
The problem is that the schema will fail validation with a hanging comma, so I don't think we can do this unfortunatley.

nguyenyou:
Ok. I'll close this pr.

PR #616, opened 2020-10-22
Fix code binary when it also exists in dotfiles

felipecrs:
Now it only looks for `code` down in the PATH, preventing an infinite loop.


felipecrs:
@Chuxel could please take a look at this PR? I'm having trouble in all Codespaces due to this. =/

Chuxel:
LGTM!  Sorry for the delay!

PR #624, opened 2020-10-27
add if statement on installation pyutils

peach-lasagna:
because i don't need them


Chuxel:
This does not build unfortunately. If you are using the pre-built image, these are already in the image so you don't need to worry about it.  Are you needing to use base.Dockerfile for some reason?

Chuxel:
Closing due to lack of response.

Issue #489, opened 2020-08-13
.NETCore image with Linux SQL Server


warrenbuckley:
Is it possible to have a pre configured example/image for .NETCore & a Linux SQL Server. So I can add the .devcontainers folder into my project nice & quickly please

Chuxel:
Thanks for the request! Certainly possible to do, and happy to take a look at a PR that does it if you want to contribute one!

warrenbuckley:
I wish I knew docker more to understand what I need to do to contribute one.

warrenbuckley:
As VS CodeSpaces with a Windows host seems to have SQL Server

https://docs.microsoft.com/en-us/visualstudio/codespaces/reference/configuring-windows#microsoft-sql-server

Is it possible for GitHub CodeSpaces to support Windows hosts too that would allow the same scenario @Chuxel

warrenbuckley:
I am closing this as I got the email from Microsoft/GitHub that VS CodeSpaces is being moved into GitHub only and that the Windows environment will be coming too at some point.

dzsquared:
Hi @warrenbuckley - I created a devcontainer defintion that covers .NET Core and SQL Server in the linked PR #632.  If you you have a chance to try it out I would love your feedback.  thanks!

PR #632, opened 2020-11-02
dotnetcore and mssql

dzsquared:
This PR adds a new dev-container definition for .NET Core with MS SQL, including:
- secondary MS SQL container
- updated .NET Core example with DB connectivity
- entrypoint script for SQL with optional dacpac deployment

Addresses: https://github.com/microsoft/vscode-dev-containers/issues/489


Chuxel:
Thanks!  Will try to look at this in detail soon soon!  One comment is that you probably don't need the "base.Dockerfile" or the `definition-manifest.json`. You should be able to directly reference the `Dockerfile` which will take advantage of the already built image.

dzsquared:
> Thanks! Will try to look at this in detail soon soon! One comment is that you probably don't need the "base.Dockerfile" or the `definition-manifest.json`. You should be able to directly reference the `Dockerfile` which will take advantage of the already built image.

thanks! I checked again and it wasn't using the `base.Dockerfile` or `definition-manifest.json`, so they've been removed.

containers/dotnetcore-mssql/.devcontainer/mssql/setup-db.sh
```diff
@@ -0,0 +1,36 @@
+sleep 60s
```
Chuxel, line 1: Could this setup-db.sh script be invoked as an option "postCreateCommand" script from devcontainer.json instead of an entrypoint? That way it doesn't need to be baked into the image which should speed up start time.

containers/dotnetcore-mssql/.devcontainer/mssql/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM mcr.microsoft.com/mssql/server:2019-latest
+
+USER root
+
+COPY setup.sql *.dacpac ./
```
Chuxel, line 5: The challenge here is this will all get backed into the image - as I mentioned in setup-db.sh, could this be done via postCreateCommand instead?

containers/dotnetcore-mssql/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,61 @@
...
+	"dockerComposeFile": "docker-compose.yml",
+	"service": "app",
+	"workspaceFolder": "/workspace",
+
+	// Set *default* container specific settings.json values on container create.
+	"settings": {
+		"terminal.integrated.shell.linux": "/bin/bash",
+		"mssql.connections": [
+			{
+				"server": "localhost,61433",
```
Chuxel, line 12: The correct port here is actually 1433. Did you do this due to GitHub Codespaces?  If so, that's because of a bug there.

containers/dotnetcore-mssql/.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+
+  db:
+    build:
+      context: ./mssql
+      dockerfile: Dockerfile
+    environment:
+      SA_PASSWORD: P@ssw0rd
+      ACCEPT_EULA: Y
+    ports:
+      - "61433:1433"
```
Chuxel, line 42: This will expost port 61443 to the host, and unless you are using host networking in the other container, you won't be able to connect to it. 

There is a bug in GitHub Codespaces right now that switches the main container to host networking without modifying the database.  They're working on a fix.  Did you happen to test this there?

Chuxel (commented):
Made a few comments below.

containers/dotnetcore-mssql/.devcontainer/docker-compose.yml
dzsquared: You are correct, this was my workaround to get networking working in GH codespaces.  I had assumed there was a block on port 1433 for security reasons in GH codespaces and didn't consider it being a bug.
I'll update both to network_mode: host and remove the change in the port forwarding (here and in devcontainer.json)

containers/dotnetcore-mssql/.devcontainer/mssql/setup-db.sh
dzsquared: I've updated the setup process to a postCreateCommand script, thanks!

containers/dotnetcore-mssql/.devcontainer/mssql/postCreateCommand.sh
```diff
@@ -0,0 +1,80 @@
...
+    if [ $f == $sqlpath/*".sql" ]
+    then
+        sqlfiles="true"
+        echo "found sql file $f"
+    fi
+done
+
+if [ $sqlfiles == "true" ]
+then
+    echo "installing mssql-tools"
```
Chuxel, line 42: This part could be in the Dockerfile for the main container rather than the SQL one.

containers/dotnetcore-mssql/.devcontainer/mssql/postCreateCommand.sh
```diff
@@ -0,0 +1,80 @@
...
+            echo "executing $f"
+            /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P $SApassword -d master -i $f
+        fi
+    done
+fi
+
+if [ $dacpac == "true" ] 
+then
+    echo "installing sqlpackage"
+    apt-get -y install wget
```
Chuxel, line 62: Ditto here - this part can be in the main container.  The one we would want to avoid building is the mssql container.

containers/dotnetcore-mssql/.devcontainer/mssql/postCreateCommand.sh
```diff
@@ -0,0 +1,80 @@
...
+        sqlfiles="true"
+        echo "found sql file $f"
+    fi
+done
+
+if [ $sqlfiles == "true" ]
+then
+    echo "installing mssql-tools"
+    curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
+    curl https://packages.microsoft.com/config/ubuntu/18.04/prod.list | tee /etc/apt/sources.list.d/msprod.list
```
Chuxel, line 44: The .NET containers are about to update to 20.04, so it would be good not to hard code the version.  You can avoid this problem by using the following instead:

```bash
DISTRO=$(lsb_release -is | tr '[:upper:]' '[:lower:]')
CODENAME=$(lsb_release -cs)
echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-${DISTRO}-${CODENAME}-prod ${CODENAME} main" > /etc/apt/sources.list.d/microsoft.list
```

containers/dotnetcore-mssql/.devcontainer/library-scripts/node-debian.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
```
Chuxel, line 1: You actually don't need this file since you aren't building `base.Dockerfile`.  You just will want `library-scripts\azcli-debian.sh`

containers/dotnetcore-mssql/.devcontainer/library-scripts/common-debian.sh
```diff
@@ -0,0 +1,327 @@
+#!/usr/bin/env bash
```
Chuxel, line 1: You actually don't need this file either since you aren't building `base.Dockerfile`.  You just will want `library-scripts\azcli-debian.sh`

containers/dotnetcore-mssql/.devcontainer/mssql/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM mcr.microsoft.com/mssql/server:2019-latest
```
Chuxel, line 1: Is this Dockerfile needed anymore? You could just reference the MSSQL image in docker-compose.yml to keep things simple.

Chuxel (commented):
Looking pretty good!  A few more comments.  I was thinking that the script would be split - any installation could happen in the main container since you'd likely want those tools avalable anyway. Then anything that depends on the source tree like running the .sql file could be in postCreateCommand.

Make sense? 

A few other comments below.

containers/dotnetcore-mssql/.devcontainer/mssql/postCreateCommand.sh
dzsquared: Splitting this between installation and execution makes sense.  Ideally we'd have a sqltools container image for simplicity, but we're not there yet.

containers/dotnetcore-mssql/.devcontainer/mssql/postCreateCommand.sh
dzsquared: thanks for the code sample - bash is not my strength

Chuxel:
Okay - LGTM. Will make some tweaks post-merge to line it up with the latest .NET updates.

PR #636, opened 2020-11-04
Modified node-postgres docker-compose for GH codespaces

MichaelJolley:
The node-postgres docker-compose works as expected if using the Remote Development - Containers extension in VS Code. However, when using this docker-compose in GitHub codespaces the networking is not created correctly and the node container cannot access the postgres container. This change allows the `5432` port to be shared with the host so it works as intended in GitHub codespaces. It will not impact users working locally.


Chuxel:
@MichaelJolley This is actually due to a bug in Codespaces that the I know the team is working on.

Chuxel:
Closing this for now since I'm hoping this is resolved soon! I also pinned the workaround in #637 to the top of the issues.

Issue #637, opened 2020-11-06
Docker Compose containers cannot communicate in GitHub Codespaces due to host networking being applied to only one container


Chuxel:
For Docker Compose definitions, there is an upstream bug in GitHub Codespaces that forces the main container into `network_mode: host` without also changing the related containers to the same networking mode.

The net effect is that the the different Compose managed containers can't communicate.  The GitHub Codespaces team is working on a fix.

In the mean time you can work around it by adding `network_mode: host` to each service. This should work both locally and in Codespaces.

e.g. For [Node + MongoDB](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/javascript-node-mongo):

```yaml
version: '3'

services:
  app:
    # *** UPDATED - used to be network_mode: service:db****
    network_mode: host

    build: 
      context: .
      dockerfile: Dockerfile
      args:
        # [Choice] Node.js version: 14, 12, 10
        VARIANT: 14
        # On Linux, you may need to update USER_UID and USER_GID below if not your local UID is not 1000.
        USER_UID: 1000
        USER_GID: 1000

    volumes:
      - ..:/workspace:cached

    # Overrides default command so things don't shut down after the process ends.
    command: sleep infinity


    # Uncomment the next line to use a non-root user for all processes.
    # user: node

    # Use "forwardPorts" in **devcontainer.json** to forward an app port locally. 
    # (Adding the "ports" property to this file will not forward from a Codespace.)

  db:
    # ***ADDED****
    network_mode: host

    image: mongo:latest
    restart: unless-stopped
    volumes:
      - mongodb-data:/data/db

    # Uncomment to change startup options
    # environment:
    #  MONGO_INITDB_ROOT_USERNAME: root
    #  MONGO_INITDB_ROOT_PASSWORD: example
    #  MONGO_INITDB_DATABASE: your-database-here

    # Add "forwardPorts": ["27017"] to **devcontainer.json** to forward MongoDB locally.
    # (Adding the "ports" property to this file will not forward from a Codespace.)

volumes:
  mongodb-data:
```

This affects:

Chuxel:
This should be resolved! 🎉🎉🎉

vermegi:
docker-compose file should be version 3.7. 
Creation in GH codespaces currently fails since services.app.init is not supported in version 3.5 of docker compose.

sethkimmel3:
Hey @Chuxel  - are you sure this is fixed? I'm running into the exact issue you seem to be describing.

PR #638, opened 2020-11-06
Update Dapr-related Dev Container Definitions

### LINKED ISSUE #620: Verify dapr definitions still working as expected, use pre-built images / refresh if needed
There are two Dapr definitions that have not been updated in a while:

- [Dapr Node 12](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/dapr-typescript-node-12)
- [Dapr .NET 3.1](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/dapr-dotnetcore-3.1)

These definitions do not use pre-built images and therefore can have longer build times that could be cut down by using `mcr.microsoft.com/vscode/devcontainers/javascript-node:12` and `mcr.microsoft.com/vscode/devcontainers/dotnetcore:3.1` respectively. There were also some initial issues with Codespaces that should be resolved, but it would be good to verify.

With that in mind, it's probably worth looking through these and make sure they're still in good shape.

/cc: @philliphoff who contributed the definitions

philliphoff:
Updates the Dapr Dev Container definitions for .NET Core and Node.js. Refactors the Dockerfiles to simplify creation of version variants as well as adopts the new "library script" scheme for installing Docker.  Verified that the definitions work with the latest Dapr version (0.11), including use of a workaround for a [Dapr issue](https://github.com/dapr/cli/issues/418) with instantiation of a Redis host.

Resolves #620.

> NOTE: In testing with GitHub CodeSpaces, I observe that the Docker network specified for the "app" service is not respected, which means the definition cannot be used as-is.  (While it can be modified to work, it means that the same definition in a repo cannot be shared between a CodeSpace and a local Dev Container.)
@Chuxel Any idea about this?


Chuxel:
> NOTE: In testing with GitHub CodeSpaces, I observe that the Docker network specified for the "app" service is not respected, which means the definition cannot be used as-is. (While it can be modified to work, it means that the same definition in a repo cannot be shared between a CodeSpace and a local Dev Container.)
> @Chuxel Any idea about this?

@philliphoff Sadly yes, this is #637 (though it's a Codespaces bug I just added here). It's setting `network_mode: host` only to the referenced container instead of all of them. This is a bug and the plan is to not do this anymore.

Will take a look with that in mind!

Chuxel:
LGTM! Once #637 is fixed, this should work in Codespaces too!

Issue #643, opened 2020-11-10
azure-terraform container installs node when INSTALL_NODE set to false.


aymond:
- VSCode Version: 1.51.0
- Local OS Version: WIndows 1903
- Name of Dev Container Definition with Issue: azure-terraform

Steps to Reproduce:

1. Clone repository
2. set ARG INSTALL_NODE="false"
3. `docker build .`
4. `node -v`

```
RUN if [ "${INSTALL_AZURE_CLI}" = "true" ]; then bash /tmp/library-scripts/azcli-debian.sh; fi \
    && if [ "${INSTALL_AZURE_CLI}" = "true" ]; then bash /tmp/library-scripts/node-debian.sh "${NVM_DIR}" "${NODE_VERSION}" "${USERNAME}"; fi \
    && if [ "${INSTALL_DOCKER}" = "true" ]; then \
        bash /tmp/library-scripts/docker-debian.sh "true" "/var/run/docker-host.sock" "/var/run/docker.sock" "${USERNAME}"; \
    else \
        echo '#!/bin/bash\n"$@"' > /usr/local/share/docker-init.sh && chmod +x /usr/local/share/docker-init.sh; \
    fi \
    && rm -rf /var/lib/apt/lists/*
```

Chuxel:
Resolved with #644! Thanks much!

PR #644, opened 2020-11-10
Replace INSTALL_AZURE_CLI with INSTALL_NODE

aymond:
This PR resolves issue https://github.com/microsoft/vscode-dev-containers/issues/643


Chuxel:
Ack!  Good catch.  Thanks for the PR!

PR #649, opened 2020-11-12
Fix paths in .npmignore for Julia

davidanthoff:
I believe the current paths in there are the reason while files are copied into user repositories that shouldn't really be copied.


Chuxel:
Yep! That would do it.  Thanks for the PR!

PR #656, opened 2020-11-14
Tweaks to Azure Bicep definitions

Chuxel:
@cmendible I noticed a couple things after I merged that we could probably tweak.  See what you think.


cmendible (approved):

cmendible:
LGTM

Issue #447, opened 2020-07-27
Jekyll - can we run postCreateCommand: bundle install by default?


jongio:
Right now, users have to run `bundle install` manually.

Can we uncomment this line so it runs automatically?

https://github.com/microsoft/vscode-dev-containers/blob/v0.128.0/containers/jekyll/.devcontainer/devcontainer.json#L17

This is what I have:
`"postCreateCommand": "bundle install",`

Not sure you need "input.projectName"

@cmendible

cmendible:
Running bundle install without a Gemfile will result in: **Could not locate Gemfile** on startup which would be the case if you are starting a new project or you are still learning the ways of jekyll without plugins.

I'll remove the **cd ${input:projectName} &&** but leave the comments. Thoughts?

jongio:
Okay, thanks.    I was also thinking...what if we also add the serve command, so when the container loads the site is all ready to be hit.  I like that.

Chuxel:
@cmendible Let me know your thoughts here - I'll wait on merging to avoid another update if you think this makes sense.

cmendible:
@jongio can you confirm that the PR can be merged? I know it's working for your usecase 😉

Chuxel:
Merged for now. Feel free to reopen if more discussion is required.

ItalyPaleAle:
I proposed some changes in #661 that would automatically run bundle install if a Gemfile is present. @jongio can you take a look too and let me know your thoughts?

PR #661, opened 2020-11-18
Updates to Jekyll container

ItalyPaleAle:
This PR contains a few updates to the Jekyll container

1. If there's a Gemfile, do not install Jekyll by default. Doing that would automatically install the latest version, which may or may not be what the user wants. Instead, run `bundle install` in the post-create step. If there's no Gemfile, then installs jekyll in the post-create step. This should be a compromise for #447 
2. Adding live reload server port to the list of ports (for `jekyll serve -l`)
3. Updated default time zone to a more universal Etc/UTC

### DIFF
--- a/containers/jekyll/.devcontainer/Dockerfile
+++ b/containers/jekyll/.devcontainer/Dockerfile
@@ -3,13 +3,13 @@ FROM mcr.microsoft.com/vscode/devcontainers/ruby:2
 # ENV Variables required by Jekyll
 ENV LANG=en_US.UTF-8 \
     LANGUAGE=en_US:en \
-    TZ=America/Chicago \
+    TZ=Etc/UTC \
     LC_ALL=en_US.UTF-8 \
     LANG=en_US.UTF-8 \
     LANGUAGE=en_US
 
-    # Install jekyll
-RUN gem install bundler jekyll
+# Install bundler
+RUN gem install bundler
 
 # [Option] Install Node.js
 ARG NODE_VERSION="lts/*"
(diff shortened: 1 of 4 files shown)


Chuxel:
@cmendible could you review?

containers/jekyll/.devcontainer/Dockerfile
```diff
@@ -3,13 +3,13 @@ FROM mcr.microsoft.com/vscode/devcontainers/ruby:2
 # ENV Variables required by Jekyll
 ENV LANG=en_US.UTF-8 \
     LANGUAGE=en_US:en \
-    TZ=America/Chicago \
+    TZ=Etc/UTC \
```
cmendible, line 6: great!

containers/jekyll/.devcontainer/post-create.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# If there's a Gemfile, then run `bundle install`
+# It's assumed that the Gemfile will install Jekyll too
+if [ -f Gemfile ]; then
```
cmendible, line 5: I  feel comfortable with the way this is implemented. Can you please add a description of this behavior to the documentation?

cmendible (changes requested):
Please update the README.md explaining the new behavior.

ItalyPaleAle:
Thanks @cmendible for the review! Good callout about the docs. I updated the README explaining the behavior.

cmendible:
LGTM

PR #664, opened 2020-11-19
Changes to the Hugo container

ItalyPaleAle:
1. Re-basing this to the Node.js container because that can be very useful to build themes (note: this container does not include Go just like the old container didn't - I don't believe this is necessary when working with Hugo).
2. Removed a command that was patching the binary. That is not necessary when not running on Alpine (the container runs on Debian)
3. The downloaded binary is always called `hugo` (even in extended mode), so changed `mv hugo*` to `mv hugo` to be on the safer side
4. Disabled problem matchers in tasks.json


Chuxel:
@AarynSmith Can you review?

AarynSmith:
LGTM. Thanks for the update @ItalyPaleAle!

Chuxel:
Merged!

Issue #654, opened 2020-11-13
Update docker-debian.sh to use Moby CLI by default


Chuxel:
In order to allow the the docker CLI to be redistributed in images, we can use the Azure moby-cli package to get the command instead of Docker sources. The script can support installing from either source based on user perferece.

Chuxel:
This is complete!

Issue #623, opened 2020-10-27
Update codespaces-linux definition to use Ubuntu Focal / 20.04


Chuxel:
Currently the codespaces-linux image is based on Oryx which is in turn based on Debian 9. Debian 9 is pretty old at this point and there's been interest in PPA use when developers are not building their own custom container images. As a result, we should work with the Oryx team to update the image to Ubuntu 20.04.

C++ projects in particular suffer from the age of this distro.

This updated image will have Java and Ruby available under `/opt` as well so we do not have to install these manually. However, SDKMAN, rvm, and rbenv (#582) can still be installed to allow the use of them to install additional versions (or other tools for SDKMAN) not in the image itself. Yarn is also at the latest version as are other tools.

- [x] Adapt existing Dockerfile to Ubuntu 20.04
- [x] Bump up major version number for image (technically "breaking" update due to older SDKs not being supported)
- [x] Update PowerShell to 7.1 - pending release to packages.microsoft.com
- [x] Known bugs fixed in image fixed
    - [x] Python compiled against older version of OS
    - [x] System python is in path instead of updated one
- [x] Add .NET 5 - Oryx update, could decouple if causes a delay

Chuxel:
Early scouting of `mcr.microsoft.com/oryx/build:vso-focal-20201110.1` uncovered a few things of note to work through:

1. PowerShell has not released an official package for Ubuntu 20.04 yet nor does 7.0 technically support it. I used the .NET Core install version, but 7.1 (which should be released this week) will have 20.04 support. We may want to hold off until this is officially available.
    - Worked around this tactically be using the .NET Core version.
2. The compiled version of Python in the /opt/python folders appear to be built for a previous version of Debian or Ubuntu and  as a result pip fails when installing packages, and I'd assume there are other issues. This will need to be resolved before we can adopt. [See this SO post on what is happening.](https://stackoverflow.com/questions/61875869/ubuntu-20-04-upgrade-python-missing-libffi-so-6)
    - Worked around this tactically by using the system version to install tools.
3. The system version of python seems to be getting used by /usr/local/pip3 which is a bit counterintuitive. This is in the path first so its not initially obvious that /opt/python/stable/bin/pip3 isn't being used.
4. There's an issue in the ruby script that hits a problem if `gem` is yet available in the PATH of a newly created user but is available for root. A small script update can use what the root user reports as the location for the runtime if not available for the primary user.

[...]

PR #669, opened 2020-11-30
Update codespaces-linux to use Ubuntu 20.04, adopt its Java/Ruby versions, add .NET 5

Chuxel:
This PR implements and #623 and #654.

Specifically it:
1. Makes the needed adjustments to adopt Oryx's new support of Java, Ruby, .NET 5 platforms (e.g. .NET symlinks).
2. Removes steps previously required with Debian Stretch and prior to this support being released while keeping additional tools installed.
3. Adds a legacy codespaces-linux-stretch definition so that it can be selected when backwards compatibility is required.
4. Adds the Moby CLI (aka docker) into both it and the stretch based image so they can be more easily used.

The major version was bumped for the focal version avoid unexpected breaks. So, for tags:

* 0-linux = stretch
* 1-linux = focal
* latest == 1-linux

In addition, a `focal` and `stretch` tag variants were added. e.g.

* stretch
* 0-stretch
* focal
* 1-focal


bratsche (approved):
Testing in my own repo and everything seems fine. Once this is merged I can test further using the "dev" tag.

PR #670, opened 2020-12-01
Include Python virtual env name on the Codespaces bash terminal prompt

jkeech:
Normally when you activate a Python virtual env, it updates the terminal prompt to include the env name wrapped in parenthesis as a prefix on the PS1. Currently, that doesn't work with the default bash `codespaces` theme which hardcodes the PS1 value in the prompt function.

This change updates the theme to prepend the virtual env name to the prompt if a virtual env is defined via the `$VIRTUAL_ENV` environment variable, which is similar to the logic that the venv activate script runs.

image


Chuxel:
@jkeech Should we merge this in with #669 ?  

Also as a FYI, if you update it in script-library, it will automatically copy to other locations, but totally cool to update it everywhere as well.

jkeech:
> @jkeech Should we merge this in with #669 ?

I almost pushed these changes to your branch, but thought it might be nice to keep them separate since your PR was more scoped to just the Codespaces image. Either way seems fine. Maybe we can just merge this now and the Ubuntu 20.04 update will go in separately? @bratsche is going to pick up that PR to finish it off.

Chuxel:
Sounds good.

As a heads up, there's likely going to be CI failures due to #671 - something is wrong with the zsh package in Debian stretch right now.

PR #678, opened 2020-12-06
Elixir + Phoenix + Node.js + PostgreSQL Dev Container revival

talk2MeGooseman:
Hey, this PR is a revival of PR #432. I noticed that it had gone stale after September and would really love to have this Elixir + Phoenix + Node + Postgres dev container finished and merged.

I have addressed the comments in the PR (#432) and added one or two more nice to have in the configurations. If you have any more feedback just let me know. I love using Dev Containers (and Github Codespaces) and excited to contribute to something awesome.

### DIFF
--- a/containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
+++ b/containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
@@ -41,22 +41,17 @@ RUN apt-get update \
   && /bin/bash /tmp/node-setup.sh "${NVM_DIR}" "${NODE_VERSION}" "${USERNAME}"; \
   fi \
   #
-  # Install rake, ruby-debug-ide, and debase, and dependencies
+  # Install dependencies
   && apt-get install -y build-essential \
-
   #
   # Clean up
   && apt-get autoremove -y \
   && apt-get clean -y \
   && rm -rf /var/lib/apt/lists/* /tmp/common-setup.sh /tmp/node-setup.sh
 
-# Switch to non-root user to install elixir dependencies
-USER $USERNAME
-
-# Install hex & rebar package managers & Phoenix framework
-RUN mix local.hex --force \
+RUN su ${USERNAME} -c "mix local.hex --force \
   && mix local.rebar --force \
-  && mix archive.install --force hex phx_new ${PHOENIX_VERSION}
+  && mix archive.install --force hex phx_new ${PHOENIX_VERSION}"
 
 # [Optional] Uncomment this section to install additional OS packages.
 # RUN apt-get update \
--- a/containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
+++ b/containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-# Update the VARIANT arg in devcontainer.json to pick an Elixir version: 1.9, 1.10, 1.10.4
+# Update the VARIANT arg in docker-compose.yml to pick an Elixir version: 1.9, 1.10, 1.10.4
 ARG VARIANT=latest
 FROM elixir:${VARIANT}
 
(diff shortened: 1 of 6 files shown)


containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,61 @@
+# Update the VARIANT arg in devcontainer.json to pick an Elixir version: 1.9, 1.10, 1.10.4
```
Chuxel, line 1: Comment tweak: needs to be updated in docker-compose.yml.

containers/elixir-phoenix-postgres/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,37 @@
...
+	],
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	"forwardPorts": [4000, 4001, 5432],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "mix deps.get"
+
+	// Uncomment to connect as a non-root user. See https://aka.ms/vscode-remote/containers/non-root.
+	// "remoteUser": "vscode"
```
Chuxel, line 36: Can you update this to default to `vscode`? We switched our standard recently.

containers/elixir-phoenix-postgres/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Adding the definition to your folder
+
+1. If this is your first time using a development container, please follow the [getting started steps](https://aka.ms/vscode-remote/containers/getting-started) to set up your machine.
+
+2. To use VS Code's copy of this definition:
+
+   1. Start VS Code and open your project folder.
+   2. Press <kbd>F1</kbd> select and **Remote-Containers: Add Development Container Configuration Files...** from the command palette.
+   3. Select the Ruby definition.
```
Chuxel, line 55: Wrong definition name in README here.

Chuxel (commented):
This is looking great!

One thing we're now doing is defaulting to the "vscode" user rather than root. However, when I switched to vscode, running `test.sh` failed likely due to something missing from the PATH or environment.

```bash
$ mix phx.new example --live

** (Mix) The task "phx.new" could not be found
Note no mix.exs was found in the current directory
```

But this works as root

```bash
$ sudo mix phx.new example --live 

* creating example/config/config.exs
* creating example/config/dev.exs
* creating example/config/prod.exs
* creating example/config/prod.secret.exs
...
```
Otherwise this is looking good!

talk2MeGooseman:
@Chuxel Thanks for the review and testing out the code. Found the issue with trying to run things as the non-root user. The user doesn't have access to anything installed by the root user with mix so you have to switch to that user before then in the Dockerfile.
I am not 100% great with the Docker-foo so let me know if this is a valid solution.

Chuxel:
@talk2MeGooseman Yep, that works, but to avoid the extra layer and leave the image running as root by default, you can use the `su` command to temporarily run as the other user.  e.g.

```bash
RUN su ${USERNAME} -c "mix local.hex --force \
  && mix local.rebar --force \
  && mix archive.install --force hex phx_new ${PHOENIX_VERSION}"
 ```

talk2MeGooseman:
@Chuxel Thanks for the feedback and awesome tip

talk2MeGooseman:
@PavelSosin-320 The image capable of running both on a local machine and within Codespaces. Curious to know if you see any glaring red flags? So I can keep a lookout for those.

Chuxel:
LGTM!  Thanks!

PavelSosin-320:
@talk2MeGooseman The Docker's "Checker flag" is the correct configuration and implementation of the HEALTHCHECK [Docker healthcheck](https://docs.docker.com/engine/reference/run/#healthcheck).
If I can configure docker run --healthcheck options in such a way that ensures that no Docker default timeout, no network timeout no ctrl/c from the user will happen because of the slow container instantiation of the Image it is OK for me regardless of how many things it contains.  Otherwise, I suspect that the user will have trouble with it. According to the Murphy law, it will certainly happen. Since docker ps will not contain any useful information about container status it will be "it is always OK on my computer" issue.

PR #686, opened 2020-12-13
codespaces-linux: fix PATH for npm

sarisia:
npm places symlink of installed tools to `{prefix}/bin`.
Reference: https://docs.npmjs.com/cli/v6/configuring-npm/folders#executables


Chuxel:
Thanks much!  This is indeed a bug.  //cc: @jkeech as FYI

PR #698, opened 2020-12-23
Tidy up whitespace in Go dev container

twpayne:
This PR:

* Removes trailing whitespace from config files to reduce future diff sizes (many editors automatically remove trailing whitespace).
* Formats Go code with `gofmt`.

There are no functional changes.


lostintangent:
LGTM!

Issue #627, opened 2020-10-28
Investigate a full docker-in-docker script


Chuxel:
Currently the docker-from-docker definitions and docker library script use the host's container engine from inside the container by forwarding the docker socket. The downside with this approach is limitations around bind mounts - which can be scenario blocking in some cases. 

While generally [docker-in-docker isn't really recommended in most cases](https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/), it is worth understanding how easy it would be to create an alternate script that sets up full docker in a container (or just moby) - downsides aside.

Chuxel:
This was merged with #702 and will be used by the default codespaces image once deplyed in addition to the new docker-in-docker definition.

PR #702, opened 2020-12-31
[Codespaces] docker-from-docker ---> docker-in-docker 

joshspicer:
Work item: https://github.com/microsoft/vssaas-planning/issues/1588, https://github.com/microsoft/vscode-dev-containers/issues/627

Changes to default codespace image to no longer pass the docker socket to child container, and instead runs it's own docker daemon inside.  Required parent container to be privileged.


containers/codespaces-linux/.devcontainer/library-scripts/docker-in-docker-debian.sh
```diff
@@ -0,0 +1,131 @@
...
+sudoIf()
+{
+    if [ "\$(id -u)" -ne 0 ]; then
+        sudo "\$@"
+    else
+        "\$@"
+    fi
+}
+
+set -e
```
joshspicer, line 118: Anything above this was borrowed from the `docker-debian.sh` script.  A bit was removed and the `sudoIf dockerd &` was added in

script-library/docker-in-docker-debian.sh
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env bash
```
joshspicer, line 1: mirror of the other docker-in-docker-debian.sh file

containers/codespaces-linux/.devcontainer/base.Dockerfile
```diff
@@ -45,8 +45,8 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && bash /tmp/scripts/github-debian.sh \
     && bash /tmp/scripts/azcli-debian.sh \
     && bash /tmp/scripts/kubectl-helm-debian.sh \
-    # Install Moby CLI
-    && bash /tmp/scripts/docker-debian.sh "true" "/var/run/docker-host.sock" "/var/run/docker.sock" "${USERNAME}" "true" \
+    # Install Moby CLI and Engine
```
jkeech, line 48: Is this true? If so, why do we need to `apt-get install moby-engine` below?

containers/codespaces-linux/.devcontainer/library-scripts/docker-in-docker-debian.sh
```diff
@@ -0,0 +1,131 @@
...
+
+# Start docker/moby engine
+sudoIf dockerd &
```
jkeech, line 121: This script doesn't seem to install moby engine, so there's no guarantee that moby engine would exist? I think the simplest solution would be to move the install of moby-engine into this script and out of the codespaces-linux Dockerfile, that way you can keep this logic here. Although you'd also need to set up the docker engine in the non-moby case as well, even though that doesn't apply to Codespaces.

jkeech (approved):
LGTM, although I recommend changing where moby-engine in installed in the scripts to consolidate it all in the `docker-in-docker-debian.sh` file.

Chuxel (commented):
A few  comments.  

We also may want to create a docker-in-docker definition like the docker-from-docker one to make testing the script easier and illustrating its use.  Right now kitchen sink takes a long time to build so validation is time consuming mainly because of build time. :smile:

Also, out of curiosity, I know the [offical "docker:dind" image](https://github.com/docker-library/docker) uses this script: https://github.com/moby/moby/blob/master/hack/dind

Do you know if any of the items listed here are still required?

I also noticed that their official image [dockerd entrypoint](https://github.com/docker-library/docker/blob/a7534626601bce99a23f0496a3f659616522ed22/dockerd-entrypoint.sh#L139) seems to remove the pid of dockerd before starting it.

Not doing this this may be similar to how we saw random "kills" happen with sshd in the past.

```
find /run /var/run -iname 'docker*.pid' -delete || :
```

joshspicer:
> We also may want to create a docker-in-docker definition like the docker-from-docker one to make testing the script easier and illustrating its use. Right now kitchen sink takes a long time to build so validation is time consuming mainly because of build time. 😄

Agreed on this - i've watched the image build several times working on this :D

joshspicer:
> Also, out of curiosity, I know the [offical "docker:dind" image](https://github.com/docker-library/docker) uses this script: https://github.com/moby/moby/blob/master/hack/dind

I was looking and based some of the changes off this [old Dockerfile](https://github.com/jpetazzo/dind/blob/master/Dockerfile) for dind along with a couple more recent blog posts on the matter - i'll take another look at what you shared

containers/docker-in-docker/.devcontainer/library-scripts/README.md
```diff
@@ -0,0 +1,5 @@
+# Warning: Folder contents may be replaced
+
```
joshspicer, line 2: @Chuxel  Is there anything I need to do additionally in setup for the comment in this readme to be true?

joshspicer:
@Chuxel I've added a definition "docker-in-docker" using the shared script file

containers/docker-in-docker/.devcontainer/library-scripts/README.md
Chuxel: Nope - as long as the name in the `/script-library` folder matches the `.devcontainer/library-scripts` folder, it will be automatically updated if the `/script-library` file is updated after a CI run.

Chuxel (commented):
A few other comments.  One is that the docker-in-docker definition doesn't appear to succesfully start `dockerd` (at least on Win 10 w/Docker in WSL mode).

That appears to be this issue: https://superuser.com/questions/1468316/which-kernel-moduules-are-really-needed-for-docker-installation

Related: https://stackoverflow.com/questions/59096569/cannot-connect-to-the-docker-daemon-on-debian-buster-on-wsl-2

The suggested fix is to update iptables to iptables-legacy.

```
update-alternatives --set iptables /usr/sbin/iptables-legacy
update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
```

That seemed to do the trick. (I did not check if this is needed for Ubuntu 20.04.)

joshspicer:
Testing with the vscode-dev-containers extension is consistently good.  Still need to test more e2e in codespaces

Chuxel:
It looks like its working, yes!  

A few other things I noticed on close inspection:


1. I noticed that the image startup has a reference to overlay2 not being supported.  I think you inadvernatly removed the `VOLUME` reference you had before.  That probably should be in both the Dockerfils and in the docs about it.

    Error:
    ``` 
    time="2021-01-07T01:59:27.089974857Z" level=error msg="failed to mount overlay: invalid argument" storage-driver=overlay2
    time="2021-01-07T01:59:27.091006166Z" level=warning msg="[graphdriver] WARNING: the aufs storage-driver is deprecated, and will be removed in a future release"
    ```

    Adding `VOLUME [ "/var/lib/docker" ]` back fixes it.

    Another up-side is as we add rebuild support, the container's d-in-d build cache and containers should survive a rebuild too.

2. Maybe we should also add `( sudoIf dockerd > /tmp/dockerd.log 2>&1 ) &` to the init script so the dockerd log can be accessed a bit more easily at /tmp/dockerd.log for troubleshooting. I had to catch the terminal window for the dev container startup to find this which has a bunch of other stuff in it.

3. We probably should also install `pigz` which is a faster gzip implementation dockerd can use.

4. I also saw this in the logs, but don't immediatley know the fix for it:

    ```
    time="2021-01-07T02:16:56.288684776Z" level=warning msg="Not using native diff for overlay2, this may cause degraded performance for building images: kernel has CONFIG_OVERLAY_FS_REDIRECT_DIR enabled" storage-driver=overlay2
    ```

joshspicer:
One more bug I fixed when docker container fails unexpectedly (or manually with `docker kill <ID>` )

```log
time="2021-01-07T15:50:30.662286100Z" level=info msg="Starting up"
time="2021-01-07T15:50:30.663969300Z" level=info msg="libcontainerd: containerd is still running" pid=28
time="2021-01-07T15:50:30.664089000Z" level=info msg="parsed scheme: \"unix\"" module=grpc
time="2021-01-07T15:50:30.664233200Z" level=info msg="scheme \"unix\" not registered, fallback to default scheme" module=grpc
time="2021-01-07T15:50:30.664290900Z" level=info msg="ccResolverWrapper: sending update to cc: {[{unix:///var/run/docker/containerd/containerd.sock 0  <nil>}] <nil>}" module=grpc
time="2021-01-07T15:50:30.664340800Z" level=info msg="ClientConn switching balancer to \"pick_first\"" module=grpc
failed to start containerd: timeout waiting for containerd to start
```

Fixed with another find/delete statement `find /run /var/run -iname 'container*.pid' -delete || :`

joshspicer:
@Chuxel On my end this looks good to go!  Are you able to kick off a release for us (bumping the minor version probably)?

Chuxel:
LGTM!  @chrmarti - FYI, I'm going to do a minor version release to queue this up for further testing in Codespaces.

Chuxel:
@joshspicer  This will be version 1.1.0 when it goes out.

Release run: https://github.com/microsoft/vscode-dev-containers/actions/runs/470010714

PR #703, opened 2021-01-03
Add Ruby on Rails & PostgreSQL Definition

jarrodldavis:
This is a new Docker Compose container definition that builds off of the existing [Ruby on Rails] definition and incorporates aspects of the [Python 3 & PostgreSQL] definition for a closer-to-production development environmnet for Postgres-backed Rails applcations.

[Ruby on Rails]: https://github.com/microsoft/vscode-dev-containers/tree/master/containers/ruby-rails
[Python 3 & PostgreSQL]: https://github.com/microsoft/vscode-dev-containers/tree/master/containers/python-3-postgres


Chuxel:
@jarrodldavis Thanks for the contribution.  This one got away from us.  Since it's been so long and this generally looks good, I'll merge.  There's a couple comment tweaks I'll make after the fact and CC you on the PR.

Issue #508, opened 2020-08-20
A new utility script for resolving host path


felipecrs:
How about something like:

`/usr/local/bin/localpath`
```bash
#!/bin/sh
echo "$1" | sed "s#/workspace#${LOCAL_WORKSPACE_FOLDER}#"
```

And of course, setting the `LOCAL_WORKSPACE_FOLDER` on the container (can we do that by default?): 
```jsonc
{
  "remoteEnv": {
    "LOCAL_WORKSPACE_FOLDER": "${localWorkspaceFolder}"
  },
}
```

So the users could use it easier:

```bash
docker run -v "$(localpath $PWD):/workdir" ubuntu
```

PR #720, opened 2021-01-15
Add localpath utility to docker scripts

felipecrs:
```sh-session
❯ pwd
/workspaces/docker-images
❯ localpath "$(pwd)"
/home/felipesantos/repos/docker-images
❯ localpath ..
/home/felipesantos/repos
❯ localpath /tmp
/tmp
```

Closes https://github.com/microsoft/vscode-dev-containers/issues/508.

### DIFF
--- a/script-library/docker-in-docker-debian.sh
+++ b/script-library/docker-in-docker-debian.sh
@@ -170,3 +170,18 @@ EOF
 
 chmod +x /usr/local/share/docker-init.sh
 chown ${USERNAME}:root /usr/local/share/docker-init.sh
+
+# localpath utility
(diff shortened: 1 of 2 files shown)


felipecrs:
I don't think this is a good idea anymore. For those using docker-on-docker (rather than docker-in-docker), I strongely suggest to mount the workspace using the same path as in host with the following:

```jsonc
// devcontainer.json
{
  "workspaceMount": "source=${localWorkspaceFolder},target=${localWorkspaceFolder},type=bind,consistency=cached",
  "workspaceFolder": "${localWorkspaceFolder}"
}
```

PR #724, opened 2021-01-17
Update F# to use .NET 5 only

cartermp:
The latest Ionide plugin (5.x) requires the .NET 5 SDK. F# developers also tend to update things very quickly, so I think we can just drop the other .NET Core SDK options. Otherwise, to make this function with the other SDKs, we'd have to install specific versions of the plugin.


cartermp:
cc @Chuxel

Chuxel:
LGTM!  Thanks!

PR #730, opened 2021-01-23
Bump pg-promise version

matthewisabel:
Underlying pg library doesn't work on node 14 with this pg-promise version. Needs pg-promise bump to fix this: https://github.com/brianc/node-postgres/issues/2180


Chuxel:
LGTM!

PR #732, opened 2021-01-25
Update R Docker image and vscode-r recommended settings

eitsupi:
This pull request updates R to latest version and add [`vscode-R`](https://github.com/Ikuyadeu/vscode-R) recommended settings.

- Change base image `rocker/r-apt:bionic`(R 3.6.3) to [`rocker/r-ver`](https://github.com/rocker-org/rocker-versioned2), which provide specific version R (4.0.0 ~ 4.0.4) based on Ubuntu LTS.
- Change not to install `libzip-dev`. (`libzip-dev` was needed for [R Tools](https://github.com/MikhailArkhipov/vscode-r) #60, which is not included by #473 ).
- Install `libxt-dev`, for to plot without warning.
- Add vscode-R recommended settings and packages.
  - Install [`R LSP Client`](https://github.com/REditorSupport/vscode-r-lsp) and [`R Language Server`](https://github.com/REditorSupport/languageserver).
    - Install `libcurl4-openssl-dev`, `libssl-dev`, `libxml2-dev`, for to install `R Language Server`.
  - Install [`devtools`](https://github.com/r-lib/devtools)
    - Upgrade OS packages and install `libgit2-dev`, for to install `devtools`
  - Install [`radian`](https://github.com/randy3k/radian).
    - Install `python3-pip`, for to install `radian`.
- Add vscode-R recommended settings.

### DIFF
--- a/containers/r/.devcontainer/Dockerfile
+++ b/containers/r/.devcontainer/Dockerfile
@@ -1,19 +1,35 @@
-FROM rocker/r-apt:bionic
+# [Choice] R version: latest, ... ,4.0.1 , 4.0.0
+ARG VARIANT="latest"
+FROM rocker/r-ver:${VARIANT}
 
+# Use the [Option] comment to specify true/false arguments that should appear in VS Code UX
+#
 # [Option] Install zsh
 ARG INSTALL_ZSH="true"
 # [Option] Upgrade OS packages to their latest versions
 ARG UPGRADE_PACKAGES="false"
 
 # Install needed packages and setup non-root user. Use a separate RUN statement to add your own dependencies.
-ARG USERNAME=docker
+ARG USERNAME=vscode
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 COPY library-scripts/*.sh /tmp/library-scripts/
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && /bin/bash /tmp/library-scripts/common-debian.sh "${INSTALL_ZSH}" "${USERNAME}" "${USER_UID}" "${USER_GID}" "${UPGRADE_PACKAGES}" \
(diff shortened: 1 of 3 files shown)


eitsupi:
@Chuxel I apologize for including the fix of username acaba59 in this PR.
You've changed username to `docker` in e2480ab, but I think `vscode` is more better username because this is other container's default username.
So, I have changed the username to `vscode` again c1548b4.

2percentsilk:
@kmehant as the R definition maintainer, can I get your eyes on this? 👀

eitsupi:
@2percentsilk @Chuxel @kmehant If this review is going to take a long time, would it be better to break the PR into base image update and others?

kmehant:
@eitsupi Looks good to me! However, I request a review from @Chuxel.
It would be great if you could squash commits wherever possible. 

@2percentsilk Thanks for tagging me, I apologize for the late response.

Chuxel:
LGTM! The new extension is low volume which gave me pause, but I see its maintained by the same maintainers as the existing extension. So from that perspective it's probably ok.  

Made one fix to move the remoteUser back to vscode given this is what the dockerfile does.

Chuxel:
@eitsupi @kmehant Actually, one thing I just noticed - the `# [Choice]` comment in the Dockerfile also can drive VS Code UX, but it has to be a concrete list.  

At the moment I've removed the comment since I'm not sure if there's higher level tags like "4" and the lower level versions are going to be a bit of a maintenance issue.

eitsupi:
@Chuxel Thank you for merging and the comment fix.

There is no higher level tags like "4", and minor versions like "4.0.4" will increase every few months.
So if there is no problem with the current state, I think it's fine the current state which only removed the `# [Choice]` comment.

PR #734, opened 2021-02-01
[docker-in-docker] sudoIf the find -delete statements

joshspicer:
Fix the docker-in-docker script to work with non-root sudoers


Chuxel:
LGTM!

PR #737, opened 2021-02-04
Add Terragrunt to Azure Terraform definition

maximivanov:
Proposed change is to include [Terragrunt](https://github.com/gruntwork-io/terragrunt) in the Azure Terraform definition.

> Terragrunt is a thin wrapper for Terraform that provides extra tools for working with multiple Terraform modules.

How I tested it: opened the `containers/azure-terraform` in VSCode and ran "Reopen in Container".
Once container started, I confirmed Terragrunt was installed:

Screen Shot 2021-02-04 at 13 36 34

I realize not everyone uses Terragrunt, so possibly we can set

```json
// containers/azure-terraform/.devcontainer/devcontainer.json

...
"args": {
    ...
    "TERRAGRUNT_VERSION": "none",
}
```

to not have it installed by default. If a developer wants to use it, they will have to change the argument to either a specific version or to `latest`.

### DIFF
--- a/containers/azure-terraform/.devcontainer/Dockerfile
+++ b/containers/azure-terraform/.devcontainer/Dockerfile
@@ -36,10 +36,11 @@ RUN if [ "${INSTALL_AZURE_CLI}" = "true" ]; then bash /tmp/library-scripts/azcli
     fi \
     && rm -rf /var/lib/apt/lists/*
 
-# Install Terraform, tflint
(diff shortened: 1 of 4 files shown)


2percentsilk:
@cmendible as the original creator/maintainer for this definition, what are your thoughts on this addition?

cmendible:
Awesome! 

LGTM

PR #741, opened 2021-02-05
hardcode azure dns for docker daemon in docker-in-docker

joshspicer:
Fixes: https://github.com/microsoft/vssaas-planning/issues/2499

This DNS address was taken from: 
https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-name-resolution-for-vms-and-role-instances#considerations
> The Azure DNS IP address is 168.63.129.16. This is a static IP address and will not change.

### DIFF
--- a/script-library/docker-in-docker-debian.sh
+++ b/script-library/docker-in-docker-debian.sh
@@ -159,7 +159,7 @@ fi
 ## Dind wrapper over.
 
 # Start docker/moby engine
-( sudoIf dockerd > /tmp/dockerd.log 2>&1 ) &
+( sudoIf dockerd --dns 168.63.129.16 > /tmp/dockerd.log 2>&1 ) &
 
 set +e
 
--- a/script-library/docker-in-docker-debian.sh
+++ b/script-library/docker-in-docker-debian.sh
@@ -160,7 +160,7 @@ fi
 
 # Handle DNS
 set +e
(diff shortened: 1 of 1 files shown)


script-library/docker-in-docker-debian.sh
```diff
@@ -159,7 +159,7 @@ fi
 ## Dind wrapper over.
 
 # Start docker/moby engine
-( sudoIf dockerd > /tmp/dockerd.log 2>&1 ) &
+( sudoIf dockerd --dns 168.63.129.16 > /tmp/dockerd.log 2>&1 ) &
```
jkeech, line 162: Can we detect if the container is running in Azure and only set this in that case? We want to be sure to allow this image to work well offline / outside of Codespaces as well. But when running in the Codespaces infrastructure, we want it to default to this DNS server.

joshspicer:
Looks like docker does take the contents of `/etc/resolv.conf` from the host and sets that to the dns servers for its child containers.  The reason this doesn't work on Azure is that the `nameserver` set is `127.0.0.53`, which does not resolve inside the containers as its not a globally routable ip address

A check for this address in `/etc/resolv.conf` should signal us to override the DNS to something known good (like the Azure recursive resolver)

jkeech (approved):

Chuxel:
LGTM - I assume we want to rev the kitchen sink image given this is blocking?

Chuxel (approved):

PR #744, opened 2021-02-09
update go container build script and devcontainer.json

hyangah:
Reduce the number of pre-installed go tools - gopls is the
default language feature provider for golang.go since its
v0.22.0, so preinstall only the required tools with the
default settings. For additional tools, the extension will
prompt users to install when necessary.

Also, clean up devcontainer's go extension settings.
Some gopls settings are outdated. Some [go] settings are
unnecessary because the extension sets them up by default,
and the gopls update setting is now obsolete.

### DIFF
--- a/containers/codespaces-linux-stretch/.devcontainer/devcontainer.json
+++ b/containers/codespaces-linux-stretch/.devcontainer/devcontainer.json
@@ -5,11 +5,10 @@
 	},
 	"settings": {
 		"terminal.integrated.shell.linux": "/bin/bash",
-		"go.useGoProxyToCheckForToolUpdates": false,
+		"go.toolsManagement.checkForUpdates": "off",
 		"go.useLanguageServer": true,
 		"go.gopath": "/go",
 		"go.goroot": "/usr/local/go",
-		"go.toolsGopath": "/go/bin",
 		"python.pythonPath": "/opt/python/latest/bin/python",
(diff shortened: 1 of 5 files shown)


hyangah:
cc @Chuxel

Chuxel:
LGTM! This is great since it reduces dependencies for both the go and codespaces-linux images.

//cc @joshspicer @jkeech - We likely should adopt the settings in this PR in the codespaces-linux, codespaces-linux-stretch, and Codespaces automatic settings applied in the default case.

Chuxel:
@hyangah Actually a couple of other questions, are either of these needed?

github.com/stamblerre/gocode
golangci/golangci-lint

I'd assume the first one is replaced by gopls, correct? Is the second needed anymore?

hyangah:
Thanks @Chuxel - Removed gocode-gomod.
Also noticed golangci-lint is installed using their install script, so removed it from the GO_TOOLS_WITH_MODULES tool list. (Now renamed it to GO_TOOLS ;-)

golangci-lint, revive, staticcheck are alternative linter tools users often configure instead of the (soon-to-be-deprecated) golint in their workspace settings. So I decided to leave them in the list for now.

Chuxel (approved):
LGTM!

@joshspicer @jkeech - This updates codespaces-linux, but we can bump up the version number there independently to decide when we take it.

PR #745, opened 2021-02-11
Install python3-dev package for scenario

dmgardiner25:
Updates the `codespaces-linux` base `Dockerfile` to install `python3-dev` per the conversation in [#2096](https://github.com/microsoft/vssaas-planning/issues/2096).

### DIFF
--- a/containers/codespaces-linux/.devcontainer/base.Dockerfile
+++ b/containers/codespaces-linux/.devcontainer/base.Dockerfile
@@ -56,7 +56,8 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get autoremove -y && apt-get clean -y
 
 # Install Python, PHP, Ruby utilities
-RUN bash /tmp/scripts/python-debian.sh "none" "/opt/python/latest" "${PIPX_HOME}" "${USERNAME}" "true" \
+RUN apt-get install -yq python3-dev \
+    && bash /tmp/scripts/python-debian.sh "none" "/opt/python/latest" "${PIPX_HOME}" "${USERNAME}" "true" \
     # Install rvm, rbenv, base gems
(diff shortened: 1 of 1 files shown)


containers/codespaces-linux/.devcontainer/base.Dockerfile
```diff
@@ -56,7 +56,8 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get autoremove -y && apt-get clean -y
 
 # Install Python, PHP, Ruby utilities
-RUN bash /tmp/scripts/python-debian.sh "none" "/opt/python/latest" "${PIPX_HOME}" "${USERNAME}" "true" \
+RUN apt-get install -yq python3-dev \
```
jkeech, line 59: Should we add this to the python-debian script instead?

dmgardiner25: Do you mean add it to the `PREREQ_PKGS`? Since we get Python from the Oryx image, we don't build from source meaning we don't install those prereqs. If you mean add it elsewhere, that would work but I assumed it was a Codespaces specific change so I only wanted to make it affect us.

jkeech: Ah, ok. In that case this seems reasonable.

Chuxel: This section doesn't otherwise install OS packages and would need a "apt-get clean -y" added at a minimum.  It may be easier to just add it to this line above:

```
 # Verify expected build and debug tools are present
 apt-get -y install build-essential cmake cppcheck valgrind clang lldb llvm gdb 
```

Chuxel:
LGTM!

PR #748, opened 2021-02-12
need to escape variables in docker-init script template

joshspicer:
When the docker-init is written to file an unescape variable will be evaluated.  This causes the `$?` to always evaluate to its first value.

### DIFF
--- a/script-library/docker-in-docker-debian.sh
+++ b/script-library/docker-in-docker-debian.sh
@@ -161,7 +161,7 @@ fi
 # Handle DNS
 set +e
 cat /etc/resolv.conf | grep -i 'internal.cloudapp.net'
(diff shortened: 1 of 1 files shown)


Chuxel (approved):
LGTM!  One note is that if the script doesn't need any variables from the wrapping shell script that uses it you can tweak the syntax slightly so you don't have to escape anything.

Use
```
<< 'EOF'
```
Rather than
```
<< EOF
```

The original docker script pulled in some info from the wrapping shell script. I don't know that this one needs to do that.

joshspicer:
Oo @Chuxel that's a great tip!  I'm going to try that out

joshspicer:
This seems to work great, @Chuxel !

PR #753, opened 2021-02-17
Updates for new versions of Go, Alpine

Chuxel:
Go 1.16 is out, so we can now build it. 1.14 is no longer receiving security patches, so that image version is now deprecated.

Alpine 3.13 is out, but 3.10 is not yet out of support, so this is just an addition.


odannyc:
@Chuxel  Do you know when this will make it to `mcr.microsoft.com` so we can use it like so:

```
FROM mcr.microsoft.com/vscode/devcontainers/go:0-1.16
```

Chuxel:
@odannyc A development image is available now.  Use `mcr.microsoft.com/vscode/devcontainers/go:dev-1.16`  the "dev" version updates with merges into master, so its unstable, but quick releasing.  We'll do a point release of the repository soon which will publish the image.

Chuxel:
@odannyc The image is now released.

PR #756, opened 2021-02-19
add ballerina language devcontainer

gokomer:
This PR adds ballerina language devcontainer.

### DIFF
--- a/containers/ballerina/.devcontainer/devcontainer.json
+++ b/containers/ballerina/.devcontainer/devcontainer.json
@@ -1,0 +1,19 @@
+{
+	"name": "Ballerina (Community)",
+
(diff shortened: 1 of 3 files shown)


bamurtaugh:
Thank you for the contribution @gokomer!

To ensure those who use this definition [know what's inside it](https://github.com/microsoft/vscode-dev-containers/blob/master/CONTRIBUTING.md#a-note-on-referenced-images-and-dockefile-contents), would you be able to include the Dockerfile used to build it (ideally from our [template](https://github.com/microsoft/vscode-dev-containers/tree/master/container-templates)/guidance so it has some of the common code we'd expect), and reference the [ballerina image from DockerHub](https://hub.docker.com/r/ballerina/ballerina)?

PR #757, opened 2021-02-19
Minor doc fix for python install script

CliffJumper:
I found a super-minor issue in the python.md when reading it to setup a dev environment.  In a previous change, the line after the `RUN apt-get update && bash /tmp/library-scripts/python-debian.sh "3.8.3" "${PYTHON_PATH}" "${PIPX_HOME}" \ ` was removed, but the `\` was left.  If someone is copy-pasting to play with it, they'll get an error because of it.


Aisuko (approved):
Approve

Aisuko:
@CliffJumper Good eye, thanks for contributing.

Issue #674, opened 2020-12-02
Remove Oh My Bash in favor of a custom PS1


Chuxel:
Currently we are using Oh My Bash for the default prompt in dev containers that use the common-debian.sh script (partly due to Bash It not having any license). While this works well, there are challenges that people are hitting where they've customized their shell prompt and Oh My Bash is overriding it.  Unlike Oh My Zsh!, Oh My Bash is not ubiquitous so this can be counter intuitive. It may be better to instead create a custom prompt that matches the UX we currently have, but not use Oh My Bash to do it. 

Anyone that then wants to use Oh My Bash or Bash It could install these on their own.

We can replace this with a PS1 script that mimics the current default shell behavior.

//cc: @jkeech given #670 is an example of this problem. @matthewisabel I think encountered a similar problem with his dotfiles.

burkeholland:
This was confusing for me because I thought it was zsh. And I could not figure out why, despite my setting it to bash in the config file, it was defaulting to zsh. But I also have never heard of oh-my-bash.

I think this is going to catch people off guard. I'd expect to see a normal, non-customized bash prompt.

Chuxel:
@burkeholland Yeah, the surprising part is it seems to be hijacking customization. That's too bad since it adds other value as well.  But I agree that the way it is now is probably more trouble than it is worth.

Chuxel:
A PR is up to remove switch to a PS1 script instead of Oh My Bash. Once it is no longer in images, you can install it using your `dotfiles` or in a Dockerfile as follows:

From dotfiles / a terminal window:
```bash
rm -rf ~/.oh-my-bash
bash -c "$(curl -fsSL https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/tools/install.sh)"
```

In a Dockerfile:
```bash
ARG USERNAME=vscode
RUN su $USERNAME -c 'rm -rf $HOME/.oh-my-bash && bash -c "$(curl -fsSL https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/tools/install.sh)"'
```

PR #761, opened 2021-02-22
Sunset Oh My Bash

Chuxel:
This PR moves away from using Oh My Bash! by default and replaces it with a PS1 script that does exactly what the default prompt did.  This should resolve conflicts described in #674 and drop a dependency.  The Python extension's prefixing should also work by default with this update.

A stub script is left in the image under `~/.oh-my-bash/oh-my-bash.sh` that lets people know this has happened in the event OMB is referenced in their `dotfiles` and by extension their `~/.bashrc` or `~/.bash_profile` files.

The only downside with doing this is you need to remove the folder before installing OMB if you want it. At the moment its erroring on the side of awareness despite that fact.  I expect that after a few months we can remove this stub.

//cc: @matthewisabel @lostintangent @bamurtaugh

### DIFF
--- a/script-library/common-alpine.sh
+++ b/script-library/common-alpine.sh
@@ -225,14 +225,24 @@ EOF
 )"
 
 # Add notice that Oh My Bash! has been removed from images and how to provide information on how to install manually
+OMB_README="$(cat \
+<<'EOF'
+"Oh My Bash!" has been removed from this image in favor of a simple shell prompt. If you still
+wish to use it, remove "~/.oh-my-bash" and install it from: https://github.com/ohmybash/oh-my-bash
+You may also want to consider "Bash-it" as an alternative: https://github.com/bash-it/bash-it
+See https://github.com/microsoft/vscode-dev-containers/issues/674#issuecomment-783474956
+EOF
+)"
 OMB_STUB="$(cat \
 <<'EOF'
 #!/usr/bin/env bash
+cd "$(dirname $0)"
 if [ -t 1 ]; then
-    echo -e '"Oh My Bash!" has been removed from this image in favor of a simple shell prompt. If you still\nwish to use it, remove "~/.oh-my-bash" and install it from: https://github.com/ohmybash/oh-my-bash\nYou may also want to consider "Bash-it" as an alternative: https://github.com/bash-it/bash-it'
+    cat README.md
 fi
 EOF
 )"
+
 
 # Adapted Oh My Zsh! install step to work with both "Oh Mys" rather than relying on an installer script
 # See https://github.com/ohmyzsh/ohmyzsh/blob/master/tools/install.sh for offical script.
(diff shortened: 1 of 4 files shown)


script-library/common-alpine.sh
```diff
@@ -184,49 +184,66 @@ chmod +x /usr/local/bin/code
...
+#!/usr/bin/env bash
+cd "$(dirname $0)"
+if [ -t 1 ]; then
+    cat README.md
+fi
+EOF
+)"
+
+
 # Adapted Oh My Zsh! install step to work with both "Oh Mys" rather than relying on an installer script
```
jkeech, line 247: Does this comment need to change now that OMB is being removed?

script-library/common-alpine.sh
```diff
@@ -249,32 +266,45 @@ install-oh-my()
...
 
     if [ "${USERNAME}" != "root" ]; then
-        cp -rf ${USER_RC_FILE} ${OH_MY_INSTALL_DIR} /root
-        chown -R ${USERNAME}:${USERNAME} ${USER_RC_PATH}
+        cp -rf "${USER_RC_FILE}" "${OH_MY_INSTALL_DIR}" /root
+        chown -R ${USERNAME}:${USERNAME} "${USER_RC_PATH}"
     fi
 }
 
+# Add RS snippet and custom bash prompt
```
jkeech, line 284: Typo? RS -> RC

```suggestion
# Add RC snippet and custom bash prompt
```

script-library/common-alpine.sh
```diff
@@ -184,49 +184,66 @@ chmod +x /usr/local/bin/code
...
+__bash_prompt
 EOF
 )"
 CODESPACES_ZSH="$(cat \
-<<EOF
+<<'EOF'
 prompt() {
-    if [ ! -z "\${GITHUB_USER}" ]; then
-        local USERNAME="gh:@\${GITHUB_USER}"
+    if [ "${GITHUB_USER}" != "" ]; then
```
jkeech, line 211: Two questions:
1) I thought these were previously escaped `\$` instead of `$` because we don't want the variable evaluated when the file is written, but rather when it is executed by the user. We want to write the literal value `${GITHUB_USER}` into the generated script on disk. Does removing the escaping break that? It would be good to confirm that the generated file has the variable references intact.
2) Is there a functional difference with the string empty check versus `-z` for 0 length strings, or is that more of a style/preference change?

jkeech (commented):
Overall looks good. I added one question around escaping that would be good to validate since it could break things in the custom prompt if it's not working like we want.

script-library/common-alpine.sh
Chuxel: We could remove it out of a function to simplify, but I opted to leave it as-is to reduce the amount of change in the PR. The Oh My Zsh! install still works fine with it, so it's not "must have" to change.

script-library/common-alpine.sh
Chuxel: Yeah on the escaping, the other thing I did was change this from `EOF` to `'EOF'` (note the ticks). `'EOF'` will not evaluate variables while `EOF` will.  Some of the other scripts take a variable from the main script - this one does not.

On `-z`, I probably could have left it alone... I must have done it at one point during edits. There's some [weirdness between unset and empty string when you're specifically looking for unset](https://mig8447.github.io/linux/bash/2018/07/25/checking-if-a-bash-variable-is-unset.html). But in this case, unset or empty string is actually what we want, so `-z` should be fine. I could revert it.

script-library/common-alpine.sh
Chuxel: Yep! Took the suggestion.

script-library/common-alpine.sh
Chuxel: I should also mention that I used a git clone directly here rather than the up-stream script since this reduces our supply chain risk a bit (we only have to consider the repository, not what the install script does). It's a supported install method for both OMB and OMZ.  OMB just had a slightly different spot for the template which is the bit I yanked.

Most of the other edits are adding quotes to things that technically should have been quoted to begin with... you just don't normally see spaces in these paths so it worked fine as it was before.

script-library/common-alpine.sh
Chuxel: @jkeech Actually, funnily enough, I noticed that code in the ZSH theme wasn't working.  Turns out, USERNAME doesn't work as a variable in this context so I switched it to a local variable called `prompt_username` and that fixed it.

script-library/common-alpine.sh
Chuxel: As a part of another change, I went ahead and dropped the function and merged it in with the ZSH setup section at the bottom.

PR #763, opened 2021-02-23
 docker composer v3.5 isn't support init flag.

philipz:
When I open in dev container, then show the below error message:
The Compose file '/home/myname/dapr/quickstarts/hello-world/.devcontainer/docker-compose.yml' is invalid because:
Unsupported config option for services.app: 'init'
Just change version 3.5 to 3.7 will be work.
[Compose file version 3 reference](https://docs.docker.com/compose/compose-file/compose-file-v3/#init)


2percentsilk:
This will be handled by https://github.com/microsoft/vscode-dev-containers/pull/765!

PR #769, opened 2021-02-25
Update Readme

tanmayeekamath:
Added a section for default kitchen sink image behavior


tanmayeekamath:
Will reopen with a few more changes.

PR #770, opened 2021-02-25
Updates to the Readme

tanmayeekamath:
Added the following:

- Section on usage with a default kitchen sink image 
- Updates on how to forward a port

### DIFF
--- a/script-library/docs/sshd.md
+++ b/script-library/docs/sshd.md
@@ -23,8 +23,8 @@
 
 This script can be used either ad-hoc in an already running container or in a Dockerfile. 
 
-### Usage when this script is already installed in an image (e.g. default Codespaces image)
-The SSH server will be installed by default for Codespaces that are created off the default container image (without a custom devcontainer configuration). 
+### Usage when this script is already installed in an image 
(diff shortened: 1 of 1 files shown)


script-library/docs/sshd.md
```diff
@@ -21,7 +21,34 @@
 
 ## Usage
 
-This script can be used either ad-hoc in an already running container or in a Dockerfile.
+This script can be used either ad-hoc in an already running container or in a Dockerfile. 
+
+### Usage with a default container image
```
Chuxel, line 26: This repository is also used for Remote - Containers repository where no concept of a default image exists.  

What I'd recommend is we say, "Usage when this script is already in an image (e.g. the default Codespaces image - codespaces-linux)" or something like that.

Chuxel (changes requested):

script-library/docs/sshd.md
tanmayeekamath: Got it.. that makes sense. I'll update the title and resubmit!

tanmayeekamath:
I have made the suggested updates now.  Thanks for the feedback!

Chuxel:
LGTM!

Issue #764, opened 2021-02-23
Add manpages, manpages-posix to C++ and kitchen sink, possibly other images


Chuxel:
We've had requests to add a few more manpages into these images.

- `manpages` and `manpages-dev` we should just be able to add to common-debian.sh
- However, `manpages-posix` and `manpages-posix-dev` are not in the "free" sources for Debian (though they are available by default in Ubuntu). We'd also need to [add non-free sources to include them](https://www.xmodulo.com/install-nonfree-packages-debian.html). On the surface this does not bother me for a general purpose dev image, but we'd likely want to have it as a script option for those using it themselves. 

This adds 6mb to the uncompressed image. The zipped image that is downloaded will be much smaller so it shouldn't significantly impact aquisition time.

//cc @2percentsilk @joshspicer @jkeech

joshspicer:
https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/common-debian.sh

PR #779, opened 2021-03-08
Aded manpages (posix, posix-dev, dev) to common-debian.sh

samruddhikhandale:
Fixes: [Add manpages, manpages-posix to C++ and kitchen sink, possibly other images](https://github.com/microsoft/vscode-dev-containers/issues/764)

Added following packages to common-debian.sh

- manpages-dev 
- manpages-posix 
- manpages-posix-dev

### DIFF
--- a/containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
+++ b/containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
@@ -70,8 +70,9 @@ apt-get-update-if-needed()
 
 # Run install apt-utils to avoid debconf warning then verify presence of other common developer tools and dependencies
 if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
-    # Adding non-free packges on Debian
-    echo -e "deb http://deb.debian.org/debian/ $(lsb_release -cs) main contrib non-free \ndeb http://deb.debian.org/debian/ $(lsb_release -cs)-updates main contrib non-free \ndeb http://deb.debian.org/debian-security/ $(lsb_release -cs)/updates main contrib non-free \ndeb http://deb.debian.org/debian/ $(lsb_release -cs)-backports main contrib non-free" >/etc/apt/sources.list
+    # Needed for adding manpages-posix and manpages-posix-dev which are non-free packages in Debian
+    CODENAME=$(lsb_release -cs)
+    echo -e "deb http://deb.debian.org/debian/ $(CODENAME) main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-updates main contrib non-free \ndeb http://deb.debian.org/debian-security/ $(CODENAME)/updates main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-backports main contrib non-free" >/etc/apt/sources.list
     apt-get-update-if-needed
 
     PACKAGE_LIST="apt-utils \
@@ -109,7 +110,7 @@ if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
         sudo \
         ncdu \
         man-db \
-        strace
+        strace \
         manpages-dev \
         manpages-posix \
         manpages-posix-dev"
--- a/containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
+++ b/containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
@@ -71,8 +71,10 @@ apt-get-update-if-needed()
 # Run install apt-utils to avoid debconf warning then verify presence of other common developer tools and dependencies
 if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
     # Needed for adding manpages-posix and manpages-posix-dev which are non-free packages in Debian
-    CODENAME=$(lsb_release -cs)
-    echo -e "deb http://deb.debian.org/debian/ $(CODENAME) main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-updates main contrib non-free \ndeb http://deb.debian.org/debian-security/ $(CODENAME)/updates main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-backports main contrib non-free" >/etc/apt/sources.list
+    if [ "$(lsb_release -is)" = "Debian" ]; then
+        CODENAME=$(lsb_release -cs)
+        echo -e "deb http://deb.debian.org/debian/ $(CODENAME) main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-updates main contrib non-free \ndeb http://deb.debian.org/debian-security/ $(CODENAME)/updates main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-backports main contrib non-free" >/etc/apt/sources.list
+    fi
     apt-get-update-if-needed
 
     PACKAGE_LIST="apt-utils \
@@ -111,6 +113,7 @@ if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
         ncdu \
         man-db \
         strace \
+        manpages \
         manpages-dev \
         manpages-posix \
         manpages-posix-dev"
--- a/containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
+++ b/containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
@@ -70,6 +70,8 @@ apt-get-update-if-needed()
 
 # Run install apt-utils to avoid debconf warning then verify presence of other common developer tools and dependencies
 if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
+    # Adding non-free packges on Debian
+    echo -e "deb http://deb.debian.org/debian/ $(lsb_release -cs) main contrib non-free \ndeb http://deb.debian.org/debian/ $(lsb_release -cs)-updates main contrib non-free \ndeb http://deb.debian.org/debian-security/ $(lsb_release -cs)/updates main contrib non-free \ndeb http://deb.debian.org/debian/ $(lsb_release -cs)-backports main contrib non-free" >/etc/apt/sources.list
     apt-get-update-if-needed
 
     PACKAGE_LIST="apt-utils \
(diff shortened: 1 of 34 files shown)


script-library/common-debian.sh
```diff
@@ -107,7 +109,10 @@ if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
         sudo \
         ncdu \
         man-db \
-        strace"
+        strace
```
jkeech, line 112: I think you need a `\` here so that it's all the same command

```suggestion
        strace \
```

script-library/common-debian.sh
```diff
@@ -70,6 +70,8 @@ apt-get-update-if-needed()
 
 # Run install apt-utils to avoid debconf warning then verify presence of other common developer tools and dependencies
 if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
+    # Adding non-free packges on Debian
+    echo -e "deb http://deb.debian.org/debian/ $(lsb_release -cs) main contrib non-free \ndeb http://deb.debian.org/debian/ $(lsb_release -cs)-updates main contrib non-free \ndeb http://deb.debian.org/debian-security/ $(lsb_release -cs)/updates main contrib non-free \ndeb http://deb.debian.org/debian/ $(lsb_release -cs)-backports main contrib non-free" >/etc/apt/sources.list
```
jkeech, line 74: The `$(lsb_release -cs)` command is executed multiple times here. It might be nice to execute it once before running this, capture the output to a variable, and then use that variable here.

script-library/common-debian.sh
```diff
@@ -70,6 +70,8 @@ apt-get-update-if-needed()
 
 # Run install apt-utils to avoid debconf warning then verify presence of other common developer tools and dependencies
 if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
+    # Adding non-free packges on Debian
```
jkeech, line 73: Can you explain why we need to do this in the comment (needed for the manages packages)?

script-library/common-debian.sh
samruddhikhandale: Done.

script-library/common-debian.sh
samruddhikhandale: Added a variable `CODENAME`

script-library/common-debian.sh
samruddhikhandale: Explained in the comment

containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
```diff
@@ -70,6 +70,9 @@ apt-get-update-if-needed()
 
 # Run install apt-utils to avoid debconf warning then verify presence of other common developer tools and dependencies
 if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
+    # Needed for adding manpages-posix and manpages-posix-dev which are non-free packages in Debian
+    CODENAME=$(lsb_release -cs)
+    echo -e "deb http://deb.debian.org/debian/ $(CODENAME) main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-updates main contrib non-free \ndeb http://deb.debian.org/debian-security/ $(CODENAME)/updates main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-backports main contrib non-free" >/etc/apt/sources.list
```
Chuxel, line 75: @samruddhikhandale This can will actually break Ubuntu, so the addition of the Debian non-free sources should first verify that the distro is actually Debian before doing this.  (You can use `if [ "$(lsb_release -is)" = "Debian" ]; then`). Ubuntu does not appear to put `manpages-posix` in a separate package feed, so it works by default and you can skip this.

 You can test this using the "ubuntu" and "debian" definitions.
 
 You should probably also **append** to `/etc/apt/sources.list` rather than replace just to be safe (use `>>` not `>`).  This will probably remove the existing "free" sources which is not what you want.

containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
```diff
@@ -107,7 +110,10 @@ if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
         sudo \
         ncdu \
         man-db \
-        strace"
+        strace \
+        manpages-dev \
```
Chuxel, line 114: Maybe add "manpages" as well?  This should get pulled in by the others, but worth calling it out explicitly IMHO.

Chuxel (commented):
Generally looks right, but there's a few key things here since this script can be run for Ubuntu and Debian... this is one case where there's some subtle differences that are important.

containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
Chuxel: Update: The Debian only comment still applies.. but I realized the append is not right since this is just adding to the existing values.

samruddhikhandale: Added the condition and tested with both definitions, works fine

image
image

containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
samruddhikhandale: Added !

Chuxel (commented):
I got pulled into meetings but I hit a couple of issues when trying the "debian" definition.  I'll type up more on how I think we can resolve it with `sed` in a bit. (**EDIT**: I've since added the `sed` idea in the related inline comment.)

BTW - A simple way to copy scripts to everything is to run the following from the vscode-dev-containers repo.

```
yarn install
build/vscdc copy-library-scripts
```
This is what CI does and will copy whatever is in the script-library folder to all definitions that use the script.  You can then just open `containers/debian` or `containers/ubuntu` in Remote - Containers to test.

containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
```diff
@@ -70,6 +70,11 @@ apt-get-update-if-needed()
 
 # Run install apt-utils to avoid debconf warning then verify presence of other common developer tools and dependencies
 if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
+    # Needed for adding manpages-posix and manpages-posix-dev which are non-free packages in Debian
+    if [ "$(lsb_release -is)" = "Debian" ]; then
```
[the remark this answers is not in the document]
Chuxel: So we can get the codename out of /etc/os-release and then use sed to do a find-replace operation. This would work on any distro but would only take affect if debian was referenced.

```bash
CODENAME="$(cat /etc/os-release | grep -oE '^VERSION_CODENAME=.+$' | cut -d'=' -f2)"
sed -i "s/deb http:\/\/deb\.debian\.org\/debian ${CODENAME} main/deb http:\/\/deb\.debian\.org\/debian ${CODENAME} main contrib non-free/" /etc/apt/sources.list
sed -i "s/deb http:\/\/deb\.debian\.org\/debian ${CODENAME}-updates main/deb http:\/\/deb\.debian\.org\/debian ${CODENAME}-updates main contrib non-free/" /etc/apt/sources.list
sed -i "s/deb http:\/\/deb\.debian\.org\/debian-security ${CODENAME}\/updates main/deb http:\/\/deb\.debian\.org\/debian-security ${CODENAME}\/updates main contrib non-free/" /etc/apt/sources.list
sed -i "s/deb http:\/\/deb\.debian\.org\/debian ${CODENAME}-backports main/deb http:\/\/deb\.debian\.org\/debian ${CODENAME}-backports main contrib non-free/" /etc/apt/sources.list
```
We also may want to do the `deb-src` equivalent of these too... and I'm sure this could be refactored into a something that processes an array of lines to replace.

In either event, this will also prevent overwriting things if someone modified sources.list (even though that's not a good idea in general) and would not add backports unless its actually in the sources.list to begin with.

That said --- looking at what we have to do here, I'm thinking we may want to have a script argument that indicates whether non-free sources should be included and only include them if this is set to true... and it probably defaults to "false". `ADD_NON_FREE_PACKAGES=${7:-"false"}`

containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
```diff
@@ -70,6 +70,11 @@ apt-get-update-if-needed()
...
+        CODENAME=$(lsb_release -cs)
+        echo -e "deb http://deb.debian.org/debian/ $(CODENAME) main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-updates main contrib non-free \ndeb http://deb.debian.org/debian-security/ $(CODENAME)/updates main contrib non-free \ndeb http://deb.debian.org/debian/ $(CODENAME)-backports main contrib non-free" >/etc/apt/sources.list
+    fi
```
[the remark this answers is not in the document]
samruddhikhandale: I have made the changes :)

containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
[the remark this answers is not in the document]
samruddhikhandale: Sounds good. I have made all the discussed changes :)

Chuxel:
@samruddhikhandale This looks good!  

Last nit: I think we can drop the arguments for the definitions for the Oh My and the free packages flags.  That would show up in the VS Code "Add Dev Container Config" UX which is going to confuse the average developer.

The main place I want to be sensitive to is those using the script for their own images given there will be some that want to stick with Debian "free" packages only.

We can just do `/bin/bash /tmp/library-scripts/common-debian.sh "${INSTALL_ZSH}" "${USERNAME}" "${USER_UID}" "${USER_GID}" "${UPGRADE_PACKAGES}" true true` in ours.

samruddhikhandale:
> @samruddhikhandale This looks good!
> 
> Last nit: I think we can drop the arguments for the definitions for the Oh My and the free packages flags. That would show up in the VS Code "Add Dev Container Config" UX which is going to confuse the average developer.
> 
> The main place I want to be sensitive to is those using the script for their own images given there will be some that want to stick with Debian "free" packages only.
> 
> We can just do `/bin/bash /tmp/library-scripts/common-debian.sh "${INSTALL_ZSH}" "${USERNAME}" "${USER_UID}" "${USER_GID}" "${UPGRADE_PACKAGES}" true true` in ours.

I see. Made the changes.

samruddhikhandale:
Closed by mistake :)

Chuxel:
LGTM!  Merging

PR #785, opened 2021-03-10
Fix first run text due to terminal refresh, OMB notice

### LINKED ISSUE #775: oh-my-bash removal warning not displayed as expected.
I have [this in my dotfiles as my .bashrc](https://github.com/joshspicer/dotfiles/blob/master/.bashrc).  Specifically, the `source` line at the end causes the following in a (right now, dev) codespace.

image

I was suuuper confused until I saw #674 !

**Expected behavior:**  Displays removal warning of oh-my-bash.
**Actual behavior:** Show error opening the file and cds me to `/bin` (??)

Chuxel:
This PR:

1. Moves the first run text code into `common-debian.sh` so it can be used in any image or Codepsaces by injecting a text file at `/usr/local/etc/vscode-dev-containers/first-run-notice.txt`
2. Marks the first notice as shown asynchronusly after 10s to avoid the issue where VS Code refreshes the terminal after extension installs. 
3. Fixes #775 while I was in there

I also bumped the breakfix version on the images so they go with the next release.

### DIFF
--- a/container-templates/docker-compose/.devcontainer/library-scripts/common-debian.sh
+++ b/container-templates/docker-compose/.devcontainer/library-scripts/common-debian.sh
@@ -47,7 +47,7 @@ elif [ "${USERNAME}" = "none" ]; then
 fi
 
 # Load markers to see which steps have already run
-MARKER_FILE="/usr/local/etc/vscode-dev-containers/common"
+MARKER_FILE="/usr/local/etc/vscode-dev-containers/common-script-run"
 if [ -f "${MARKER_FILE}" ]; then
     echo "Marker file found:"
     cat "${MARKER_FILE}"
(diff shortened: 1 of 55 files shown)


containers/codespaces-linux/.devcontainer/setup-user.sh
```diff
@@ -35,10 +35,7 @@ sudo -u ${USERNAME} git config --global core.editor "/home/${USERNAME}/.local/bi
 rm -f /tmp/scripts/git-ed.sh
 
 # Add one time notice
-tee -a /etc/bash.bashrc /etc/zsh/zshrc << EOF
-if [ -t 1 ] && [ ! -f \$HOME/.config/vscode-dev-containers/first-run-notice ]; then
-  echo -e "Welcome to Codespaces! Note that the default Codespaces image is now Ubuntu 20.04-based!\nIf you need to use the old image (or a custom one) see https://aka.ms/ghcs-default-focal."
-  mkdir -p \$HOME/.config/vscode-dev-containers
-  touch \$HOME/.config/vscode-dev-containers/first-run-notice
-fi
+cat << 'EOF' > "/usr/local/etc/vscode-dev-containers/first-run-notice.txt"
```
jkeech, line 38: What's your thinking around how long we should offer this update help text? I was thinking about a month, but maybe we want to go longer? I'm just not sure how relevant this will be for new users who are onboarding to Codespaces.

jkeech (approved):
Looks good to me. Thanks for making this change! Left one question around the plan for the duration of update notices.

containers/codespaces-linux/.devcontainer/setup-user.sh
Chuxel: Given most users are going to be returning right now, I left it. Adoption takes time since existing Codespaces will still be on the old image, so I was more thinking 2-3 mos for a big update like this.

I did, however, also add text into codespaces-linux-stretch that calls it "depricated" and points the other direction now.  So I was thinking we could leave it a month with both and then remove this one.  That's one of the reasons I moved it into common code for future proofing.

@2percentsilk any thoughts from your perspective on when we should consider dropping the notice in the Focal image?

Chuxel: Let's merge this "as-is" for now to decouple the message conversation.  We can easily create a new PR just for that part.

container-templates/docker-compose/.devcontainer/library-scripts/common-debian.sh
```diff
@@ -186,9 +186,20 @@ else
 fi
 
 # .bashrc/.zshrc snippet
-RC_SNIPPET="$(cat << EOF
-export USER=\$(whoami)
-if [[ "\${PATH}" != *"\$HOME/.local/bin"* ]]; then export PATH="\${PATH}:\$HOME/.local/bin"; fi
+RC_SNIPPET="$(cat << 'EOF'
+
+if [ -z "${USER}"]; then export USER=$(whoami); fi
```
enihsyou, line 191: A missing space! I've got ```bash: [: missing `]'``` when execute `bash` in the container.

Change `[ -z "${USER}"]` into `[ -z "${USER}" ]` will fix the issue.

Chuxel: Thanks! Fixed.

PR #789, opened 2021-03-12
[codespaces] bump oryx 20210308.3

joshspicer:
Bumps oryx to latest vso-focal build.  Also changes how we symlink dotnet to dynamically accept the versions included with oryx.

Tested with vscode dev containers locally.

Have not validated that this oryx build fixes the python issues we are seeing.  I think the easiest way to validate _that_ is to get this merged and use the generated dev image to validate - right @Chuxel ?


Chuxel:
@joshspicer You can also just copy the `.devcontainer` folder into a repo, update devcontainer.json to point to "base.Dockerfile", and open it in a codespace.  It takes a long time because of how huge the image is, but it works. The devcontianer.json calls `oryx build` as postCreateCommand... you can also call it manually of course.

joshspicer:
ah yes, running the post-create manually makes sense :)

Chuxel:
@joshspicer LGTM! Should we merge it or do you want to do more checking?

Chuxel:
I'm going to merge this in since it looks good and I need to push an update to fix a bug anyway.

PR #795, opened 2021-03-19
Update welcome text for codespaces-linux

Chuxel:
Re: https://github.com/github/codespaces/issues/2511

image

The text is now externalized in a file, so anyone can tweak or make suggestions in the PR.  Right now the aka.ms link points to the codepsaces-linux definition since I don't think there's anywhere else. We can update the link if we get something into docs.

/cc: @2percentsilk @bailey @matthewisabel @jkeech


2percentsilk:
Looks awesome! Super excited to be able to see how this helps our new users 😄

PR #797, opened 2021-03-19
Fix #793 (bash prompt shows tag)

### LINKED ISSUE #793: Codespaces bash prompt doesn't display correct branch
Reported in https://github.com/github/feedback/discussions/3563

The Codespaces bash prompt includes the name of the current branch. This is currently calculated via `git describe --contains --all HEAD` which finds all branches/tags that point to the HEAD commit. A better approach to look at the current checked out ref might be `git rev-parse --abbrev-ref HEAD`.

@chuxel, @joshspicer are you aware of any downsides of switching to the rev-parse command for the prompt?

https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/common-debian.sh#L257

Chuxel:
This is a simple one-liner fix as described in #793. 

@matthewisabel is interested in an updated image for #795, so I got this one in quick while I was at it.

### DIFF
--- a/script-library/common-debian.sh
+++ b/script-library/common-debian.sh
@@ -254,7 +254,7 @@ __bash_prompt() {
     local userpart='`export XIT=$? \
         && [ ! -z "${GITHUB_USER}" ] && echo -n "\[\033[0;32m\]@${GITHUB_USER} " || echo -n "\[\033[0;32m\]\u " \
         && [ "$XIT" -ne "0" ] && echo -n "\[\033[1;31m\]➜" || echo -n "\[\033[0m\]➜"`'
-    local gitbranch='`export BRANCH=$(git describe --contains --all HEAD 2>/dev/null); \
+    local gitbranch='`export BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null || git describe --contains --all HEAD 2>/dev/null); \
         if [ "${BRANCH}" != "" ]; then \
             echo -n "\[\033[0;36m\](\[\033[1;31m\]${BRANCH}" \
             && if git ls-files --error-unmatch -m --directory --no-empty-directory -o --exclude-standard ":/*" > /dev/null 2>&1; then \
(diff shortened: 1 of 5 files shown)


joshspicer (approved):

script-library/common-debian.sh
```diff
@@ -254,7 +254,7 @@ __bash_prompt() {
     local userpart='`export XIT=$? \
         && [ ! -z "${GITHUB_USER}" ] && echo -n "\[\033[0;32m\]@${GITHUB_USER} " || echo -n "\[\033[0;32m\]\u " \
         && [ "$XIT" -ne "0" ] && echo -n "\[\033[1;31m\]➜" || echo -n "\[\033[0m\]➜"`'
-    local gitbranch='`export BRANCH=$(git describe --contains --all HEAD 2>/dev/null); \
+    local gitbranch='`export BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null || git describe --contains --all HEAD 2>/dev/null); \
```
jkeech, line 257: We need to check whether the output equals "HEAD" and then fallback. I'm not sure exactly how to do that in a one-liner, but something like this seems to work locally for me:

```
BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null)
if [ "${BRANCH}" == "HEAD" ]; then
    BRANCH=$(git describe --contains --all HEAD 2>/dev/null);
fi
```

jkeech (changes requested):

script-library/common-debian.sh
jkeech: A quick way to test is to checkout a branch, and then run `git checkout HEAD~1`. When running this script, the outcome should be `{branch}~1` instead of `HEAD`.

Chuxel: One liner for that would be
```
export BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null);  test "${BRANCH}" = "HEAD" || export BRANCH=$(git describe --contains --all HEAD 2>/dev/null);
```
... but that's less readable.  I'll update.

Chuxel: Ok - updated! Thanks for checking that.

jkeech (approved):

PR #810, opened 2021-04-06
[codespaces] supply PFS domain to allow domain through during ruby development

joshspicer:
Fixes: https://github.com/microsoft/vssaas-planning/issues/755

Adds in the PFS domain as an environment variable to unblock ruby on rails developers on the codespaces-linux image.

Rails docs:
https://edgeguides.rubyonrails.org/configuring.html#configuring-middleware

### DIFF
--- a/containers/codespaces-linux/.devcontainer/base.Dockerfile
+++ b/containers/codespaces-linux/.devcontainer/base.Dockerfile
@@ -19,6 +19,7 @@ ENV SHELL=/bin/bash \
     PIPX_HOME="/usr/local/py-utils" \
     PIPX_BIN_DIR="/usr/local/py-utils/bin" \
     RVM_PATH="/usr/local/rvm" \
+    RAILS_DEVELOPMENT_HOSTS=".githubpreview.dev" \ 
     GOROOT="/usr/local/go" \
     GOPATH="/go" \
     CARGO_HOME="/usr/local/cargo" \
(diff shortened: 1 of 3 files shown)


jkeech (approved):

Chuxel:
@joshspicer Should we also do this for the `ruby` image?  Also, does this affect local development at all? (I assume this is just "additional" domains beyond localhost.)

joshspicer:
@Chuxel - @jkeech and I discussed adding it to the `ruby` image, but decided against it (as to not impose our "codespace-ness" on other images). It would certainly be useful for any ruby dev using the image + codespaces though, so something for us to consider.

The docs say the env variable is "additional comma-separated hosts for development."

Chuxel:
@joshspicer As long as it's "additional", it might make sense to at least add it to user `Dockerfile` in the Ruby + Rails one if not the others. That's what people will actually get when running the command and would then allow them to change it with the codespaces one being a default example.

joshspicer:
[Here is a test ruby repo](https://github.com/joshspicer/rails-starter-e2e) (the one that @bratsche was having trouble on) with the kitchensink changes copied over.  Building it now to validate

joshspicer:
Validated that the change fixes ruby apps without any user intervention. The 2nd screenshot I ran `unset RAILS_DEVELOPMENT_HOSTS` to clear the allowed domain

Screen Shot 2021-04-07 at 1 02 55 PM
Screen Shot 2021-04-07 at 1 03 49 PM

joshspicer:
Validated that `localhost` doesn't break when making this change to the ruby image.  Tested with remote containers on local machine

image

joshspicer:
@Chuxel  - added in the change to Rails container - let me know what you think.

Chuxel:
LGTM!

PR #811, opened 2021-04-06
codespaces - utilize PROMPT_DIRTRIM to clean up PROMPT

### LINKED ISSUE #808: Shorten very long working directories in codespaces PROMPT
Would be great to limit the max length of the path with some ellipsis when it gets too long. For example, replacing this path:
image
with
`/workspaces/../.../VMAgent.Local.Tests` (or similar)

A lot of oh my zsh themes do this (and i love it)

_Originally posted by @joshspicer in https://github.com/microsoft/vscode-dev-containers/pull/807#discussion_r607257026_

joshspicer:
Set `PROMPT_DIRTRIM` for the default codespaces prompt for a more tidy look.

See the `...` below.

image

closes #808

### DIFF
--- a/script-library/common-debian.sh
+++ b/script-library/common-debian.sh
@@ -321,8 +321,10 @@ EOF
 if [ "${RC_SNIPPET_ALREADY_ADDED}" != "true" ]; then
     echo "${RC_SNIPPET}" >> /etc/bash.bashrc
     echo "${CODESPACES_BASH}" >> "${USER_RC_PATH}/.bashrc"
+    echo 'export PROMPT_DIRTRIM=2' >> "${USER_RC_PATH}/.bashrc"
     if [ "${USERNAME}" != "root" ]; then
         echo "${CODESPACES_BASH}" >> "/root/.bashrc"
+        echo 'export PROMPT_DIRTRIM=2' >> "/root/.bashrc"
     fi
     chown ${USERNAME}:${USERNAME} "${USER_RC_PATH}/.bashrc"
     RC_SNIPPET_ALREADY_ADDED="true"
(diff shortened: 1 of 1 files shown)


jkeech:
The change looks good, but it's definitely a personal preference. Not sure which way we want to go here. I actually prefer the full path without trimming so I get more context on where I am within the folder structure without having to run extra commands like `pwd`, but I know a lot of people like the trimming.  @Chuxel, thoughts?

joshspicer:
I think a middle group could be to bump up the `...` threshold from `2` to maybe `4` or `5` (although imo the smaller the better).

Here is another example, you can see the line itself needs to wrap because the path is so long (and this isn't even _that_ many layers deep!)

a

Chuxel:
@joshspicer If you switch to `zsh` what happens?   I can't remember the logic that zsh implements in this scenario... I seem to recall you can pick whether you see truncated or full and I don't remember which it is using. Regardless, their patterns are very well tested in the wild since it's the gold standard for prompt customization so if we subset it would be good to line it up. Either way we'd want to line up both themes I think as well.

joshspicer:
as expected `PROMPT_DIRTRIM` doesn't give it to use for free with `zsh`
image

joshspicer:
maybe i'm missing something, but how do I open a `zsh` shell?  I've running `zsh` and changing the shell manually with `chsh`...  Regardless i'm just dropped back into a bash shell.

image

Chuxel:
@joshspicer The theme is identical so you likely are in the zsh shell. :smile:

joshspicer:
:D My `echo $SHELL` did not change so i'm not sure

Chuxel:
@joshspicer A more foolproof way to tell is `echo $0`.  The SHELL doesn't always update particularly if it was set in image somewhere.

joshspicer:
ah ha! 
image

joshspicer:
For `zsh`, this is what I propose (previous commit):
image

joshspicer:
image

@jkeech @Chuxel - How do we feel about a DIRTRIM at 4 layers deep (This screenshot is bash, I implemented a slightly different, but similar shortening in zsh in the previous commit)

jkeech:
@joshspicer, four levels seems good to me 👍

Chuxel:
LGTM as well!

PR #812, opened 2021-04-07
Python Anaconda container - Run Jupyter Notebooks

diberry:


Chuxel:
You may want to use `postStartCommand` instead of `postCreateCommand`, put it in the background, and use nohup so it isn't killed when the command finishes.  

```
"postStartCommand": "nohup bash -c 'jupyter notebook --ip=0.0.0.0 --port=8888 --allow-root &'"
```

Otherwise eventually it will get killed because the script runs too long.  `postCreateCommand` only runs on create, not when 

If you wanted to, you could also actually move this up before the "Adding the definition to your folder" section since there's some tweaks to devcontainer.json.  Also as a FYI, there's now autoforwarding too.

FYI @bamurtaugh @2percentsilk

Chuxel (changes requested):
Left a few comments above.

diberry:
@Chuxel I didn't move it up because I'm not sure if it is a primary use case. Please review the other changes and merge if approved. #sign-off

Chuxel (approved):

PR #813, opened 2021-04-08
php-mariadb: install pdo_mysql driver

oliverpool:
The PHP+MariaDB image does not install any PHP extension.

This PR adds the pdo_mysql extension since the user will likely want to connect to the database in the app.

### DIFF
--- a/containers/php-mariadb/.devcontainer/Dockerfile
+++ b/containers/php-mariadb/.devcontainer/Dockerfile
@@ -24,3 +24,5 @@ RUN if [ "${INSTALL_NODE}" = "true" ]; then su vscode -c "umask 0002 && . /usr/l
 # [Optional] Uncomment this line to install global node packages.
 # RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
 
+# Install php-mysql driver
+RUN docker-php-ext-install mysqli pdo pdo_mysql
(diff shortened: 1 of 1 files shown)


bamurtaugh:
Would be great to get your thoughts @ThoolooExpress. Do users need pdo_mysql to more fully connect to the database in the app?

@oliverpool Would you be able to clarify:

> The PHP+MariaDB image does not install any PHP extension.

The definition currently installs php-debug and php-intellisense. Do you just mean it doesn't install the pdo_mysql component?

ThoolooExpress:
The extension is for PHP not VsCode.  PHP on its own can't connect to MySQL / MariaDB, it needs a driver extension to be installed.  I didn't add one when I created the Dockerfile initially, because there are multiple and I figured the one to choose is a project-specific detail.  However, it might make more sense just to install both of them, since the container size increase is negligible, and does make the user experience considerably easier just to have both ready to go.

oliverpool:
Yes, I meant extension for PHP (not vscode).

You are right, my command also installs `mysqli`, not just the `pdo` extension (as @ThoolooExpress mentions, the size overhead is negligible in comparison to the ease of use it brings).

2percentsilk:
So to confirm @ThoolooExpress, are you approving this change? If so, we can go ahead and merge it :)

ThoolooExpress:
Yes, but I would suggest moving the added line earlier in the Dockerfile to better take advantage of layer caching.  It should be before all the [Optional] instructions as those are far more likely to need to be changed in most use cases, and doing so would invalidate the cache for the newly added instructions unnecessarily.

oliverpool:
@ThoolooExpress done!

PR #815, opened 2021-04-12
Ensure getting started info factors in codespaces

Chuxel:
This PR updates the directions in this repository to better factor in Codespaces now that the "Reopen Container" flow is available. Previously customers were directed to either use Remote - Containers to add content or copy the `.devcontainer` folder manually. This leads to a poor experience both because its harder to do and because it drags additional unneeded files.

Lots of files are updated because of the number of definitions, but there's really three core updates repeated:

1. The [Main README](https://github.com/microsoft/vscode-dev-containers/blob/clantz/update-instructions/README.md)
2. [READMEs for community definitions](https://github.com/microsoft/vscode-dev-containers/tree/clantz/update-instructions/containers/deno) or definitions w/o a pre-built image
3. [READMEs for definitions with pre-built images that include info about fully customizing](https://github.com/microsoft/vscode-dev-containers/blob/clantz/update-instructions/containers/javascript-node/README.md#adding-the-definition-to-a-project-or-codespace)

//cc: @matthewisabel since he'd mentioned the READMEs being out-of-whack. @bailey as FYI.

Right now the "Test" section is kept with Remote - Containers since those assets aren't included in end user scenarios.  LMK if anyone feels strongly we should tweak those instructions as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,39 +8,25 @@
 </td>
 </tr></table>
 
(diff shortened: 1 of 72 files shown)


2percentsilk (approved):
LGTM. Thanks for dealing with this update Chuck!

bamurtaugh:
Agreed, thank you for this in-depth update, Chuck!

PR #817, opened 2021-04-14
install buildx when installing moby from microsoft repo

tonistiigi:
`docker-ce-cli` already contains buildx but in microsoft repos it is in a separate package.

I have no idea where most of these scripts are being used but the same code seems to copied in a lot of places and didn't want to leave them out of sync.

@cpuguy83

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>


Chuxel:
Thanks for catching that. Looks good.  Scripts that are updated in the "script-library" folder are automatically copied to the other locations via CI. They're in the other spots to ensure the builds work during development of the different containers.

FYI @joshspicer since we can rev the codespaces image to pick this change up.

PR #821, opened 2021-04-18
Place fish prompt in /etc/fish/functions/fish_prompt.fish to allow user override

technicalpickles:
The fish prompt is currently placed in `/etc/fish/conf.d/fish_prompt.fish`. Normally, `fish_prompt.fish` is expected to contain an function and is put in a `functions` directory, either `/etc/fish/functions` by default, or overridden by a user in `~/.config/fish/function`. Fish will then autoload the function when it's needed.

I'm not 100% sure of the specific load order, but it seems like having the `function fish_prompt` in `/etc/fish/conf.d/fish_prompt.fish` means that a user's custom fish_prompt.fish is never loaded. This can be confirmed by removing `/etc/fish/functions/fish_prompt.fish` and seeing your custom fish_prompt.fish loaded as expected.


technicalpickles:
Marked as a draft. It looks like the codespaces that's live right now is behind master 😅  Trying to figure out if it is still a problem or not in master.

joshspicer:
Thank you for this! I don't think we have any regular `fish` users as reviewers for this image :)

joshspicer (approved):

PR #822, opened 2021-04-19
[codespaces] Add libreadline-dev for building ruby

technicalpickles:
rbenv fails to install rubies without libreadline-dev.


joshspicer:
Looks good to me!

Could you add this change to https://github.com/microsoft/vscode-dev-containers/blob/master/script-library/ruby-debian.sh - this is the "centralized" source for this script. The file you edited is just a copy of this file (and your change to this mirror would eventually get overwritten by CI)

technicalpickles:
@joshspicer thanks! It took me another try to find the right ruby-debian.sh. Do I need to worry about those copies being overwritten, as in revert those changes? Or should it be alright to let CI handle it?

joshspicer:
This looks good now, thanks!

No need to revert, the scripts in container's directory will get overwritten regardless of its content.

joshspicer (approved):

PR #823, opened 2021-04-19
update workspaceMount and workspaceFolder to reflect what codespaces …

joshspicer:
Update workspaceMount and workspaceFolder to reflect what codespaces … does internally

### DIFF
--- a/containers/codespaces-linux/.devcontainer/devcontainer.json
+++ b/containers/codespaces-linux/.devcontainer/devcontainer.json
@@ -28,8 +28,8 @@
 	},
 	"remoteUser": "codespace",
 	"overrideCommand": false,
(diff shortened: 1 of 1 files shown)


Chuxel:
@joshspicer Actually we can probably just remove this now.  That was there because Coddespaces differed from remote containers but that is no longer the case.

Also, as a FYI `consistency=cached` is important to ensure decent perf on macOS when running locally. It's ignored for Linux (and WSL for Windows). The default consistency is very conservative. We default Remote - Containers to using it `cached`.  [More info.](https://code.visualstudio.com/docs/remote/containers-advanced#_update-the-mount-consistency-to-delegated-for-macos)

PR #828, opened 2021-04-23
Add version metadata, command into pre-built images, do not overwrite existing history files

Chuxel:
This PR adds build system capabilities to help developers understand what version of an image they are using - both for triage purposes and just their own information.  It:

1. Adds metadata and a `devcontainer-info` command inside containers that surfaces the injected information.
2. Adds labels to the image to help identify before you get into the image. (e.g. if you download 1-linux you can see in a label what the real version is)
3. Avoids accidently overwriting the "release" value in the history files we write on release - this is essentially a bug, but could cause further confusion.

Most of the PR is actually then adding a stub `meta.env` file into various definitions that is copied in. The build system generates the real one at build time and the common-*.sh scripts optionally add the command to run them.

Example output of the command:

```
$ devcontainer-info

Development container image information

- Image version: 0.201.4
- Definition ID: debian
- Variant: buster
- Source code repository: https://github.com/microsoft/vscode-dev-containers/
- Source code release/branch: v0.167.0

More info: https://github.com/microsoft/vscode-dev-containers/tree/master/containers/debian/history/0.201.4.md
```

Then for scriptability:
```
$ devcontainer-info version
0.201.4
$ devcontainer-info release
v0.167.0
$ devcontainer-info content-url
https://github.com/microsoft/vscode-dev-containers/tree/master/containers/debian/history/0.201.4.md
```

### DIFF
--- a/build/.vscode/launch.json
+++ b/build/.vscode/launch.json
@@ -11,7 +11,7 @@
 			"program": "${workspaceFolder}/vscdc",
 			"args": [
 				"push",
-				"go",
+				"debian",
 				"--no-push",
 				"--page", "1",
 				"--page-total", "1",
@@ -42,14 +42,17 @@
 		}, {
 			"type": "node",
 			"request": "launch",
-			"name": "Run CG Manifest Script",
+			"name": "Run CG/Info Script",
 			"program": "${workspaceFolder}/vscdc",
 			"args": [
-				"cg",
-				"codespaces-linux-stretch",
+				"info",
+				"codespaces-linux",
 				"--no-build",
 				"--markdown",
-				"--release", "v0.166.2",
+				"--cg",
+				"--overwrite",
+				"--prune",
+				"--release", "master",
 				"--github-repo", "microsoft/vscode-dev-containers",
 				"--registry", "mcr.microsoft.com",
 				"--registryPath", "vscode/devcontainers"
--- a/containers/go/.devcontainer/library-scripts/meta.env
+++ b/containers/go/.devcontainer/library-scripts/meta.env
@@ -1,0 +1,1 @@
+VERSION='dev'
--- a/script-library/common-redhat.sh
+++ b/script-library/common-redhat.sh
@@ -319,6 +319,44 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     fi
 fi
 
+# Persist image metadata info, script if meta.env found in same directory
+META_INFO_SCRIPT="$(cat << 'EOF'
+#!/bin/sh
+. /usr/local/etc/vscode-dev-containers/meta.env
+
+# Minimal output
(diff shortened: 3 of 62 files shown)


joshspicer:
This is very cool/useful! Taking a look...

containers/go/.devcontainer/library-scripts/meta.env
```diff
@@ -0,0 +1 @@
+VERSION='dev'
```
joshspicer, line 1: Is this something that be put into a shared spot/created programatically, so in the future we don't have to change it in each directory (if it is already automatically created?)

joshspicer (approved):
lgtm!

script-library/common-redhat.sh
```diff
@@ -319,6 +319,44 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     fi
 fi
 
+# Persist image metadata info, script if meta.env found in same directory
+META_INFO_SCRIPT="$(cat << 'EOF'
```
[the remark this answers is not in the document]
Chuxel: @joshspicer To do that, we'd have to commit to making every change work in all three distributions: Alpine, Debian, RedHat.  The goal is that these scripts can be curl'd in so we don't want them to have dependencies.

build/.vscode/launch.json
```diff
@@ -11,7 +11,7 @@
 			"program": "${workspaceFolder}/vscdc",
 			"args": [
 				"push",
-				"go",
+				"debian",
```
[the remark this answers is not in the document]
Chuxel: This is just launch.json that I used for testing. I flip it between different definitions to test. In this case the last was debian. Could exclude the change if its a big deal.

containers/go/.devcontainer/library-scripts/meta.env
Chuxel: It actually is programmatically updated on build [here](https://github.com/microsoft/vscode-dev-containers/blob/86fc18bd7cdeee51db6284db039e3d6ecfcc3403/build/src/prep.js#L248) using a handlebars template. This is just a stub that is there to get the "COPY" statement to execute in all cases when you're working on the definition outside of the build system (e.g. open the folder using Remote - Containers).

`COPY` will bomb if it doesn't find any files to copy annoyingly even if you use `*`.  There's a similar trick with a `noop.txt` file in the anaconda and miniconda definitions. It's particularly important for definitions like `cpp` that do not have a library-scripts folder otherwise. Anything could be in this file really. I just added the dev version so you'd see something in the `devcontainer-info` command but technically it could be empty.

The actual build script injects a meta.env that overwrites this, so it's not used for any other purpose.

Chuxel: We could remove it everywhere from anywhere that already has a `library-scripts` folder, but I went for consistency.

PR #836, opened 2021-04-27
Fix 'official' typo in common-*.sh scripts

pellared:
Found thanks to https://github.com/client9/misspell/ used in https://github.com/open-telemetry/opentelemetry-dotnet/blob/6b7f2dd77cf9d37260a853fcc95f7b77e296065d/.github/workflows/sanitycheck.yml


Chuxel:
Ack - good catch.  Wonder if we should consider running something like that automatically.

PR #843, opened 2021-04-29
Prevent GitHub API call for install docker-compose

felipecrs:
The GitHub API has rate limits.


Chuxel:
Good suggestion. There's a couple other spots that could get updated as well.

PR #848, opened 2021-04-30
Set "userEnvProbe": "interactiveShell" by default

Chuxel:
This PR is a proposal to switch to `interactiveShell` mode by default for definitions here.  We hear enough about confusion that `~/.bashrc` and `~/.zshrc` are not getting fired for processes in VS Code that I think this makes sense to recommend despite a minimal perf it on startup.

We didn't do this initially since Codespaces didn't support the property, but I believe that's been resolved.

I should also add that the problem with "login" is that most images reset the environment which wipes out any `ENV` set variables. `common-debian.sh` has a script to fix this for anything set up until that point, but it seems like a bad starting point for those that are creating custom Dockerfiles.

//cc: @chrmarti @bamurtaugh @2percentsilk @joshspicer

### DIFF
--- a/container-templates/docker-compose/.devcontainer/devcontainer.json
+++ b/container-templates/docker-compose/.devcontainer/devcontainer.json
@@ -4,6 +4,9 @@
 	// Update the 'dockerComposeFile' list if you have more compose files or use different names.
 	"dockerComposeFile": "docker-compose.yml",
 
+	// Fire ~/.bashrc, ~/.zshrc before starting VS Code processes
+	"userEnvProbe": "interactiveShell",
+
 	// The 'service' property is the name of the service for the container that VS Code should
 	// use. Update this value and .devcontainer/docker-compose.yml to the real service name.
 	"service": "app",
--- a/container-templates/dockerfile/.devcontainer/devcontainer.json
+++ b/container-templates/dockerfile/.devcontainer/devcontainer.json
@@ -8,6 +8,9 @@
 		"args": { "VARIANT": "buster" }
 	},
 
+	// Fire ~/.bashrc, ~/.zshrc before starting VS Code processes
+	"userEnvProbe": "interactiveShell",
+
 	// Set *default* container specific settings.json values on container create.
 	"settings": { 
 		"terminal.integrated.shell.linux": null
--- a/container-templates/image/.devcontainer/devcontainer.json
+++ b/container-templates/image/.devcontainer/devcontainer.json
@@ -4,6 +4,9 @@
 	// Update the 'image' property with your Docker image name.
 	"image": "mcr.microsoft.com/vscode/devcontainers/base:debian-10",
 
+	// Fire ~/.bashrc, ~/.zshrc before starting VS Code processes
+	"userEnvProbe": "interactiveShell",
+
 	// Set *default* container specific settings.json values on container create.
 	"settings": { 
 		// If you are using an Alpine-based image, change this to /bin/ash
--- a/containers/alpine/.devcontainer/devcontainer.json
+++ b/containers/alpine/.devcontainer/devcontainer.json
@@ -5,7 +5,8 @@
 		// Update 'VARIANT' to pick an Alpine version: 3.10, 3.11, 3.12, 3.13
 		"args": { "VARIANT": "3.13" }
 	},
-	
+	"userEnvProbe": "interactiveShell", // Fire ~/.bashrc, ~/.zshrc before starting VS Code processes
+
 	// Set *default* container specific settings.json values on container create. 
 	"settings": {
 		"terminal.integrated.shell.linux": "/bin/bash"
--- a/containers/azure-ansible/.devcontainer/devcontainer.json
+++ b/containers/azure-ansible/.devcontainer/devcontainer.json
@@ -8,6 +8,7 @@
 			"INSTALL_NODE": "true"
 		}
 	},
+	"userEnvProbe": "interactiveShell", // Fire ~/.bashrc, ~/.zshrc before starting VS Code processes
 	"mounts": [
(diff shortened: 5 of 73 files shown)


bamurtaugh:
This change makes sense to me; I was actually just looking into the history of https://github.com/microsoft/vscode-remote-release/issues/3224 today, so this feels pretty timely. As you mention Codespaces should support the property too, it sounds like a change that could help quite a few users in both Remote-Containers and Codespaces.

chrmarti:
Note that VS Code uses an interactive login shell. Using an interactive non-login shell will result in an environment that has never run the login startup scripts, not sure that will always have the expected result.

I think we could try to merge the container's PATH with the PATH we get from the environment probe to address the lost PATH entries issue. That could be done in Remote-Containers (in TypeScript, no shell scripting involved).

Whatever we decide on, we could do in Remote-Containers by default to avoid having to add it to every devcontainer.json.

Chuxel:
@chrmarti @bamurtaugh Generally it makes sense to do one of these to me. There's too much confusion about setting variables right now and it adds complexity for the scripts we've created.  The main thing we need to ensure if we choose login is that we get an environment that isn't wiped clean and values from `ENV` are no longer available. This will be particularly important if we change the default (which generally I'm okay with doing since I think it's more in-line with what customers expect).

jkeech:
@Chuxel, will this change have any impact on the first run welcome message?

Chuxel:
@jkeech That checks for a terminal using `if [-t 1]`, so it shouldn't unless @chrmarti's code passes in `-t` to the docker command. I didn't rely on interactive or login for exactly this reason.

Now that said, @chrmarti - I actually do see some errors caused by nvs which is a function.  Honestly we probably could remove nvs given we have nvm.  Open this branch in Remote - Containers: https://github.com/Chuxel/empty/tree/userEnvProbe/test-project

You'll see this in a terminal:
```
bash: nvs: line 1: syntax error: unexpected end of file
bash: error importing function definition for `nvs'
bash: nvsudo: line 1: syntax error: unexpected end of file
bash: error importing function definition for `nvsudo'
👋 Welcome to Codespaces! You are on our default image. It includes runtimes and tools for: 
   - Python 
   - Node.js
   - Docker
   - And more! To see the full list, go here: https://aka.ms/ghcs-default-image

🔍 To explore VS Code to its fullest, search using the Command Palette (Cmd/Ctrl + Shift + P or F1).

📝 Edit away, run your app as usual, and we'll automatically make it available for you to access.

bash: nvsudo: line 1: syntax error: unexpected end of file
bash: error importing function definition for `nvsudo'
bash: nvs: line 1: syntax error: unexpected end of file
bash: error importing function definition for `nvs'
bash: nvsudo: line 1: syntax error: unexpected end of file
bash: error importing function definition for `nvsudo'
bash: nvs: line 1: syntax error: unexpected end of file
bash: error importing function definition for `nvs'
bash: nvsudo: line 1: syntax error: unexpected end of file
bash: error importing function definition for `nvsudo'
bash: nvs: line 1: syntax error: unexpected end of file
bash: error importing function definition for `nvs'
```

chrmarti:
@Chuxel That looks like https://github.com/microsoft/vscode-remote-release/issues/4848.

`"userEnvProbe"` should work in Codespaces when the agent uses the CLI to start the VS Code server. Not sure that is enabled yet. /cc @edgonmsft

bamurtaugh:
> Whatever we decide on, we could do in Remote-Containers by default to avoid having to add it to every devcontainer.json.

I agree this sounds like the more scalable option - folks wouldn't have to start from one of the definitions here or be aware of the userEnvProbe property to get a functional experience.

Chuxel:
@jkeech @edgonmsft I had thought this was already done in Codespaces. Let us know if that's not the case (or there's a bug).  Right now the docs show it as supported as well.  I didn't see an issue open for it.

@bamurtaugh @chrmarti Sounds like we'd need to fix microsoft/vscode-remote-release#4848 first since this affects kitchen sink.

jkeech:
> `"userEnvProbe"` should work in Codespaces when the agent uses the CLI to start the VS Code server. Not sure that is enabled yet. /cc @edgonmsft

> @jkeech @edgonmsft I had thought this was already done in Codespaces. Let us know if that's not the case (or there's a bug). Right now the docs show it as supported as well. I didn't see an issue open for it.

Should be rolling out this week! We've been testing internally and getting to parity with the previous implementation of the VS Code server startup logic before switching over. The last known issues are fixed now, so unless testing turns up anything else, we'll start rolling it out for all users soon.

Chuxel:
@jkeech @chrmarti One thing I did check is that passing `-t` (terminal) into the docker CLI does in fact trip the first run notice logic if the process lasts more than 10 seconds.  Passing in `-i` (interactive) does not.  If the process runs less than 10s and terminates, the sub-shell with the "sleep" in it will also terminate. 

I guess a related question is whether there's any way from the environment to reliably check whether we're in a VS Code terminal window.  Is `TERM_PROGRAM=vscode` set by VS Code in that way? If so, we could check for that actually.

Also, just to check - for Codespaces, I assume the shell that would be used to do the userEnvProbe is the one in VS Code settings?  Or is it the default shell whatever that happens to be?

chrmarti:
We use the user's default shell configured in `SHELL` if that is set or `/etc/passwd`.

chrmarti:
Fixed microsoft/vscode-remote-release#4848 (not published yet).

Chuxel:
Okay, I think at this point we've come to the conclusion we should update defaults after some fixes here, so I'm going to close this PR. 

I'll submit a different one to check to make sure `TERM_PROGRAM` is set for the first run notice to ensure we do not accidently fire it outside of that context.  I noticed when I used docker -it, that value is not there. The only downside here will be that if someone SSH's in, they won't see it, but I think that is ok for now.   We can wire it into the actual welcome message in that case.

PR #856, opened 2021-05-07
Check for TERM_PROGRAM variable for notice

Chuxel:
Related to #848. This updates the first run notice check to verify that the TERM_PROGRAM is set for scenarios when the `userEnvProbe` setting will cause bashrc to be sourced.

//cc: @chrmarti @bamurtaugh @2percentsilk @joshspicer

### DIFF
--- a/script-library/common-alpine.sh
+++ b/script-library/common-alpine.sh
@@ -157,7 +157,7 @@ if [ -z "${USER}" ]; then export USER=$(whoami); fi
 if [[ "${PATH}" != *"$HOME/.local/bin"* ]]; then export PATH="${PATH}:$HOME/.local/bin"; fi
 
 # Display optional first run image specific notice if configured and terminal is interactive
-if [ -t 1 ] && [ ! -z "${TERM_PROGRAM}" ] && [ ! -f "$HOME/.config/vscode-dev-containers/first-run-notice-already-displayed" ]; then
+if [ -t 1 ] && [[ "${TERM_PROGRAM}" = "vscode" || "${TERM_PROGRAM}" = "codespaces" ]] && [ ! -f "$HOME/.config/vscode-dev-containers/first-run-notice-already-displayed" ]; then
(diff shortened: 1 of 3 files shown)


chrmarti:
We run the probe without `-t`, so checking for `[ -t 1 ]` might actually be sufficient.

Chuxel:
@chrmarti Are any of the start up steps run with `-t` or can we assume the terminal is the first shell that would have it?

Tyriar:
This is the right way, it should check that the value is `vscode` though

Chuxel:
> This is the right way, it should check that the value is vscode though

Yeah, I checked for it being set since it's not set by default in container images and conceivably another terminal program could ultimately be used in the Codespaces case over time (via SSH for example).  I could probably check for `vscode` or another value like `codespaces-terminal`  as the terminal program to future proof a bit.

@jkeech - Any preferences on what the second terminal value we check for here should be thinking forward? It would be something we could alias that could be set for any scenarios over time where terminal is accessed via something other than the VS Code internal terminal (which can technically happen now with the ssh script). Right now `TERM_PROGRAM` is not set when SSH'ing in.

jkeech:
> @jkeech - Any preferences on what the second terminal value we check for here should be thinking forward? It would be something we could alias that could be set for any scenarios over time where terminal is accessed via something other than the VS Code internal terminal. Right now `TERM_PROGRAM` is not set when SSH'ing in.

I don't have a strong preference. If we use `vscode` for VS Code integrated terminals, then perhaps just `codespaces` for any external terminals that we might open through other codespaces clients in the future? It's sort of hypothetical right now.



> @chrmarti Are any of the start up steps run with `-t` or can we assume the terminal is the first shell that would have it?

I thought that `postCreateCommand` and others are launched with `-t`, at least in Codespaces, in order to get control character support for the output of those commands. If that's true, we can't rely on the current logic since there's no guarantee the user connects within 10s of the `postCreateCommand` starting. Checking the `TERM_PROGRAM` seems like a good additional check to add.

PR #859, opened 2021-05-13
Deprecate Azure Blockchain definition

Chuxel:
This PR deprecates the Azure Blockchain definition given it will be shut down in Sept 2021. The README includes migration details and a pointer to the last version of the definition.

https://www.onmsft.com/news/microsoft-announces-shutdown-of-its-azure-blockchain-service


Chuxel:
//cc: @cmendible as FYI

cmendible:
Is the pluging also being deprecated? If not the container definition is valid for OSS Ethereum development.

Other option would be adding the Solidity plugin and renaming.

Thoughts?

Chuxel:
@cmendible That I don't know.  Mainly didn't want to have a service that is being deprecated in the list in VS Code.  Right now the extension is looks like it was renamed to "Blockchain Development Kit for Ethereum".

Mainly didn't want anything encouraging use of the deprecated service, so certainly renaming it is an option if you think that makes sense.

cmendible:
@Chuxel I also saw this today: https://azure.microsoft.com/en-us/updates/action-required-migrate-your-azure-blockchain-service-data-by-10-september-2021 so we are now partnering with ConsenSys and offering the Quorum Blockchain Service.

I think renaming from **Azure Blockchain (Community)** to: **Blockchain Development Kit for Ethereum (Community)** would make sense.

Then also add a PR to update the plugin dependencies.

Issue #838, opened 2021-04-27
Linux shell defined in *devcontainer.json* will be not applied


kirillsafin:
- VSCode Version: 1.55.2
- Local OS Version: Debian GNU/Linux bullseye/sid
- Name of Dev Container Definition with Issue: debian:testing-slim

Steps to Reproduce:

1. In VS Code on my device I use zsh as `terminal.integrated.shell.linux`,
2. Create devcontainer.json. for running container
2. Change the autocreated `"terminal.integrated.shell.linux": "zsh"` to `"terminal.integrated.shell.linux": "bash",`,
3. After attaching to container and opening terminal I get the message **The terminal process failed to launch: Path to shell executable "zsh" does not exist.**

I have already tryied to restart the container and reopen vs code. But it seems that this setting will not overwritten on attaching to container.

Chuxel:
@kirillsafin The base Debian image likely does not have zsh installed.  If it does, try adding the full path: /usr/bin/zsh or /bin/zsh

kirillsafin:
> @kirillsafin The base Debian image likely does not have zsh installed. If it does, try adding the full path: `/usr/bin/zsh` or `/bin/zsh`. Most definitions in this repository run `common-debian.sh` from the `script-library` which includes zsh and Oh My Zsh! installation options.

That's true, I guess there are not so much images which have zsh installed. I meant that in my VS Code settings on my physical device I have set the default linux shell to zsh `"terminal.integrated.shell.linux": "zsh"`. If I select **Remote-Containers: Add Development Container Configuration Files ...** from Command Palette there will be **devcontainer.json** created. Also the setting ``"terminal.integrated.shell.linux":` will be set and the value will be taken from my global settings for vs code, so the line `"terminal.integrated.shell.linux": "zsh"` will be automatically generated in **devcontainer.json**.
I thought I can change it for example to `"terminal.integrated.shell.linux": "bash"` (because I know debian image has bash installed, If it would be alpine based image or so I would set it to `sh`), but after clicking on  *attach to container* vs code window appears running inside the attached container and it seems to me like it ignores this `"terminal.integrated.shell.linux": "bash"` line inside **devcontainer.json** and applies my global settings. For example the defined extensions inside **devcontainer.json** will be applied, so I thought I can override or add some settings within **devcontainer.json** which will be applied to vs code instance running in the container

Chuxel:
[...]

PR #860, opened 2021-05-13
Remove pre-set terminal setting once 1.57 is released

Chuxel:
This PR removes the `"terminal.integrated.shell.linux"` setting that was in place due to the fact that shells like `zsh` are often not in containers. Previously in this scenario, a user setting would cause the terminal not to appear.  

The new terminal "profile" capability in VS Code solves this by falling back to either `$SHELL` or the shell in `/etc/passwd` if the specified profile fails as of the upcoming VS Coe 1.57 release (currently in insiders).

Related: https://github.com/microsoft/vscode/issues/123157, https://github.com/microsoft/vscode-dev-containers/issues/838


container-templates/dockerfile/.devcontainer/devcontainer.json
```diff
@@ -9,9 +9,7 @@
 	},
 
 	// Set *default* container specific settings.json values on container create.
-	"settings": { 
-		"terminal.integrated.shell.linux": null
```
Tyriar, line 13: What was this meant to accomplish? Overriding the user setting in case it's not installed?

Tyriar (commented):
It would be a good idea to verify this in containers before you push this

container-templates/dockerfile/.devcontainer/devcontainer.json
Chuxel: Correct. The fallback eliminates the need for it.

Chuxel:
> It would be a good idea to verify this in containers before you push this

Yep - this gives us an easy PR to test them. Given 1.56 just came out, we can hold a bit before merging.  I did see this working in VS Code Insiders 1.57, but 1.56.1 didn't seem to pick up updated the setting when I tried it.  (Its set in the PowerShell definition since the point of that one is to use powershell. 😄 ) So I assume it and the fallback aren't there yet.

Chuxel:
Merging in prep for 1.57 endgame week.  FYI @chrmarti

Issue #863, opened 2021-05-14
master => main


Chuxel:
As a notice, like other repositories we will be moving the default branch of this repository from master to main. This should be transparent for most things, but if you have a locally cloned copy, you'll need to run `git fetch` and then `git checkout main` if you need to interact with the default branch.

Chuxel:
This has been completed. To update your local copy of the vscode-dev-containers repository,  run the following commands:

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

PR #864, opened 2021-05-15
Update master to main

Chuxel:
Implements https://github.com/microsoft/vscode-dev-containers/issues/863. All other VS Code repositories have already switched.

//cc: @jkeech @joshspicer @chrmarti to see if there are any other impacts we need to consider before switching.

@bamurtaugh @2percentsilk We can update docs / links as it makes sense since master will auto redirect to main.

### DIFF
--- a/.github/workflows/patch.yml
+++ b/.github/workflows/patch.yml
@@ -3,7 +3,7 @@ name: Run Patches
 on: 
   workflow_dispatch:
   push:
-    branches: [master]
+    branches: [main]
     paths:
       - 'build/patch/**'
 
--- a/.github/workflows/push-and-package.yml
+++ b/.github/workflows/push-and-package.yml
@@ -161,4 +161,4 @@ jobs:
           -H "Accept: application/vnd.github.v3+json" \
           -H "Content-Type: application/json" \
           https://api.github.com/repos/${{ github.repository }}/actions/workflows/version-history.yml/dispatches \
-          --data "{\"ref\": \"master\", \"inputs\": {\"release\": \"${{ steps.get_tag_name.outputs.tag }}\", \"cg\": \"false\", \"push\": \"true\", \"overwrite\": \"false\"}"
+          --data "{\"ref\": \"main\", \"inputs\": {\"release\": \"${{ steps.get_tag_name.outputs.tag }}\", \"cg\": \"false\", \"push\": \"true\", \"overwrite\": \"false\"}"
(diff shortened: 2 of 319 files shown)


bamurtaugh:
Thanks for making this update! Once it's merged, I can make a pass through our VS Code docs (and readme's) for references of master and update to main, i.e. in the [Create a dev container doc](https://code.visualstudio.com/docs/remote/create-dev-container) I found: 

> You can see an example of `network_mode: service:db` in the [Node.js and Mongo DB example dev container](https://github.com/microsoft/vscode-dev-containers/blob/master/containers/javascript-node-mongo/.devcontainer/docker-compose.yml#L22).

We could also update aka.ms links. I found https://aka.ms/vscode-dev-containers/definitions redirects to https://github.com/microsoft/vscode-dev-containers/tree/master/containers. 

There are also https://aka.ms/vscode-dev-containers/common-redhat.sh, https://aka.ms/vscode-dev-containers/common-debian.sh, https://aka.ms/vscode-dev-containers/definitions/node, and https://aka.ms/vscode-dev-containers/common-alpine.sh, which redirect to URLs like https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/script-library/common-redhat.sh.

Chuxel:
Thanks @bamurtaugh for checking!  Sounds great.  @chrmarti I took a look at the remote containers extension code and there's a reference to `https://github.com/microsoft/vscode-dev-containers/tree/master/repository-containers/` to handle the `openDevContainerFile` for a built in repository container, but that's it.  That should auto-redirect, so I don't think it's blocking to make the change over.

Chuxel:
Ok @joshspicer is back later this week (I thought he was back today), so I'll wait to get confirmation this won't impact anything in the new automated import hooks he put in place. (Pretty sure it doesn't, but he's not here to verify / fix if it does.)

chrmarti:
Sounds good to me!

joshspicer:
Thanks for waiting! This should have no impact on our codespace image deployment pipeline 👍

joshspicer (approved):

Chuxel:
Merged!  As a FYI, you'll need to update your local copies of the source code as follows:

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

bamurtaugh:
Thanks for merging!

Looks like @gregvanl updated our docs already (thanks Greg!).

I just updated the aka.ms links to point to main. I also opened a PR on Remote-Containers to update the reference to https://github.com/microsoft/vscode-dev-containers/tree/master/repository-containers/ point to main.

Issue #809, opened 2021-04-06
Update .NET, codespaces images once .NET 5 update released for verisign CA issue


Chuxel:
Background is here: https://github.com/dotnet/announcements/issues/180

Update for the dotnet images should land on 4/6, so we can do a point release that includes #806 as well (which is already merged and ready for release).

/cc: @joshspicer @2percentsilk

Chuxel:
/cc: @TsuyoshiUshio @anthonychu as a FYI for the Azure Functions .NET images just in case this isn't on radar yet.

TsuyoshiUshio:
Thank you, @Chuxel ! Azure Functions are generally speaking, for .NetCore3.1 until 6 is coming, however, we also have .Net 5 implementation. We need to do it for that image.

Chuxel:
`mcr.microsoft.com/vscode/devcontainers/dotnet:5.0` image has been updated w/the 0.166.2 release - image with the fix will be  `dotnet:0.201.4-5.0` and later. 

We'll hope to get an oryx image soon to do the `universal` image (codespaces-linux) update in the next couple days.

anthonychu:
Thanks @Chuxel. As @TsuyoshiUshio mentioned, we don't have .NET 5 devcontainers in this repo yet, so I think we should be okay here for now.

@TsuyoshiUshio Yeah let's see if we can add a .NET 5 (isolated) devcontainer here. It's slightly more complicated than just using the 5.0 sdk image (it currently needs 3.1 sdk as well). Let's sync up offline.

PR #865, opened 2021-05-15
Update codespaces-linux to Oryx 20210514.2 

Chuxel:
Given this image now has Jekyll in it and .NET version handling, we can remove a few things. Xdebug is also now in the Oryx base image so we no longer need to install it separately.  However, to keep the image size neutral, this pulls out JDK 8.  JDK 11 is still present and is the current LTS.  SDKMAN is also installed to make it easy to add.

All other versions are updated including .NET to resolve https://github.com/microsoft/vscode-dev-containers/issues/809.


joshspicer (approved):
thank you!

Chuxel:
@joshspicer Merged - "dev" CI should finish up soon for further testing in the service.

PR #866, opened 2021-05-17
Fix apt sources rewrite to handle httpredir.

aebrahim:
The dart container fails to load because it fails to install manpages-posix and manpages-posix-dev. I checked and found this:

```
> cat /etc/apt-sources-list
deb http://httpredir.debian.org/debian buster main
deb http://httpredir.debian.org/debian buster-updates main
deb http://security.debian.org buster/updates main
```

The sed rewriting was only looking for deb.debian.org, and this change makes it also look for httpredir.debian.org.


Chuxel:
Strange - `httpredir` was deprecated way back in 2017 and is just a redirect https://lists.debian.org/debian-mirrors/2017/02/msg00000.html

That said, this makes sense given dart uses it.

aebrahim:
Thanks. Looks like the issue is in Google base images which was used to build the Dart container. Filed GoogleContainerTools/base-images-docker#6567 in case we want to roll this back if this gets merged.

PR #870, opened 2021-05-21
Adding the SWA CLI to devcontainer

aaronpowell:
@anthonychu I feel like this should be in the devcontainer by default


Chuxel:
I believe this ended up getting covered by #873. LMK if you disagree.

aaronpowell:
Yep it does

PR #871, opened 2021-05-24
Update sshd-debian.sh to restore environment variables for login shells

Chuxel:
While we are considering switching to using a login shell by default for Remote - Containers and Codespaces and merging the PATH variable automatically from VS Code (https://github.com/microsoft/vscode-remote-release/issues/5032), this does not work in the SSH case where a completely separate, empty login shell is created.

This PR modifies the `sshd-debian.sh` script to work around a number of issues that this causes.  In particular:

1. No environment variables set in the Dockerfile are present
2. No secrets from Codespaces are present
3. Ubuntu images include a hard coded PATH in `/etc/environment`
4. Other images that set no defaults should get the PATH from `/etc/login.defs`
5. However, most debian images include a hard coding of the default PATH in `/etc/profile`, but fortunately this path is the same as `/etc/login.defs` (see note [here](https://sources.debian.org/src/adduser/3.118/examples/adduser.local.conf.examples/profile/?hl=30#L30)).
6. ZSH does not source `/etc/profile` or `/etc/environment`, but has an entirely different value if no PATH is set based on `/etc/zshenv`.

It then uses the ENTRYPOINT to persist expected paths in each of these cases and then a script set in `/etc/profile.d` without overwriting existing environment variables and doing a substring match for the PATH.

Its setup to handle bash, sh/ash/dash, and zsh.

Given its complexity, there's a test script under `/script-library/test/sshd/test.sh` that runs it against a number of image configurations.

Finally, because of this new test, I had to move the location of the existing script library regression tests, so you'll notice some updates there as well.

//cc: @chrmarti @jkeech @joshspicer @matthewisabel @cwndrws


Chuxel:
FYI - Looks like I accidentally committed a few fixes outside of `/script-library/test/regression` as well. 

The following files with updates are unrelated to this change but still worth doing since they reported false positives when I did the a test run:

- containers/javascript-node-azurite/test-project/test.sh 
- containers/javascript-node-mongo/test-project/test.sh 
- containers/javascript-node-postgres/test-project/test.sh
- containers/javascript-node/test-project/test.sh 
- containers/kubernetes-helm-minikube/test-project/test.sh
- containers/kubernetes-helm/test-project/test.sh
- containers/typescript-node/test-project/test.sh

script-library/sshd-debian.sh
```diff
@@ -82,34 +77,194 @@ fi
 
 # Setup sshd
 mkdir -p /var/run/sshd
-sed -i 's@session\s*required\s*pam_loginuid.so@session optional pam_loginuid.so@g' /etc/pam.d/sshd
+sed -i 's/session\s*required\s*pam_loginuid\.so/ession optional pam_loginuid.so/g' /etc/pam.d/sshd
 sed -i 's/#*PermitRootLogin prohibit-password/PermitRootLogin yes/g' /etc/ssh/sshd_config
 sed -i -E "s/#*\s*Port\s+.+/Port ${SSHD_PORT}/g" /etc/ssh/sshd_config
 
-# Write out a script that can be referenced as an ENTRYPOINT to auto-start sshd
+# Script to store variables that exist at the time the ENTRPOINT is fired
```
eitsupi, line 84: typo? (ENTRYPOINT vs. ENTRPOINT)

Chuxel: Yep!  Thanks.

Chuxel:
@chrmarti Suggested an alternate model for managing the PATH problem that appears to work well. It updates PATH statements in `/etc/profile` and `/etc/zsh/zshenv` w/o a `$` in it to what the ENTRYPOINT encountered as the path, and then adds this to `/etc/environment` to covers scenarios where there's no hard coding. So far this seems to achieve the same thing with less code, though there is a bit of a risk of overwriting an intended hard-code.  That said, typically this would not be done in `/etc/profile` specifically but rather `/etc/profile.d` or `~/.profile`.

Updated the PR to try this and ran a test battery with a number of images where this appears to work.

The remainder of the env var handling is the same since we still need to restore other variables if they are not already set. More testing is needed to determine if we could safely update some of these variables in the environment file without impacting other features... inadvertent setting of things like HOME can be disastrous, so while we can make a best effort to filter - it may simply be safer to do this later.

Chuxel:
FYI - Holding this PR pending conversations in microsoft/vscode-remote-release#5032. We may be able to get dev container setup to handle much (if not all) of this and I'd rather not impact startup until we decided we have to. The script is, however, fully functional for anyone that needs it tactically.

Chuxel:
Okay, @chrmarti has merged in code for https://github.com/microsoft/vscode-remote-release/issues/4781 and https://github.com/microsoft/vscode-remote-release/issues/5032. That should eliminate PATH and env var hacks.  I can remove that code. 🎉

@jkeech Currently I suspect we should merge in code to do `jq` for the secrets and then do an update once https://github.com/microsoft/vssaas-planning/issues/4161 is complete.  Does that make sense?  The alternative would be to omit that part as a limitation until that is done.

jkeech:
> @jkeech Currently I suspect we should merge in code to do `jq` for the secrets and then do an update once [microsoft/vssaas-planning#4161](https://github.com/microsoft/vssaas-planning/issues/4161) is complete. Does that make sense? The alternative would be to omit that part as a limitation until that is done.

That sounds good to me. We don't have any plans to change that file or break anything related to this between now and when the `.env` file is added, so this seems like an OK temporary solution.

chrmarti:
Would it help if the CLI would support the `.env` file? Or is the agent taking care of this and only the SSH case needs fixing?

Chuxel:
@chrmarti Yeah that could make sense - @jkeech given discussions about refreshing the GITHUB_TOKEN automatically after it expires in a way that could change the token, is it safe to set all of this up front when the container starts going forward?

PR #873, opened 2021-05-26
Static Web Apps dev container update

sinedied:
Following the GA release of SWA, I updated a few thing in this dev container configuration:

- ~~Added Java (+ tools) installation, so all languages supported by Functions are now available.~~
- Moved NVM installation to Dockerfile, and added options to change the Node.js version
- Added installation of the new SWA CLI, and exposed port 4280 to match it
- Removed the live server extension (duplicate with SWA CLI)


Chuxel:
@anthonychu @@TsuyoshiUshio Does this look good to you?

anthonychu:
Thanks.

Agree with adding CLI and removing Live Server.

We don't support Java Functions in Static Web Apps right now. It could theoretically run as a separate function app but that's probably not something many will do.

This should be good if we remove Java.

sinedied:
My bad, I though that Java was supported since the Python/C# support was added! Any particular reason for Java not being supported, or is it just something that will come later?

I'll remove Java then to merge the CLI update.

containers/azure-static-web-apps/.devcontainer/Dockerfile
```diff
@@ -2,8 +2,22 @@
...
-ENV NVM_SYMLINK_CURRENT=true \
-    PATH=${NVM_DIR}/current/bin:${PATH}
-COPY library-scripts/node-debian.sh /tmp/library-scripts/
-RUN apt-get update && bash /tmp/library-scripts/node-debian.sh "${NVM_DIR}"
\ No newline at end of file
+# Copy library scripts to execute
+COPY library-scripts/*.sh library-scripts/*.env /tmp/library-scripts/
+
+# Install Node.js and Azure Static Web Apps CLI
+ARG NODE_VERSION="lts/*"
```
Chuxel, line 9: Does Azure Static Web Apps support Node 16 yet?  That's soon to be LTS.

If there's a static list of supported versions you could add this in the comment to let people pick:
```
# [Choice] Node.js version: 14, 12
```

anthonychu:
> Any particular reason for Java not being supported, or is it just something that will come later?

There's overhead in supporting new languages in the functions we manage on behalf of the customer. So far we haven't seen any interest in using Java functions in Static Web Apps.

containers/azure-static-web-apps/.devcontainer/Dockerfile
sinedied: What would you recommend then? I left `lts` as it was there initially, but also wondered about that.
Should we default to Node 12 with a little comment explanation here, and then update the version later when Node 14/16 is supported in SWA?

Chuxel: @anthonychu ? 👆 I can merge as "lts/*" if that makes sense... right now that's 14.

anthonychu: Yep. I think that'll be okay. Thanks.

Chuxel:
Sounds good!

PR #875, opened 2021-05-28
Create a container that uses the PyPy interpreter

Cyb3r-Jak3:
Creates a python container using the [PyPy](https://www.pypy.org/) interpreter.

### DIFF
--- a/containers/python-3-pypy/.devcontainer/Dockerfile
+++ b/containers/python-3-pypy/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-# [Choice] Pypy docker images. Choices are 2,3. Complete list of tags is avaiable at: https://hub.docker.com/_/pypy
+# [Choice] Pypy version: 2, 3
 ARG VARIANT="3"
 FROM pypy:${VARIANT}
 
--- a/containers/python-3-pypy/.devcontainer/Dockerfile
+++ b/containers/python-3-pypy/.devcontainer/Dockerfile
@@ -1,0 +1,23 @@
+# [Choice] Pypy docker images. Choices are 2,3. Complete list of tags is avaiable at: https://hub.docker.com/_/pypy
+ARG VARIANT="3"
+FROM pypy:${VARIANT}
+
+# Use the [Option] comment to specify true/false arguments that should appear in VS Code UX
+#
+# [Option] Install zsh
+ARG INSTALL_ZSH="true"
+# [Option] Upgrade OS packages to their latest versions
+ARG UPGRADE_PACKAGES="false"
+
+# Install needed packages and setup non-root user. Use a separate RUN statement to add your own dependencies.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+COPY library-scripts/*.sh library-scripts/*.env /tmp/library-scripts/
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && /bin/bash /tmp/library-scripts/common-debian.sh "${INSTALL_ZSH}" "${USERNAME}" "${USER_UID}" "${USER_GID}" "${UPGRADE_PACKAGES}" "true" "true"\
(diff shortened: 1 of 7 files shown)


Chuxel:
@brettcannon ?

brettcannon:
@Chuxel I guess so 😉 I have comments, but they seem to all relate to what was directly lifted from the Python container image. I guess the only real difference here is the use of `pypy:3` as a base image which I don't know anything about.

containers/python-3-pypy/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# [Choice] Pypy docker images. Choices are 2,3. Complete list of tags is avaiable at: https://hub.docker.com/_/pypy
```
Chuxel, line 1: @Cyb3r-Jak3 This is not a valid format for the [Choice] argument since this drives UX. A valid example would be:

```suggestion
# [Choice] Pypy version: 2, 3
```
...which would then allow users to pick when adding this definition from VS Code.

If you don't want to maintain a list, you can remove the `[Choice]` prefix.

Chuxel:
> @Chuxel I guess so 😉 I have comments, but they seem to all relate to what was directly lifted from the Python container image. I guess the only real difference here is the use of `pypy:3` as a base image which I don't know anything about.

@brettcannon Happy to take updates there too. The definition has been largely as-is for a while now so I'm sure there's tweaks.

containers/python-3-pypy/.devcontainer/Dockerfile
Cyb3r-Jak3: Fixed

containers/python-3-pypy/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+| Metadata                    | Value                                                                        |
+|---------------------------- | -----------------------------------------------------------------------------|
+| *Contributors*              | [Cyb3r Jak3](https://github.com/Cyb3r-Jak3)                                  |
+| *Categories*                | Community, Languages                                                         |
+| *Definition type*           | Dockerfile                                                                   |
+| *Works in Codespaces*       | Yes                                                                          |
+| *Container host OS support* | Linux, macOS, Windows                                                        |
+| *Container OS*              | Debian                                                                       |
+| *Languages, platforms*      | [Languages and platforms the container supports]                             |
```
Chuxel, line 15: You can add the languages here since this goes into a suggestion algorithm.

```suggestion
| *Languages, platforms*      | Python, PyPy                            |
```

containers/python-3-pypy/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,50 @@
...
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "uname -a",
+
+	// Uncomment to use Docker from inside the container. See https://aka.ms/vscode-remote/samples/docker-from-docker.
+	// "mounts": [ "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind" ],
+
+	// Uncomment when using a ptrace-based debugger like C++, Go, and Rust
+	// "runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],
+
```
Chuxel, line 47: You can remove these lines since they don't apply to Python.

```suggestion
```

containers/python-3-pypy/README.md
```diff
@@ -0,0 +1,52 @@
+# PyPy (Community)
+
+## Summary
+
+*Develop python applications using the [PyPy](https://www.pypy.org/) interpreter*
```
Chuxel, line 5: I don't believe links will render properly. Let's remove it from this spot since it drives VS Code UX.

```suggestion
*Develop python applications using the PyPy interpreter*
```

Chuxel (changes requested):
A few other suggested changes on a more detailed pass. Otherwise LGTM!

Chuxel:
Merged. Thanks!

PR #880, opened 2021-06-04
Remove login shell PATH workaround given this is supported in product

Chuxel:
Resolved by https://github.com/microsoft/vscode-remote-release/issues/4781


Chuxel:
FYI @chrmarti I think we should stagger this update until stable is out and we've confirmed everything has landed in Codespaces so we don't inadvertently break people with updated images on an older version of the extension/CLI.

Issue #884, opened 2021-06-06
Dart container wont build - unable to locate packages manpages-posix and manpages-posix-dev 


mwinters-stuff:
- VSCode Version: 1.56.2
- Local OS Version: Ubuntu 21.04
- Name of Dev Container Definition with Issue: Dart

Steps to Reproduce:

1. Create empty folder
2. Open in VSCode
3. Add Development Container - Dart
4. Open in container
5. Build fails, unable to locate packages manpages-posix and manpages-posix-dev

These files are listed in containers/dart/.devcontainer/library-scripts/common-debian.sh

Log file:
[remoteContainers-2021-06-06T23-27-18.418Z.log](https://github.com/microsoft/vscode-dev-containers/files/6605125/remoteContainers-2021-06-06T23-27-18.418Z.log)

I will see if I can do a PR

Chuxel:
These packages are needed for other distributions so we will need to detect and adapt.

Chuxel:
Hmm this does not repro for me.  Do you have an old version of this definition?  There was a period where there was a bug but that should be resolved.  There should be some sed statements in `library-scripts/common-debian.sh` that reference `deb|httpredir`rather than just `deb`.  Do you see that?  If not, grab the latest script [from here](https://github.com/microsoft/vscode-dev-containers/blob/main/containers/dart/.devcontainer/library-scripts/common-debian.sh) and retry to see if that fixes it for you.

```
    # Needed for adding manpages-posix and manpages-posix-dev which are non-free packages in Debian
    if [ "${ADD_NON_FREE_PACKAGES}" = "true" ]; then
        CODENAME="$(cat /etc/os-release | grep -oE '^VERSION_CODENAME=.+$' | cut -d'=' -f2)"
        sed -i -E "s/deb http:\/\/(deb|httpredir)\.debian\.org\/debian ${CODENAME} main/deb http:\/\/\1\.debian\.org\/debian ${CODENAME} main contrib non-free/" /etc/apt/sources.list
        sed -i -E "s/deb-src http:\/\/(deb|httredir)\.debian\.org\/debian ${CODENAME} main/deb http:\/\/\1\.debian\.org\/debian ${CODENAME} main contrib non-free/" /etc/apt/sources.list
[...]

PR #885, opened 2021-06-07
Issue #884 - the two manpages packages can not be found.

mwinters-stuff:
remove dependencies that cant be found.
These packages can not be found when creating the container.


mwinters-stuff:
#884

mwinters-stuff:
Not required, already fixed just had not made into master.

PR #891, opened 2021-06-09
Ensure Jupyter server starts, is forwarded

Chuxel:
This PR is a tweak to the Jupyter Data Science Notebooks definition that ensures Juypter server starts automatically and is forwaded locally so people can access it directly as neeeded w/o doing anything from the command line.

Look good @nathancarter ?

//cc @dynamicwebpaige


nathancarter:
Sure!  You're more familiar with all the details of the `devcontainer.json` than I am anyway, so I'll trust any way that you want to muck around in there. :)

Issue #878, opened 2021-05-31
Universal image 1.4 has an incomplete .bashrc file.


berkant:
- VSCode Version: Codespaces
- Local OS Version: Codespaces
- Name of Dev Container Definition with Issue: `mcr.microsoft.com/vscode/devcontainers/universal`

Steps to Reproduce:

1. Create a new codespace.

or

1. Include a `.devcontainer.json` with `image` field set to `mcr.microsoft.com/vscode/devcontainers/universal:1.4-focal`.

---------

`~/.bashrc` (1.4):
```bash
codespace ➜ /workspaces/codespaces-triage-1 (main) $ cat ~/.bashrc 

# Codespaces bash prompt theme
__bash_prompt() {
    local userpart='`export XIT=$? \
        && [ ! -z "${GITHUB_USER}" ] && echo -n "\[\033[0;32m\]@${GITHUB_USER} " || echo -n "\[\033[0;32m\]\u " \
        && [ "$XIT" -ne "0" ] && echo -n "\[\033[1;31m\]➜" || echo -n "\[\033[0m\]➜"`'
    local gitbranch='`\
        export BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null); \
        if [ "${BRANCH}" = "HEAD" ]; then \
            export BRANCH=$(git describe --contains --all HEAD 2>/dev/null); \
        fi; \
        if [ "${BRANCH}" != "" ]; then \
            echo -n "\[\033[0;36m\](\[\033[1;31m\]${BRANCH}" \
            && if git ls-files --error-unmatch -m --directory --no-empty-directory -o --exclude-standard ":/*" > /dev/null 2>&1; then \
                    echo -n " \[\033[1;33m\]✗"; \
            fi \
            && echo -n "\[\033[0;36m\]) "; \
        fi`'
    local lightblue='\[\033[1;34m\]'
    local removecolor='\[\033[0m\]'
    PS1="${userpart} ${lightblue}\w ${gitbranch}${removecolor}\$ "
    unset -f __bash_prompt
}
__bash_prompt
export PROMPT_DIRTRIM=4
export NVS_HOME="$HOME/.nvs"
[ -s "$NVS_HOME/nvs.sh" ] && . "$NVS_HOME/nvs.sh"
```

`~/.bashrc` (1.3):
```bash
codespace ➜ /workspaces/codespaces-triage-1 (main) $ cat ~/.bashrc 
[...]

PR #896, opened 2021-06-16
Restore bashrc and profile from skeleton file if file doesn't exist

joshspicer:
fixes: https://github.com/microsoft/vscode-dev-containers/issues/878

Pulls from the skeleton file in our `common-debian.sh` if the .bashrc and .profile do not exist (or are empty).


Chuxel:
LGTM!

Issue #558, opened 2020-09-15
Support ARM64 containers / images


Chuxel:
Currently the definitions in this repository are x86_64 based. Some scripts will likely need to be modified to support ARM64. In addition, pre-built definitions will need to be compiled for ARM64 after this has been done.

gfmatthews:
Hey y'all - I have a Surface Pro X and am interested in helping to get some of these containers supported on the ARM64 platform. I've seen that [javascript-node](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/javascript-node) builds correctly without modification already for ARM64 so for that container, the only thing missing is getting the pre-built containers updated. 

Is there a build pipeline or something similar somewhere that can be used to test this out? Happy to reach out internally (@microsoft) if that would be helpful.

jasonsummers:
Hi all, new to GitHub - apologies in advance for breaches of etiquette, please correct any mistakes I make.

I'm keen to help make progress in on this issue and plan to look at how multi-arch images can be used as well as cleaning up / automating debugging linux on ARM.  As a temporary solution I've managed to get a working .net core devcontainer working by building from the arm64 sdk image and using the scripts from this repo.

If work is already in progress on resolving this issue, please let me know if and how I can be of help.

philliphoff:
My shiny new M1 Mac arrived yesterday and I've already run into a `qemu` issue when building upon one of the base `amd64` images. It would be great to have `arm64` images to test.

tkreind:
I'm on an Apple M1 and although I totally understand this is beta and low priority, it would seem as simple as building and publishing an aarch64 container for each dev environment.

I got the dotnet devcontainer to work perfectly locally in docker simply by changing
`FROM mcr.microsoft.com/dotnet/sdk:5.0-focal`
to
`FROM mcr.microsoft.com/dotnet/sdk:5.0-buster-slim-arm64v8`
in the parent dockerfile.

jroes:
I've been hacking around with things here to get my m1 mac working with various things. Here's a recent [Dockerfile](https://github.com/jroes/streamlit-getting-started-m1/blob/main/.devcontainer/Dockerfile) I put together in case it's helpful to anyone. It seems to work okay so far!

chrmarti:
@Chuxel Maybe we should introduce an automatic build arg in Remote-Containers for the architecture that can be appended to the image name in `FROM` to avoid duplication?

jamesthurley:
[...]

PR #901, opened 2021-06-19
Updated some Debian scripts to support arm

ItalyPaleAle:
This PR updates **some** (not all) scripts in script-library to add fixes so they can be used to build arm64 containers too. Note that the PR doesn't add full arm64 support for the repo (see #558), but it does offer some help in that direction.

How this came to be: I wanted to be able to use a devcontainer on a M1 Mac. The base images that we currently publish don't support ARM, so I resorted to creating my own containers but I used the scripts in the script-library folder to make things easier for me.

### DIFF
--- a/script-library/azcli-debian.sh
+++ b/script-library/azcli-debian.sh
@@ -27,7 +27,7 @@ if ! dpkg -s apt-transport-https curl ca-certificates lsb-release > /dev/null 2>
 fi
 
 # Install the Azure CLI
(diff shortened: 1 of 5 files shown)


Chuxel:
There's actually a branch with all the scripts updated to support this along with more secure use of GPG keys for apt. I'm waiting on a M1 mac to arrive to allow me to test, but if you want to help verify, it is here:  https://github.com/microsoft/vscode-dev-containers/tree/clantz/apt-key-and-architecture

ItalyPaleAle:
@Chuxel I had missed that :) Looks like it already includes everything I did in this PR and then more. Closing this PR!

PR #902, opened 2021-06-19
Update NVM version

ItalyPaleAle:
NVM version can now be set via the NVM_VERSION variable in the script, so future updates should be easier


Chuxel:
Did a quick check just to make sure nothing broke unexpectedly with the update.  LGTM!

PR #903, opened 2021-06-21
Codespaces/git editor

joshspicer:
fixes https://github.com/microsoft/vssaas-planning/issues/4165

Sets the `$GIT_EDITOR` environment variable in the common-*.sh scripts when the vscode integrated terminal is detected, causing git operations (commit msgs, etc) to open up in the editor by default.  

This is an improvement on the previous method for several reasons:
- It previous config only applied to the kitchensink, not other VS Code devcontainer images/samples- 
- It caused issues for non-VS Code clients, like SSH connections
- It was set in the global git config, which will be overwritten by a lot of user's dotfiles in codespaces.

### DIFF
--- a/containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
+++ b/containers/codespaces-linux/.devcontainer/library-scripts/common-debian.sh
@@ -298,10 +298,16 @@ __bash_prompt() {
     PS1="${userpart} ${lightblue}\w ${gitbranch}${removecolor}\$ "
     unset -f __bash_prompt
 }
-__bash_prompt
 
+# Set the default git editor
+if  [ "${TERM_PROGRAM}" = "vscode" ]; then
+    $GIT_EDITOR="vscode"
+fi
+
+__bash_prompt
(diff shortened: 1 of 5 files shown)


Chuxel (commented):
You could move the snippet into `RC_SNIPPET` rather than having to do it twice.

Also, this would pull two "which" statements into shell startup. Did you check the impact of doing this here rather than in a command? I wouldn't expect it to be big, but there's also `type` that can be used that may or may not be faster. We're discussing switching the `userEnvProbe` default to `-li` so milliseconds count. :smile:

joshspicer:
Will move into RC_SNIPPET, good call. 

I did not measure the speed at all, do you think the best way to test that would be to simply do a `time . ~/.bashrc` in our shell with both approaches?

joshspicer:
image

Looks pretty negligible but which does appear slowest. I'll actually move to use `command -v <BINARY>` since we use that all over, and it doesn't produce any stderr

Chuxel:
LGTM!

PR #904, opened 2021-06-21
Add editorconfig as a recommended extension

### LINKED ISSUE #814: Use `.editorconfig` when editing with VSCode
This repository contains `.editorconfig` file, which is configured for indentation.
However, VS Code, which users who want to contribute to this project may be using, does not read the `.editorconfig` settings by default, so it is highly likely that you will unintentionally use the wrong indentation.

How about adding `.vscode/extensions.json` to this project to recommend the [`EditorConfig.EditorConfig` extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)?

2percentsilk:
Fixes #814 to recommend the EditorConfig extension when users are contributing to this repo

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,13 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"EditorConfig.EditorConfig",
+	],
+	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
+	"unwantedRecommendations": [
+		
+	]
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
 
 	// List of extensions which should be recommended for users of this workspace.
 	"recommendations": [
+		// We recommend this extension for contributors to assist with the correct styling and formatting of files in this repo.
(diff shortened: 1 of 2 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,13 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
```
bamurtaugh, line 2: Is there a reason we use `https://go.microsoft.com/fwlink/?LinkId=827846` instead of the URL it redirects to (which looks a bit cleaner) https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions?

bamurtaugh:
LGTM - tried it out, and I got the appropriate recommendation: 
image

One small thought - do we want to add a note so that users are also aware of this? i.e. I notice in the linked issue 814, the user mentions this helps with checking out and contributing to the repo. Maybe a quick comment in the new file or statement in https://github.com/microsoft/vscode-dev-containers/blob/main/CONTRIBUTING.md that this extension rec is tailored for this repo/helpful for getting the correct indentation?

.vscode/extensions.json
2percentsilk: It's a standard practice so that if we need to update the doc URL for recommended extensions, we can just update the Link versus every single copy of a devcontainer out there (past and present)

2percentsilk: A link like this stays evergreen with any URL changes - assuming we update it to a working link :)

bamurtaugh: Got it - thanks for the explanation :)

Issue #895, opened 2021-06-15
Use of DevContainers with GitHub Actions


joewood:
I've been trying to use the Microsoft vscode-dev-containers with Github actions over the past couple of days. It was surprisingly difficult. I'm wondering if parity between devcontainers and GitHub CI/CD is a goal.  

The primary issue that caused compatibility problems was the use of SDKMAN by the devcontainers Dockerfiles. The setup for SDKMAN relies on an interactive login shell (to set the PATH etc.). GitHub Actions use a non-interactive shell, with command by command against the created container. I had to resort to using a custom image with maven installed manually.

The feature request is for improved support in DevContainers for GitHub Actions Job Containers.  This may be possible with SDKMAN, but will require changes to the interactive login dependency.

Chuxel:
@joewood Can you be a bit more specific on which definition or image you were using and how you attempted to use it with Actions?

For example, SDKMAN is in the Java images, but the currently installed software is in the PATH already. Maven and Gradle are not in the images by default but are added by a Dockerfile which creates another image based on the options you select. [These would be in the PATH, however.](https://github.com/microsoft/vscode-dev-containers/blob/a8845747217d1f9c200a7834df1ab7df8efe1204/containers/java/.devcontainer/base.Dockerfile#L27)  As a result, I don't know that this is what is causing the problem you are running into.

That said, tools like SDKMAN, nvm, rvm, and more expect to be "sourced" before they are used. It is not possible to do this automatically in a non-login, non-interactive shell. What you can do, however, is put their target locations in the PATH.

Chuxel:
I did notice there's a path ordering problem in the `java:8` image. Is that what you used?

GitHub Actions also seems to override the entrypoint for the container, so that means the shell will be `sh` (really `dash`) rather than `bash` by default and you can't override it. 🙁 FWIW - This is how I'd set things up if you need the interactive shell.   You'll get a warning about Ioctl but you can ignore it.

```
name: Test

on: 
  workflow_dispatch:
  push:
    branches: [main]

jobs:
  test:
    runs-on: ubuntu-latest
    container:
      image: mcr.microsoft.com/vscode/devcontainers/java:11
      options: -u vscode
    steps:
    - name: Test
      run: |
        bash -i << 'BASHEOF'
        
        sdk version

        BASHEOF
```

Anything between BASHEOF will run in the interactive shell.  You can also run `bash -i your-script-here.sh` if you want to keep this a separate file.  

[...]

PR #905, opened 2021-06-23
Fix issue with path for SDKMAN installed tools

Chuxel:
As discussed in #895, the PATH in both the java and codespaces-linux definition is supposed to include the location that SDKMAN would install maven or gradle... specifically `${SDKMAN_DIR}/candidates/maven/current/bin` and `${SDKMAN_DIR}/candidates/gradle/current/bin`

Unfortunatley the `candidates` part was missing. As a result, the image works if you use it interactively since SDKMAN injects it, but any non-interactive interactions will not find either tool.

This fixes the paths.


testforstephen (approved):
LGTM

Issue #919, opened 2021-06-24
Pre-built Jekyll image


Chuxel:
Jekyll is used heavily throughout GitHub and generally in static web sites. We can pre-build a version of this image for use.  @2percentsilk and @parkr are discussing requirements.

2percentsilk:
Moving our convos here versus slack!
@parkr recommended the following initially:
> To be consistent with GitHub Pages, it would be great to install the latest patch version of Ruby 2.7.x, and Jekyll 3.9.x
Dependencies (Jekyll, Ruby, etc) are shared here: https://pages.github.com/versions/
If you install Jekyll 3.9.x and the Gemfile requires 4.x, I think bundle install will ignore the systemwide Jekyll and install the repo-specific one. The only potential problem would be running `jekyll build` – will it pick the pre-installed one or the one in Gemfile? I'd want to test that...

We're now discussing if it makes sense to have an overarching general use Jekyll image as well as a more specific one geared to GitHub pages. My thought is that we want to have one image that ideally works for both so that we don't have to deal with the maintenance of two very similar images and definitions.

The lastest question from @parkr is "could we `alias jekyll="bundle exec jekyll"`? Then we can just run bundle install and it'll automatically use the bundled version"
Pardon my ignorance but would this let us just prebuild one image and have it work for general Jekyll development & Pages? I think our goal is to have images that don't require us to run any installations in the `postCreateCommand` since that's what takes the longest at the moment.

parkr:
If you're worried about post-creation command speed, then the best bet is to install the `github-pages` gem (Jekyll 3.x) and then install Jekyll v4. The most common configurations for Jekyll sites are GitHub Pages and the latest Jekyll. If these are pre-installed on the system, then when `postCreateCommand` runs `bundle install`, it will be very fast.

Chuxel:
//cc: @cmendible as well since we'd be converting that definition into a pre-built one. 

So:

```Dockerfile
FROM mcr.microsoft.com/vscode/devcontainers/ruby:2

# ...

RUN gem install bundler jekyll github-pages
```

...and have a pre-built image with all this in it. Then leave https://github.com/microsoft/vscode-dev-containers/blob/main/containers/jekyll/.devcontainer/post-create.sh largely as-is.

That accurate?

Chuxel:
Put up a PR for review (#929). Let me know if it looks good.

Chuxel:
@parkr Quick check-in: Any feedback here?

[...]

Issue #532, opened 2020-08-31
[Tracking] Image support, deprecation / EOL dates


Chuxel:
The following are release dates and EOL / end of support dates for languages / OS's with pre-built images. After runtime or OS support for a given image has ended, we will stop building images for that version. We also typically only release images for "LTS" or "stable" releases of runtimes/OS's. Others can be built from `.devcontaiiner/base.Dockerfile` available in the definition source code in this repository. 

Releases typically happen monthly on the same release cadence as VS Code and include the most recent changes from the upstream base image or runtime version including security updates (e.g. the latest OS updates and the latest point release of a runtime or platform). You can see the current schedule for the VS Code release in a pinned "Iteration Plan" issue at the [VS Code repository](https://github.com/microsoft/vscode/issues). We may update earlier if a critical need or bug is discovered. 

Dev container images are also versioned independent of the runtime. To avoid unexpected breaks, we recommend pinning to the latest major version of an image release to pick up security and feature updates without breaking changes. This image version number is a prefix to the variant or runtime version. e.g. `mcr.microsoft.com/vscode/devcontainers/typescript-node:0-14` where `0` is the major version number of the dev container image. This happens automatically when using the **Add Development Container Configuration Files...** command in the Remote - Containers extension or GitHub Codespaces. 

To make pinning to a version easier if you wish to do so, recent versions of images include a `devcontainer-info` command that will output information about the image in use including version information and a link to a file with additional information on contents.

> Note that in some cases, we may have to end support earlier than the stated date  below if there are fundamental OS requirements that force it or if the primary extension(s) supporting it have dropped support. In these cases, there would no longer be a functioning VS Code story. Our expectation is that this should be rare given typical GA to EOL timelines and is something we strive to avoid.

| Language  / OS | Version | Release | EOL / End of Support | Image |
|-----------|---------|---------|-----|-----|
| <strike>[Alpine](https://alpinelinux.org/releases/e)</strike> | <strike>3.10</strike> | <strike>June 19th, 2019</strike> | <strike>May 1st, 2021</strike> | <strike>[`mcr.microsoft.com/vscode/devcontainers/base:alpine-3.10`](https://github.com/microsoft/vscode-dev-containers/tree/master/containers/alpine)</strike> |
[...]

PR #929, opened 2021-07-01
Onboard Jekyll to image build

Chuxel:
This implements #919 to generate a jekyll image. Like other images, it will include nvm and yarn, but does not include a specific version of Node.js in `base.Dockerfile`.  Instead this is added in the user `Dockerfile` based on a version in `devcontainer.json`. The user will be prompted for which version when this definition is added via **Add Development Configuration Files**.  Unneeded files (like `base.Dockerfile`) are also filtered out in this case.

To test the image alone:

1. Open this repository in a Codespace
2. Run the following in a terminal in the repository root
    ```bash
    yarn install
    build/vscdc push --no-push --registry mcr.microsoft.com --registry-path vscode/devcontainers --release main jekyll
    docker run -it -u vscode --rm mcr.microsoft.com/vscode/devcontainers/jekyll:dev bash
    ```
3. Once in the container, run `nvm install lts/*` to ensure node installs
4. Do any verification you want to do.

Once merged, `mcr.microsoft.com/vscode/devcontainers/jekyll:dev` will be published, and on release the inital version of the image will go (currently set to `0.0.1`).


Chuxel:
//cc: @cmendible @parkr for a review.  @2percentsilk, @joshspicer, @bailey, @matthewisabel  as FYI.

build/src/utils/image-content-extractor.js
```diff
@@ -533,8 +533,19 @@ async function getImageInfo(imageTagOrContainerName) {
...
+        const imageNameAndDigest = await asyncUtils.spawn('docker', ['inspect', "--format='{{index .RepoDigests 0}}'", image], { shell: true, stdio: 'pipe' });
+        [name, digest] = imageNameAndDigest.trim().split('@');
+    } catch(err) {
+        if(err.result.indexOf('Template parsing error') > 0) {
+            name = 'N/A';
+            digest = 'N/A';
+        } else {
+            throw err;
+        }
+    }
```
Chuxel, line 548: FYI - This is a bug I ran into while testing history markdown generation locally. Works file in CI. To see what ends up in history, run:

```bash
build/vscdc info --registry mcr.microsoft.com --registry-path vscode/devcontainers --release main jekyll
```

The result will be in containers/jekyll/history/dev.md

containers/jekyll/.devcontainer/base.Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# [Choice] Node.js version: none, lts/*, 16, 14, 12
+ARG NODE_VERSION="none"
+RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/local/share/nvm/nvm.sh && nvm install ${NODE_VERSION} 2>&1"; fi
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# [Optional] Uncomment this line to install additional gems.
+# RUN gem install <your-gem-names-here>
```
parkr, line 24: In general, we recommend folks add gems to their Gemfile and use `bundle install` to manage them.

containers/jekyll/README.md
```diff
@@ -6,20 +6,46 @@
...
 
 - If your Jekyll project contains a `Gemfile` in the root folder, the development container will install all gems at startup by running `bundle install`. This is the [recommended](https://jekyllrb.com/docs/step-by-step/10-deployment/#gemfile) approach as it allows you to specify the exact Jekyll version your project requires and list all additional Jekyll plugins.
 - If there's no `Gemfile`, the development container will install Jekyll automatically, picking the latest version. You might need to manually install the other dependencies your project relies on, including all relevant Jekyll plugins.
 
-## Using this definition
+You can decide how often you want updates by referencing a [semantic version](https://semver.org/) of each image. For example:
+
+- `mcr.microsoft.com/vscode/devcontainers/jekyll:0`
+- `mcr.microsoft.com/vscode/devcontainers/jekyll:0.0`
+- `mcr.microsoft.com/vscode/devcontainers/jekyll:0.0.1`
```
parkr, line 32: How do you intend to version the Jekyll container? Since it's not presently tied to a Jekyll or GitHub Pages version, and because the post-install script will still run `bundle install`, it may not need to be versioned at all.

containers/jekyll/README.md
```diff
@@ -6,20 +6,46 @@
...
+
+See [history](history) for information on the contents of each version and [here for a complete list of available tags](https://mcr.microsoft.com/v2/vscode/devcontainers/ruby/tags/list).
```
parkr, line 34:
```suggestion
See [history](history) for information on the contents of each version and [here for a complete list of available tags](https://mcr.microsoft.com/v2/vscode/devcontainers/jekyll/tags/list).
```

Should the links be updated to point to the Jekyll dev container?

containers/jekyll/README.md
```diff
@@ -6,20 +6,46 @@
...
+
+Alternatively, you can use the contents of `base.Dockerfile` to fully customize your container's contents or to build it for a container host architecture not supported by the image.
+
+### Installing Node.js
+
+Given JavaScript front-end web client code written for use in conjunction with a PHP back-end often requires the use of Node.js-based utilities to build, this container also includes `nvm` so that you can easily install Node.js. You can change the version of Node.js installed or disable its installation by updating the `args` property in `.devcontainer/devcontainer.json`.
```
parkr, line 40:
```suggestion
Given JavaScript front-end web client code written for use in conjunction with a Jekyll site often requires the use of Node.js-based utilities to build, this container also includes `nvm` so that you can easily install Node.js. You can change the version of Node.js installed or disable its installation by updating the `args` property in `.devcontainer/devcontainer.json`.
```

containers/jekyll/.devcontainer/base.Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# ENV Variables required by Jekyll
+ENV LANG=en_US.UTF-8 \
+    LANGUAGE=en_US:en \
+    TZ=Etc/UTC \
+    LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8 \
+    LANGUAGE=en_US
+
+# Install bundler, latest jekyll, and github-pages for older jekyll
+RUN gem install bundler jekyll github-pages
```
parkr, line 13: This appears to work, installing both Jekyll 3 and Jekyll 4 as desired to speed up the `bundle install` process.

parkr (approved):
Looks good to me – left a few comments. Thank you!

containers/jekyll/README.md
Chuxel: Yeah this versioning is for the image contents. But if we don't have a reason to release, we wouldn't bump the version number up.

At a minimum, there's monthly OS patching, so we'd update roughly once a month and that would get the latest of everything. That just increments the last digit (since we use semver). We do the same thing for all images in the repo. [This is explained here](https://github.com/microsoft/vscode-dev-containers/issues/532).

The other thing this does is future proof for when/if we need to make breaking changes.

PR #937, opened 2021-07-07
Change bash git ref prompt to more closely match zsh

joshspicer:
fixes https://github.com/github/codespaces/issues/3012

More closely matches [how ohmyzsh displays git ref information on a detached HEAD](https://github.com/ohmyzsh/ohmyzsh/blob/d646884add277d134235a9b18ab755388d6e0d8d/lib/git.zsh#L20-L23).  

The behavior in a standard common-debian bash shell, when browsing a detached HEAD, changes from: 

image

to

image

### DIFF
--- a/script-library/common-debian.sh
+++ b/script-library/common-debian.sh
@@ -292,11 +292,13 @@ __bash_prompt() {
         && [ "$XIT" -ne "0" ] && echo -n "\[\033[1;31m\]➜" || echo -n "\[\033[0m\]➜"`'
     local gitbranch='`\
         export BRANCH=$(git symbolic-ref --short HEAD 2>/dev/null || git rev-parse --short HEAD 2>/dev/null); \
+        if [ "${BRANCH}" != "" ]; then \
             echo -n "\[\033[0;36m\](\[\033[1;31m\]${BRANCH}" \
             && if git ls-files --error-unmatch -m --directory --no-empty-directory -o --exclude-standard ":/*" > /dev/null 2>&1; then \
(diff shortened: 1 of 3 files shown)


script-library/common-debian.sh
```diff
@@ -291,7 +291,7 @@ __bash_prompt() {
         && [ ! -z "${GITHUB_USER}" ] && echo -n "\[\033[0;32m\]@${GITHUB_USER} " || echo -n "\[\033[0;32m\]\u " \
         && [ "$XIT" -ne "0" ] && echo -n "\[\033[1;31m\]➜" || echo -n "\[\033[0m\]➜"`'
     local gitbranch='`\
-        export BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null); \
+        export BRANCH=$(git symbolic-ref --short HEAD 2>/dev/null || git rev-parse --short HEAD 2>/dev/null); \
         if [ "${BRANCH}" = "HEAD" ]; then \
```
jkeech, line 295: Is this block ever possible anymore, or is it safe to remove?

jkeech (commented):
Thanks! Can you also update `common-alpine` and `common-redhat` with the same fix?

script-library/common-debian.sh
joshspicer: Good point -  i'll take a closer look, but probably not.

joshspicer:
These changes locally in remote-containers in the ubuntu image

image

jkeech (approved):
LGTM!

PR #953, opened 2021-07-19
enable html format templating

AarynSmith:
Add the template format option to prevent vscode from messing with the hugo template format in html files.


Chuxel:
LGTM!

PR #954, opened 2021-07-20
Update the R definition

eitsupi:
- Remove `reditorsupport.r-lsp` as its functionality has been merged into `ikuyadeu.r`. (https://github.com/REditorSupport/vscode-R/pull/695)
- `r.sessionWatcher` is now true by default, so remove it from settings. (https://github.com/REditorSupport/vscode-R/pull/670)
- Remove the `--repos` option from the `install2.r` command, which install the R packages in the container, as it is unnecessary. (Related to the base image improvement: https://github.com/rocker-org/rocker-versioned2/pull/195)
- Install the `httpgd` package for better plot viewer. (https://github.com/REditorSupport/vscode-R/pull/620)
- Edit the Rprofile to enable self-managed R sessions and use `httpgd` by default in VSCode.
- Add a setting to make it easier to use `radian` in the integrated terminal.

cc @kmehant @Ikuyadeu @randy3k @renkun-ken

### DIFF
--- a/containers/r/.devcontainer/Dockerfile
+++ b/containers/r/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-# R version: latest, ... ,4.0.1 , 4.0.0
+# R version: latest, ... , 4.0.3
 ARG VARIANT="latest"
 FROM rocker/r-ver:${VARIANT}
 
@@ -29,7 +29,13 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && install2.r --error --skipinstalled --ncpus -1 \
         devtools \
         languageserver \
+        httpgd \
     && rm -rf /tmp/downloaded_packages
+
+# R Session watcher settings.
+# See more details: https://github.com/REditorSupport/vscode-R/wiki/R-Session-watcher
+RUN echo 'source(file.path(Sys.getenv("HOME"), ".vscode-R", "init.R"))' >> ${R_HOME}/etc/Rprofile.site \
+    && echo 'options(vsc.use_httpgd = TRUE)' >> ${R_HOME}/etc/Rprofile.site
 
 # [Optional] Uncomment this section to install additional OS packages.
 # RUN apt-get update \
--- a/containers/r/.devcontainer/Dockerfile
+++ b/containers/r/.devcontainer/Dockerfile
@@ -26,7 +26,7 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
         libxt-dev \
     && apt-get autoremove -y && apt-get clean -y && rm -rf /var/lib/apt/lists/* /tmp/library-scripts \
     && python3 -m pip --no-cache-dir install radian \
-    && install2.r --error --skipinstalled --repos ${CRAN} --ncpus -1 \
+    && install2.r --error --skipinstalled --ncpus -1 \
         devtools \
         languageserver \
     && rm -rf /tmp/downloaded_packages
--- a/containers/r/.devcontainer/devcontainer.json
+++ b/containers/r/.devcontainer/devcontainer.json
@@ -2,7 +2,7 @@
 	"name": "R (Community)",
 	"build": {
 		"dockerfile": "Dockerfile",
-		// Update VARIANT to pick a specific R version: latest, ... ,4.0.1 , 4.0.0
+		// Update VARIANT to pick a specific R version: latest, ... , 4.0.3
 		"args": { "VARIANT": "latest" }
 	},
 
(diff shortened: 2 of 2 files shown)


eitsupi:
By the way, the current R definition says that it doesn't work with Codespaces, but containers and extensions for R have changed a lot between when this research was done 5ca22d160cbadcb3d274674618752517e81fabaa and now.

I'd like to find out what's going on now if I could, but I don't have access to Codespaces so I can't find out...

renkun-ken:
Is there a way to get access to codespaces for container maintainers?

renkun-ken:
@eitsupi Do you think we should also add [R Debugger](https://github.com/ManuelHentschel/VSCode-R-Debugger) and also install [vscDebugger](https://github.com/ManuelHentschel/vscDebugger) to provide a full R development experience including debugging?

eitsupi:
@renkun-ken Yes, it would be great if it could be added.
I also considered adding it when I added the current definition in #732, but I had never used the R Debugger myself, so I decided to forgo it.

And, it would be ideal to have the `{httpgd}` package installed as well, but the current container takes about 10 minutes to build, so I decided to forgo it this time.
(Half of the build time is spent running common-debian.sh in this repository, and the other half is installing R packages)

If we can use images pre-built by REditorSupport, which owns the vscode-R extension and the `{languageserver}` package, as Julia's definition added in #450, users can easily use R in Remote-Containers (and Codespaces) with more pre-installed packages in container.

renkun-ken:
Thanks for pointing to the Julia's devcontainer. We'll probably maintain some pre-built images under [REditorSupport](https://github.com/REditorSupport).

eitsupi:
@renkun-ken Once again, I thought about the container settings and tried various things, and I think it is indeed significant to provide a better R development environment with VSCode as you said, so I would like to add the following settings.

- Add the `httpgd` package and use it by default (httpgd's plots are really nice!)
- Edit the Rprofile to enable self-managed R sessions
- Add integrated terminal profile for `radian`

Since the base image `rocker/r-ver` installs packages from past CRAN snapshots by default, images tagged with versions earlier than `httpgd` was registered with CRAN will fail to install `httpgd`.
Therefore, I have updated the comments to raise the minimum version to R 4.0.3.

I decided not to install the `vscDebugger` package because it is not registered with CRAN and I cannot eliminate the possibility of version inconsistencies with dependent packages installed from past CRAN mirrors.
And, if we  add the R Debugger, we might have to add launch.json and update the sample project and documentation.

renkun-ken:
Thanks for the nice work!

>I decided not to install the vscDebugger package because it is not registered with CRAN and I cannot eliminate the possibility of version inconsistencies with dependent packages installed from past CRAN mirrors.

From the [DESCRIPTION](https://github.com/ManuelHentschel/vscDebugger/blob/master/DESCRIPTION), `vscDebugger` only depends on `jsonlite` and `R6` with no specific minimum version requirement. I'm not sure what you mean by "version inconsistencies with dependent packages"?

> And, if we add the R Debugger, we might have to add launch.json and update the sample project and documentation.

The current R Debugger has built-in configs to work out of the box for most cases. I guess we don't have to add `launch.json` for now.

eitsupi:
> From the [DESCRIPTION](https://github.com/ManuelHentschel/vscDebugger/blob/master/DESCRIPTION), `vscDebugger` only depends on `jsonlite` and `R6` with no specific minimum version requirement. I'm not sure what you mean by "version inconsistencies with dependent packages"?

Does it seem unlikely that a future update of vscDebugger will require version restrictions? @ManuelHentschel
(Of course, if that's the case, we could just update the Dockerfile to fix it, so maybe I'm worrying too much)

Also, I haven't decided whether it's better to install from the latest GitHub commit at the docker build stage or by using the command from R Debugger.
I believe it is possible to install from the latest release using the GitHub API, but I feel it is too long to write in this Dockerfile...

> The current R Debugger has built-in configs to work out of the box for most cases. I guess we don't have to add `launch.json` for now.

Thank you for telling me about it. I don't understand too much about the R Debugger (and about debugging tasks in R), so I'll try to use it tomorrow.

eitsupi:
I'm sorry, but I've touched the R Debugger, but I don't understand the proper `launch.json` and test-project configuration at this time, and I can't decide if I should include the vscDebugger package in the Dockerfile.
Please let me exclude the addition of R Debugger from this PR.
(Please compare "[r](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/r)" with containers in other languages, e.g. "[python-3](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/python-3)", for the need to add `launch.json`.)

Chuxel:
LGTM!

eitsupi:
Thank you for merging !

PR #959, opened 2021-07-21
Set the default git editor if not already set

### LINKED ISSUE #952: VSCode started injecting `GIT_EDITOR=code --wait` to remote container environment variables
All my remote container environments suddenly started having the `GIT_EDITOR=code --wait` environment variable injected. This means that any git operation run from a terminal window causes a new tab to open the `COMMIT_EDITMSG` instead of the environment default (typically `nano`).

VSCode open to a local folder does not have this issue, git operations open in `nano` as expected.

I tried to find a new setting that might control this behaviour but I did not find anything, nor did I find mention of such a change in recent [release notes](https://github.com/microsoft/vscode-docs/tree/main/remote-release-notes).

The worst part for me is that this variable takes precedence over any value in `~/.gitconfig`, and VSCode overrides any variable set in my `.devcontainer` folder; I tried adding it to my `docker-compose.yaml` file and rebuilding the container but the `code --wait` value persists. e.g.:

```yaml
services:
  vscode:
    image: mcr.microsoft.com/vscode/devcontainers/base:debian
    user: vscode
    environment:
      GIT_EDITOR: nano
```

I hope this is not a new default setting that we cannot change and is just a configuration change that I messed up locally, unfortunately this change is highly disruptive for me (and probably many other that prefer to work in the terminal).  😰

Any tips to go back to the original behaviour are appreciated!

- VSCode Version: 1.58.2
- Local OS Version: macOS BigSur
- Remote OS Version: `ubuntu:focal`
- Remote Extension/Connection Type: Docker
- Logs:

Steps to Reproduce:

1. Open a remote container
2. try to `git commit`

Does this issue occur when you try this locally?: No
Does this issue occur when you try this locally and all extensions are disabled?: N/A

stellirin:
closes #952

### DIFF
--- a/script-library/common-alpine.sh
+++ b/script-library/common-alpine.sh
@@ -172,12 +172,14 @@ if [ -t 1 ] && [[ "${TERM_PROGRAM}" = "vscode" || "${TERM_PROGRAM}" = "codespace
     ((sleep 10s; touch "$HOME/.config/vscode-dev-containers/first-run-notice-already-displayed") &)
 fi
 
-# Set the default git editor
-if  [ "${TERM_PROGRAM}" = "vscode" ]; then
-    if [[ -n $(command -v code-insiders) &&  -z $(command -v code) ]]; then 
-        export GIT_EDITOR="code-insiders --wait"
-    else 
-        export GIT_EDITOR="code --wait"
+# Set the default git editor if not already set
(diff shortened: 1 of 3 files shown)


Chuxel:
LGTM!  Thanks.

PR #964, opened 2021-07-25
add bash and zsh completion for kubectl

bartr:
This PR adds bash and zsh completion for kubectl which provides a better developer experience and prevents inherited containers from having to implement.


Chuxel:
LGTM!

PR #966, opened 2021-07-26
Improve the default experience for Deno containers

kitsonk:
👋  I am part of the @denoland core time and maintain denoland/vscode_deno.

There are a few things can be adjusted in the configuration to make a better out of the box experience with Deno when setting up a dev container.


Chuxel:
LGTM!

PR #968, opened 2021-07-26
updated package install block

bartr:
- the package install block got overwritten during a recent PR
- ensure all packages we check for are installed


Chuxel:
@bartr Thanks - though this is correct as is. PR https://github.com/microsoft/vscode-dev-containers/pull/963 adds a utility function called check_packages that does a check and install. The updates in the other PR were included. The reason git is separate is that it may have been installed via something other than a package manager (e.g. built from source).

PR #980, opened 2021-08-03
Queue up bullseye release for debian, cpp definitions

Chuxel:
Debian 11 / Bullseye release date: August 14th, 2021

Default is still buster/debian-10 so we can merge a bit ahead of the bullseye official release.


joshspicer (approved):
Queuing up the codespaces PR to pull this in/check sizes 👍

joshspicer:
@Chuxel - should we also update the node variants?
https://github.com/microsoft/vscode-dev-containers/blob/main/containers/javascript-node/definition-manifest.json#L2

Chuxel:
@joshspicer There  are not bullseye variants of them yet.  Only `buildpack-deps` currently has bullseye which is what the debian definition uses and cpp then uses that.

Even then, `node` defaulted to Debian 9 if you don't specify a distro well after the release of 10 (and in fact may still do so).

Chuxel:
@joshspicer I made a tweak here to keep Debian 10 as the default since there could be issues with 11 for some period after release.  Bullseye will be an option, however.

PR #989, opened 2021-08-13
Add global ts-node to TypeScript image

Jolg42:
I think a Node/TypeScript image could include ts-node by default installed globally, it would be smoother for many projects.


Chuxel:
@DanielRosenwasser - What do you think? Should we include `ts-node` by default in the TypeScript dev container? I use it personally, but good to get your input.

DanielRosenwasser:
I don't know if I can strongly recommend it. ts-node currently turns type-checking on by default if I recall correctly, so there's some level of configuration to think about if you don't want that (and personally I don't think that that's the ideal default). It's also easy enough to install on a fresh container anyway.

Chuxel:
Given Daniel's feedback here, I think we'll pass on this but thanks so much for contributing!

PR #988, opened 2021-08-13
Fixed bad conditional, fixed nvm install for lts

cbbm142:
Updated nvm commands to install a specific version, or lts properly. Fixes #987

PR #992, opened 2021-08-13
Revert "Fixed bad conditional, fixed nvm install for lts"

Chuxel:
Reverts microsoft/vscode-dev-containers#988


Chuxel:
Bumped wrong button. 😅

Issue #1000, opened 2021-08-18
Updating `go` container to latest release (1.17) and dropping unsupported  (1.15)


joshspicer:
**see more in the [image tracking issue](https://github.com/microsoft/vscode-dev-containers/issues/532).**

Go is supported two release back, so we are picking up 1.17 and subsequently dropping 1.15

PR #1001, opened 2021-08-18
update golang to in-support versions (1.16 and 1.17)

joshspicer:
closes https://github.com/microsoft/vscode-dev-containers/issues/1000

Updates our versions of `go` to track latest supported

### TODO
- [ ] Test with `dev` tag
- [ ] Bump repo version to do full release
- [ ] Update [tracking issue](https://github.com/microsoft/vscode-dev-containers/issues/532)
- [ ] Pull into Codespaces

### DIFF
--- a/containers/go/definition-manifest.json
+++ b/containers/go/definition-manifest.json
@@ -1,6 +1,6 @@
 {
-	"variants": ["1.16", "1.15"],
-	"definitionVersion": "0.203.0",
+	"variants": ["1.16", "1.17"],
+	"definitionVersion": "0.204.0",
 	"build": {
 		"latest": true,
 		"rootDistro": "debian",
@@ -8,7 +8,7 @@
 			"go:${VERSION}-${VARIANT}"
 		],
 		"variantTags": {
-			"1.16": [ "go:${VERSION}-1" ]
+			"1.17": [ "go:${VERSION}-1" ]
 		}
 	},
 	"dependencies": {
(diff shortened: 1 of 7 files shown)


joshspicer:
builds fine in remote containers 👍 
image

containers/go/definition-manifest.json
```diff
@@ -1,14 +1,14 @@
 {
-	"variants": ["1.16", "1.15"],
-	"definitionVersion": "0.203.0",
+	"variants": ["1.16", "1.17"],
```
Chuxel, line 2:
```suggestion
	"variants": ["1.17", "1.16"],
```

The order here matters. The zero index is assigned "latest", so put 1.17 there.

Chuxel (changes requested):
Two things.

It looks like there's a bug in the go-debian script this exposed.  It turns out that `go version` emits a version of `go1.17` instead of `go1.17.0` like you'd normally expect.

Therefore this line

```
if [[ "1.16" > "$(go version | grep -oP 'go\K[0-9]+\.[0-9]+\.[0-9]+')" ]]; then
```
needs to make the last section of the semver optional. I think this should do it:
```
if [[ "1.16" > "$(go version | grep -oP 'go\K[0-9]+\.[0-9]+(\.[0-9]+)?')" ]]; then
```
Go 1.17 recommends using "go install" for CLIs instead of "go get", but older versions still need the old syntax. So this detects the actually installed version (since we may have skipped installing go itself) and only uses `go install` for go 1.16+.

The other is a minor tweak inline...

joshspicer:
nice - that looks good
image

Chuxel (approved):

Chuxel:
@joshspicer Looks good! Feel free to merge when ready!

Issue #914, opened 2021-06-24
Update build scripts to use "docker buildx build"


Chuxel:
As a part of the work to enable ARM64 for images, we'll need to switch the build system to use buildx instead of the default docker build. e.g.

```
docker buildx create --use --name multiarch
docker run --privileged --rm tonistiigi/binfmt --install all
docker buildx build .devcontainer -f .devcontainer/base.Dockerfile --platform linux/amd64,linux/arm64
```

Chuxel:
The branch https://github.com/microsoft/vscode-dev-containers/tree/clantz/mutli-arch-build has a working version of the build, but buildx is reporting a native crash in a simple `curl` call (likely tied to something QEMU and openSSL) when trying to execute the build in GitHub Actions or emulated.  This is blocking merge.

imakahn:
Hi @Chuxel 

I was working through the `node-debian.sh` script and also experienced the segfault on the curl call to fetch the yarn gpg file. The issue has been detailed here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=989604

A immediate workaround would be to use `wget`, as such:

`wget https://dl.yarnpkg.com/debian/pubkey.gpg -O /tmp/pubkey.gpg`
`cat /tmp/pubkey.gpg | gpg --dearmor > /usr/share/keyrings/yarn-archive-keyring.gpg`

However this would still leave the global curl being shipped as broken for some https urls.

So to instead resolve that (and leave the curl call unchanged):

`wget https://cloudfront.debian.net/debian/pool/main/o/openssl/libssl1.1_1.1.1k-1_arm64.deb -O /tmp/libssl1.1_1.1.1k-1_arm64.deb`
`sudo dpkg -i /tmp/libssl1.1_1.1.1k-1_arm64.deb`

I hope this helps!

Chuxel:
@imakahn Thanks for the tip. Yeah, beyond curl, its libssl itself that is broken, so I saw a number of issues. This one was just the really obvious one.  I'm a bit wary to bring in testing bits into stable. I'd expecting this will get pulled into stable (soon to be old-stable), but there's a lot of focus on bullseye right now.

Dids:
Have there been any updates with this?

So far I've been creating my own images/forks of the definitions available in this repo, as well as replacing all curl calls with wget, to get rid of the QEMU crashes when doing multi-arch builds.

[...]

PR #1018, opened 2021-08-30
Update build system to use buildx, support multi-arch builds

Chuxel:
Implements https://github.com/microsoft/vscode-dev-containers/issues/914. 

This PR does not update config to start building linux/arm64, but rather sets up the build system to do so. See https://github.com/microsoft/vscode-dev-containers/issues/558#issuecomment-905117722 for issues preventing switch, current plan of action.

### DIFF
--- a/build/src/push.js
+++ b/build/src/push.js
@@ -30,6 +30,16 @@ async function push(repo, release, updateLatest, registry, registryPath, stubReg
     const stagingFolder = await configUtils.getStagingFolder(release);
     await configUtils.loadConfig(stagingFolder);
 
+    // Use or create builder
+    console.log('(*) Setting up builder...');
+    const builders = await asyncUtils.exec('docker buildx ls');
+    if(builders.indexOf('vscode-dev-containers') < 0) {
+        await asyncUtils.spawn('docker', ['buildx', 'create', '--use', '--name', 'vscode-dev-containers']);
+    } else {
+        await asyncUtils.spawn('docker', ['buildx', 'use', 'vscode-dev-containers']);
+    }
+    await asyncUtils.spawn('docker', ['run', '--privileged', '--rm', 'tonistiigi/binfmt', '--install', 'all']);
+
     // Build and push subset of images
     const definitionsToPush = definitionId ? [definitionId] : configUtils.getSortedDefinitionBuildList(page, pageTotal, definitionsToSkip);
     await asyncUtils.forEach(definitionsToPush, async (currentDefinitionId) => {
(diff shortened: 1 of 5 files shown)


build/src/push.js
```diff
@@ -30,6 +30,16 @@ async function push(repo, release, updateLatest, registry, registryPath, stubReg
...
 
+    // Use or create builder
+    console.log('(*) Setting up builder...');
+    const builders = await asyncUtils.exec('docker buildx ls');
+    if(builders.indexOf('vscode-dev-containers') < 0) {
+        await asyncUtils.spawn('docker', ['buildx', 'create', '--use', '--name', 'vscode-dev-containers']);
+    } else {
+        await asyncUtils.spawn('docker', ['buildx', 'use', 'vscode-dev-containers']);
+    }
+    await asyncUtils.spawn('docker', ['run', '--privileged', '--rm', 'tonistiigi/binfmt', '--install', 'all']);
```
joshspicer, line 41: No idea what this line is doing :D .  Maybe add a comment to make it more clear?

build/src/push.js
```diff
@@ -84,7 +94,26 @@ async function pushImage(definitionId, repo, release, updateLatest,
...
+                case 'undefined': architectures = ['linux/amd64']; break;
+            }
+            console.log(`(*) Target image architectures: ${architectures.reduce((prev, current) => prev += `\n     ${current}`, '')}`);
+            let localArchitecture = process.arch;
+            switch(localArchitecture) {
+                case 'arm': localArchitecture = 'linux/arm/v7'; break;
+                case 'aarch32': localArchitecture = 'linux/arm/v7'; break;
+                case 'aarch64': localArchitecture = 'linux/arm64'; break;
+                case 'x64': localArchitecture = 'linux/amd64'; break;
+                case 'x32': localArchitecture = 'linux/386'; break;
```
joshspicer, line 110: is this supposed to be `linux/i386`?  Doesn't look right to me without the `i` for some reason, but I might just be wrong

joshspicer (approved):

build/src/push.js
Chuxel: Sure - it sets up QEMU to enable multi-platform builds. Docker Desktop does htis as a part of install so people don't know. Will add a comment.

build/src/push.js
Chuxel: Yeah for Docker it's w/o the `i`.  Type `docker buildx ls` locally to see architectures that are setup to emulate. Desktop has them all installed.

Chuxel:
@joshspicer Triggered dev builds and they all successfully ran. The codespaces-linux image will now be built using buildx which should help reduce total layers.

PR #1039, opened 2021-09-10
Java arm64 images, buster variants for back compat

Chuxel:
As described in https://github.com/microsoft/vscode-dev-containers/issues/558#issuecomment-905117722, the situation for Docker on M1 macs is not ideal since both `debian:buster` nor `ubuntu:focal` experience segmentation faults due to an issue in `libss1.1`. It seems unlikely that `buster` will be patched now that `bullseye` is out (and buster therefore gets security fixes only). 

Therefore this PR:
1. Adds arm64 builds for Debian `bullseye` 
2. Adds `buster` variants that are x86 only for backwards compat.
3. Keeps the default image tags (8, 11, 16) at `buster`

@testforstephen @akaroml - Debian `bullseye` is pretty recent, but works on arm64. Would it make sense to default to stick with `buster` as is done here?  Java should be fairly abstracted from OS version updates, but I didn't want to make an update that could break users. Thoughts?

//cc: @joshspicer @2percentsilk @bamurtaugh @chrmarti

### DIFF
--- a/containers/java-8/.devcontainer/Dockerfile
+++ b/containers/java-8/.devcontainer/Dockerfile
@@ -1,4 +1,6 @@
-FROM mcr.microsoft.com/vscode/devcontainers/java:0-8
+# [Choice] Debian OS version: buster, bullseye
+ARG VARIANT="buster"
+FROM mcr.microsoft.com/vscode/devcontainers/java:0-8-${VARIANT}
 
 # [Option] Install Maven
 ARG INSTALL_MAVEN="false"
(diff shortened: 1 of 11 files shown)


testforstephen:
Looks good to me. Only one concern is how user can easily distinguish which Java variant to use when they use "Remote-Containers: Add Remote Development Configuration Files..." command to pick up a Java definition?

This is current dropdown for Java versions:
image

I assume the PR will add two more options like "11-bullseye" and "16-bullseye". Is it possible to show some details for these options to mention the architecture they support?

Chuxel:
@testforstephen Yeah the list would include both buster and bullseye.  I could have the label ("Java version") mention that bullseye should be used for arm64 and M1 macs.

testforstephen:
yes, since `bullseye` is very new and we can stay on `buster` as default for a while. But eventually we need to switch to `bullseye` at some point in future.

Issue #917, opened 2021-06-24
Add --init to all definitions


Chuxel:
For the most part, containers that come from this repository are intended to run long enough that having an init process that cleans up "zombie processes" is useful. For performance reasons, systemd or other init systems are not used by default. As a result, it's likely a good idea for us to have the built-in "tini" init process enabled via `"runArgs" : [ "--init" ]` or in the Docker Compose file.

joshspicer:
for features, would we just want to include the arg automatically when the CLI builds the image?

Chuxel:
@chrmarti I assume you didn't wire up "--init" into `features.json` yet, correct?  We certainly could.

PR #1049, opened 2021-09-14
add --init to definitions

joshspicer:
closes: https://github.com/microsoft/vscode-dev-containers/issues/917


Chuxel:
One note - We'll also need to add the equivalent to docker-compose.yml for those definitions.

```yaml
services:
  app:
    #...
    init: true
```

joshspicer:
I did in some places, and was going to ask you.  Many of the docker-compose definitions have explicit versions at the top, and init was introduced in version 3.7.  https://docs.docker.com/compose/compose-file/compose-file-v3/#init

You see any harm in updating/removing the version for those docker-compose definitions?

joshspicer:
One example is https://github.com/microsoft/vscode-dev-containers/blob/main/containers/javascript-node-postgres/.devcontainer/docker-compose.yml pinned at version 3.0

Chuxel:
I think `3` is interpreted as `3.x` -- so `3.9`.   You'd need to use `3.0` to pin to that version.  (The only place this wasn't happening in the old version of docker-compose.) That said, I think it's fine to pin to a recent Docker Compose version.

joshspicer:
> I think `3` is interpreted as `3.x` -- so `3.9`.

I'd assumed that too, but the docs say differently: 
https://docs.docker.com/compose/compose-file/compose-versioning/#v2-and-v3-declaration

"When specifying the Compose file version to use, make sure to specify both the major and minor numbers. If no minor version is given, 0 is used by default and not the latest minor version."

Chuxel:
Interesting, then,  yeah, we probably should pick a later version.

chrmarti:
Shouldn't we add this to the extension instead of adding it to all definitions?

Chuxel:
@chrmarti Very happy to do so as I commented here https://github.com/microsoft/vscode-dev-containers/issues/917#issuecomment-919383191.  Do we want to make that the plan?  If so, I'm good with it.

chrmarti:
We could add it automatically if `overrideCommand` is not false to give the user some control. What do you think?

Chuxel:
@chrmarti Yeah it sort of needs to be its own property.  The challenge is that if you want to use systemd, you can't have this enabled since you have to be PID 0.  We have `runArgs` for when we need to add it ad-hoc, but with features we'll need some way of enabling it.

eitsupi:
Hello. It looks like `--init` hasn't been added to some definitions, is this intentional?

joshspicer:
@eitsupi - which definition are you referring to? There are a couple I don't think i updated for various reasons, and also some I may have missed.

eitsupi:
@joshspicer I've noticed R (I was paying attention because I am contributing) and Julia as far as I can see from the container list.
But I don't fully understand the need for `--init`, so it may not be the case.

joshspicer:
The intention of these changes was just as a performance optimization (since most of the time these containers are intended to be long-running processes). Looks like there was no reason to exclude these two - so adding the line in.  Thanks!

PR #1057, opened 2021-09-17
update install.sh

joshspicer:
Update to work with https://github.com/microsoft/vscode-remote-containers/pull/153


Chuxel:
Does the Dockerfile need to get updated too for the copy statement?

joshspicer:
This dockerfile won't be used anymore, I now keep it here:
https://github.com/joshspicer/vscode-remote-containers/blob/4f3126aa0832c55bfb04c793971d2b853b83611c/src/configuration/containerFeaturesConfiguration.ts#L55

joshspicer:
I suppose for clarity, i'll delete that Dockerfile

Chuxel:
Can we still regression test?  The Dockerfile in this repo was there to allow us to run a test.sh to verify the script is working as things change.  We can wire that into CI.  Using the test script and test.sh right now.

joshspicer:
ah yes, will fix that up.

joshspicer:
Added new CI job to run that test.sh

Chuxel:
LGTM!! Merge when ready!

PR #1063, opened 2021-09-25
Make VNC work with Python 3.9

jonico:
* update WEBSOCKETIFY to version 0.10.0 to make it work with Python 3.9
* version 0.9.0 resulted in error "handler exception: 'array.array' object has no attribute 'fromstring'
* upstream fix: https://github.com/novnc/websockify/issues/473


Chuxel:
Good catch!  Thanks.

PR #1064, opened 2021-09-25
Provide VNC env variables in Dockerfile example

jonico:
* without setting VNC_RESOLUTION env variable, tigervnc will not start up
* copied values from repository-containers/images/github.com/microsoft/vscode/.devcontainer/base.Dockerfile


Chuxel:
Thanks!

PR #1073, opened 2021-09-28
Update docker-debian.sh for Docker compose v2.0

bigtallcampbell:
As of Docker Compose 2.0 (28Sept2021), Docker Compose replaced the "x86_64" suffixed to "AMD64".  Update the target arch variable to pull the AMD64 build if we're running on x86_64 architecture


Chuxel:
@bigtallcampbell Like #1074, unfortunately, this doesn't work. Docker Compose v2 is fundamentally different in the way it is installed. It actually is installed via the moby-compose package (so docker compose is v2). We are pinning docker-compose to v1 currently.

See #1075 as we've now resolved this.

PR #1074, opened 2021-09-28
Update docker-in-debian.sh for Docker compose v2

bigtallcampbell:
Starting with Docker Compose v2, x86_64 suffixed builds have been replaced with AMD64.  Updating the logic to find the right package.


Chuxel:
@bigtallcampbell This doesn't work. Docker Compose v2 is fundamentally different in the way it is installed.  It actually is installed via the moby-compose package (so `docker compose` is v2). We are pinning docker-compose to v1 currently.

See https://github.com/microsoft/vscode-dev-containers/issues/1075

PR #1078, opened 2021-09-28
Select docker engine and CLI versions for Docker scripts

joshspicer:


script-library/docker-debian.sh
```diff
@@ -109,6 +110,52 @@ find_version_from_git_tags() {
...
+
+# Refresh apt lists
+apt-get update
+
+# Soft version matching for CLI
+if [ "${CLI_VERSION}" = "latest" ] || [ "${CLI_VERSION}" = "lts" ] || [ "${CLI_VERSION}" = "stable" ]; then
+    # Empty, meaning grab whatever "latest" is in apt repo
+    cli_version_suffix=""
+else    
+    find_version_from_git_tags CLI_VERSION ${cli_git_tags}
```
Chuxel, line 147: The interesting thing here is that, since we're using apt, we can actually use the list we get back from apt-cache to do this and avoid a hit to git.  That would then only include versions that can actually be installed for the given distro version.  You can then use grep to match. e.g. 

```
requested_version="$(echo "${version_list}" | grep -E -m 1 "^${requested_version//./\\.}([\\.\\s\\+]|$)")"
```

Chuxel (commented):
A couple of thoughts here inline.

script-library/docker-debian.sh
```diff
@@ -109,6 +110,49 @@ find_version_from_git_tags() {
...
+# Refresh apt lists
+apt-get update
+
+# Soft version matching for CLI
+if [ "${CLI_VERSION}" = "latest" ] || [ "${CLI_VERSION}" = "lts" ] || [ "${CLI_VERSION}" = "stable" ]; then
+    # Empty, meaning grab whatever "latest" is in apt repo
+    cli_version_suffix=""
+else    
+    # Fetch a valid version from the apt-cache (eg: the Microsoft repo appends +azure, breakfix, etc...)
+    cli_version_suffix="=$(apt-cache madison ${cli_package_name} | awk -F"|" '{print $2}' | grep -m 1 "${CLI_VERSION}" | xargs)"
```
Chuxel, line 146: On second look on the grep statement.  I think to get partial matches to work you'll need to tweak the grep to  use `-E` start at the beginning of the line with "^" and escape ".". You'd then want to add an end of the match being something like "(\\.|$)".  Otherwise you could match a value like "1" anywhere in the string.

script-library/container-features/src/features.json
```diff
@@ -7,9 +7,9 @@
 			"options": {
 				"version": {
 					"type": "string",
-					"enum": ["latest"],
+					"proposals": ["latest"],
 					"default": "latest",
-					"description": "Currently unused."
+					"description": "Select the Docker/Moby Engine+CLI versions"
```
Chuxel, line 12:
```suggestion
					"description": "Select or enter a Docker/Moby Engine version. (Available versions may vary by distribution)"
```

script-library/container-features/src/features.json
```diff
@@ -7,9 +7,9 @@
 			"options": {
 				"version": {
 					"type": "string",
-					"enum": ["latest"],
+					"proposals": ["latest"],
```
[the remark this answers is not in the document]
Chuxel: Looks like the offical Docker CLI for stretch tops out at 19
```suggestion
					"proposals": ["latest", "20.10" ],
```

script-library/container-features/src/features.json
```diff
@@ -65,9 +65,9 @@
...
```
Chuxel, line 68: Same suggestion as above.
```suggestion
					"proposals": ["latest", "20.10"],
```

script-library/container-features/src/features.json
```diff
@@ -65,9 +65,9 @@
...
 					"default": "latest",
-					"description": "Currently unused."
+					"description": "Select the Docker/Moby CLI version"
```
Chuxel, line 70:
```suggestion
					"description": "Select or enter a Docker/Moby CLI version. (Available versions may vary by Linux OS.)"
```

script-library/docker-debian.sh
Chuxel: Same would apply to the docker-in-docker script

joshspicer: image

I still think this can be improved (some edge cases when someone supplies something like `2.` - but I think good enough for now!

Chuxel: I suspect if you remove the `.+` it will work.  You only need to match on part of the string since you're not passing `-o` - the entire line will get returned.  I can try

PR #1080, opened 2021-09-29
add version selection to powershell

joshspicer:


script-library/container-features/src/features.json
```diff
@@ -965,7 +965,7 @@
 					"type": "string",
 					"enum": ["latest"],
```
Chuxel, line 966: Since we can fall back to pulling from a repository here we likely can safely allow people to enter a version.  So this could be a `proposal`.  7.2 is coming out soon, so listing latest and 7.1 is probably good.
```suggestion
					"proposals": ["latest", "7.1"],
```

script-library/container-features/src/features.json
```diff
@@ -965,7 +965,7 @@
...
 					"default": "latest",
-					"description": "Currently unused."
+					"description": "Select version of Powershell, or latest."
```
Chuxel, line 968:
```suggestion
					"description": "Select or enter a version of PowerShell."
```

Chuxel (commented):
Couple suggestions.

Two other non-blocking notes.  
1. We should update docs/powershell.md (and docs/docker.md and docker-in-docker.md) with the new arguments.
1. The PowerShell folks are going to give us a more reliable way to get the SHA than HTML parsing going forwards.  This has started to show up in 7.2 preview releases: https://github.com/PowerShell/PowerShell/releases/download/v7.2.0-preview.10/hashes.sha256 - We'll still need the existing fallback for older releases.

Chuxel:
LGTM!

Chuxel (approved):

joshspicer:
also needed to add a match for `-.*`

image

PR #1085, opened 2021-09-30
update git/python/azure scripts

joshspicer:


script-library/python-debian.sh
```diff
@@ -275,9 +275,40 @@ install_using_oryx() {
     ln -s ${PYTHON_INSTALL_PATH}/bin/python3-config ${PYTHON_INSTALL_PATH}/bin/python-config
 }
 
+link_existing_python() {
```
joshspicer, line 278: Not sure if I even need to link this.

I'm only symlinking the binary here- I was going to do the whole python folder (with `python -c "import sys; print(sys.exec_prefix)`, but in some cases the system version is _just_ a binary in `/usr`.

script-library/python-debian.sh
```diff
@@ -278,6 +278,18 @@ install_using_oryx() {
...
+# General requirements
+check_packages curl ca-certificates gnupg2 tar make gcc libssl-dev zlib1g-dev libncurses5-dev \
+            libbz2-dev libreadline-dev libxml2-dev xz-utils libgdbm-dev tk-dev dirmngr \
+            libxmlsec1-dev libsqlite3-dev libffi-dev liblzma-dev uuid-dev 
+
+# If the os-provided versions are "good enough", detect that and bail out.
+if [ ${PYTHON_VERSION} = "os-provided" ] || [ ${PYTHON_VERSION} = "system" ] ; then
+    check_packages python3 python3-doc python3-pip python3-venv python3-dev python3-tk
+    echo "Done!"
+    exit 0
```
Chuxel, line 290: We should still install the python tools (e.g. pylint) in the lower section.

script-library/container-features/src/features.json
```diff
@@ -233,13 +233,13 @@
...
-			"name": "Git (latest version - may require compilation)",
+			"name": "Git (may require compilation)",
 			"options": {
 				"version": {
 					"type": "string",
-					"enum": ["latest"],
-					"default": "latest",
-					"description": "Currently unused."
+					"proposals": ["latest", "os-provided"],
+					"default": "system",
```
Chuxel, line 241: Need to tweak this to be os-provided

script-library/container-features/src/features.json
```diff
@@ -580,9 +580,9 @@
 			"options": {
 				"version": {
 					"type": "string",
-					"enum": [ "latest", "3.9", "3.8", "3.7", "3.6" ],
-					"default": "latest",
-					"description": "Select a Python version to install"
+					"enum": [ "latest", "os-provided", "3.9", "3.8", "3.7", "3.6" ],
+					"default": "system",
```
Chuxel, line 584: Ditto here os-provided

script-library/container-features/src/features.json
Chuxel:
```suggestion
					"default": "os-provided",
```

script-library/container-features/src/features.json
Chuxel:
```suggestion
					"default": "os-provided",
```

Issue #1084, opened 2021-09-30
Install compose-switch in docker-in-docker and docker-from-docker scripts to enable v1/v2 compose selection


Chuxel:
As discussed in microsoft/vscode-remote-release#5647, https://github.com/docker/compose-switch is a project that provides a way for users to pick between Compose v1 and v2 if both are installed.

moby-compose installs v2  as `docker compose` when using Moby, and the script now also locks to v1 for `docker-compose`. The switch will allow us allow users to select which one they prefer at run time.

PR #1087, opened 2021-10-01
Install compose-switch, default to Compose v2

Chuxel:
Implements #1084

This PR updates the `docker-in-docker-debian.sh` and `docker-debian.sh` scripts to ensure that the Compose V2 CLI plugin is installed, and then installs https://github.com/docker/compose-switch to enable its use from the `docker-compose` command in backwards compatibility mode.

It also still installs the Compose v1 docker-compose command and sets it up as an alternative so developers can switch back to V1 if they run into trouble - particularly as V2 is relatively new.


script-library/docker-debian.sh
```diff
@@ -179,40 +194,80 @@ else
...
+# Install Docker Compose v1 if not already installed
+if [ "${DOCKER_COMPOSE_V1_VERSION}" != "none" ]; then
+    if type docker-compose > /dev/null 2>&1; then
+        echo "(*) Compose v1 already installed. Skipping..."
+    else
+        if [ "${target_compose_arch}" != "x86_64" ]; then
+            # Use pip to get a version that runns on this architecture
+            echo "(*) Installing Compose v1..."
+            if ! dpkg -s python3-minimal python3-pip libffi-dev python3-venv > /dev/null 2>&1; then
+                apt_get_update_if_needed
```
joshspicer, line 225: I wonder if here (in another PR) would make sense to call out to the python script?

script-library/docker-debian.sh
```diff
@@ -15,8 +15,11 @@ TARGET_SOCKET=${3:-"/var/run/docker.sock"}
 USERNAME=${4:-"automatic"}
 USE_MOBY=${5:-"true"}
 DOCKER_VERSION=${6:-"latest"}
+DOCKER_COMPOSE_PLUGIN_VERSION=${7:-"latest"}
```
joshspicer, line 18: Do we want to update the features.json and .env file with these new flags?

joshspicer (approved):
lgtm!

script-library/docker-debian.sh
```diff
@@ -124,6 +127,18 @@ fi
 . /etc/os-release
 # Fetch host/container arch.
 architecture="$(dpkg --print-architecture)"
+target_compose_arch="${architecture}"
```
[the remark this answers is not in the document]
Chuxel: Yeah this situation is a bit odd because certain things want what dpkg --print-architecture returns (e.g. amd64 or arm64) while others want a subset of what uname provides. (e.g. aarch64 -- though uname can also return amd64).

script-library/docker-debian.sh
Chuxel: Yeah I was debating that.  Some of these seem like they'd be "hidden" fields as we have UX that allows you to select options. Probably does make sense to add them at the end of the day.

PR #1119, opened 2021-10-20
Clarify that Graphic UI apps open automatically

wrobelda:
UI is too vague – after all, CLI is also a UI. What you mean here is that *GUI* apps open automatically inside the desktop.


Chuxel:
Fair enough. Thanks!

PR #1142, opened 2021-10-29
Adding extension to devcontainer for Salesforce Device Code Auth Flow via Command Palette

gfarb:
When using GitHub Codespaces in the browser, all authentication flows provided by Salesforce via their Extension Package do not work because of URL callback. This causes confusion, primarily for folks in the Salesforce ecosystem who do not use Salesforce's CLI.

The extension that I have added to .devcontainer was created and open-sourced by me. The extension simply provides a seamless and intuitive way to leverage Salesforce's Device Code auth flow which is the only known way to authenticate with a Salesforce org while using GitHub Codespaces in the browser.

The code can be found [here](https://github.com/gfarb/sfdx-codepsaces-auth). I would understand if this PR can't be merged due to security concerns, but it would make GitHub Codespaces onboarding a lot easier for many people in the Salesforce ecosystem.

### DIFF
--- a/containers/sfdx-project/.devcontainer/devcontainer.json
+++ b/containers/sfdx-project/.devcontainer/devcontainer.json
@@ -5,7 +5,8 @@
       "salesforce.salesforcedx-vscode",
       "redhat.vscode-xml",
       "dbaeumer.vscode-eslint",
-      "esbenp.prettier-vscode"
+      "esbenp.prettier-vscode",
+      "gfarb.sfdx-codespaces-auth"
     ],
     "settings": {
       "salesforcedx-vscode-apex.java.home": "/usr/lib/jvm/java-11-openjdk-amd64"


Chuxel:
@gfarb I'd actually suggest raising this with the salesforce extension team at https://github.com/forcedotcom/salesforcedx-vscode

That's likely the best place to PR this rather than as a workaround here. @dehru What are your thoughts on this?

dehru:
Hello @gfarb.  The Salesforce cli supports device flow with this command.  Does this work for you?

https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_auth_device.htm

gfarb:
@dehru the command does work, but the reason I created the extension is because a lot of folks in the Salesforce ecosystem don't use the CLI and there is no official documentation about Salesforce org authentication in VS Code containers (specifically about the Device Code flow being the only possible way at the moment). Due to the fact that a lot of people rely on the command palette for their Salesforce DX, I thought this addition would be beneficial by making the process more intuitive. The extension simply provides a declarative way to initiate the Device Code flow, and the extension actually uses the command referenced above based on user input.

@Chuxel if you and @dehru prefer, I can certainly try and get this code merged into salesforcedx-vscode instead in which case we can close this PR

dehru:
@gfarb Oh I See!  Very interesting.  Let me check this out more next week.

dehru:
@gfarb - Please submit this to the salesforce-vscode repo.  I'll make sure our product manager sees this.  It'd be great if you took some screenshots and/or a screen-movie of the interaction and included them in the PR.

gfarb:
@dehru sounds like a plan, thank you for for the great communication!

PR #1146, opened 2021-10-31
Base rust dev-containers directly off debian

eggyal:
The `debian-rust.sh` library script that is invoked while building the
rust image installs rustup & rustc—so there's no need to base off the
official Rust Docker image, as that's all it does too.


Chuxel:
Thanks for the consideration, but what benefit do end users get from making this switch?  

The slim image is going to will have different contents than default rust one which can break existing users, and any change comes with some amount of risk.  Also, likely the one to use is the mcr.microsoft.com/vscode/devcontainers/base where common are already installed which would help with layering.

eggyal:
I suppose the main benefit is circa 25% reduction in image size (which is pretty significant on a 2GiB image), but besides that it probably isn't all that important.

Although it did strike me somewhat strange that these VS Code images are installing Rust atop an image that already has Rust installed: when I noticed that I initially thought there might be two installations present, but it appears they both installed to the same locations (so the second overwrote they first).  The `$PATH` does end up including `$CARGO_HOME/bin` twice however, which leads to some peculiar (albeit by no means serious) behaviours.

I just suppose from the mere standpoint of correctness, these images should *either* should use the Rust installation provided by the official Rust images (so not run `library-scripts/debian-rust.sh`) or else base off something that doesn't include Rust (such as, as you say, the `base` image).

I'm certainly not attached to this PR and you're welcome to close it if it's not useful.

Chuxel:
@eggyal The scripts are here to support the new features capability and for one off use in your own images. We use these same scripts in these images to ensure what we've got is consistent, but that's not their only goal. Indeed, if you use the pre-built image, you never even see the script nor need it.

Generally though, you'll see the slim images are generally stated as things you use when you only need to run one thing. What often happens is many of these dependencies end up getting added back in to make it a good development image that you work inside of.

All of the non-slim images use `buildpack-deps` so you get layer consistency and reuse. We opted to go with `buildpack-deps:curl` for the `base` images since they're a bit smaller.

We could opt to also create `slim` variants of each image if there's demand. There's already a "slim" variation of the base rust image as well.

Ui0iU (approved):

PR #1150, opened 2021-11-01
fix: change user before run go get/install

giautm:
change user before run go get/install to avoid permission denied on `/go/pkg/mod/cache`.
```
go: writing go.mod cache: mkdir /go/pkg/mod/cache/download/go.uber.org: permission denied
go: writing go.mod cache: mkdir /go/pkg/mod/cache/download/go.uber.org: permission denied
```

Another way to fix the issue is run go commands as vscode
```dockerfile
RUN su vscode -c "go get -x <your-dependency-or-tool>"
```


Chuxel:
Thanks!

PR #1155, opened 2021-11-09
Update to latest dind script, reduce overhead

Chuxel:
This PR updates to the latest docker-in-docker script from the github.com/moby/moby (and docker/docker) repo. The script update improves things by moving all processes in the root cgroup to the init cgroup instead of just process 1. Not doing this causes errors in some cases - I stubmled on this when experimenting with using "lima" and "colima" to host Docker, but there are other cases that could happen whenever running directly on Linux as the host.

Given the nature of this update, it also switches from using lots of "sudo" statments when required to one sudo statement that is only fired of the main container user is non-root. It then also switches to /bin/sh to further improve performance given it is lower overhead.


joshspicer (approved):
you'll want to copy these changes to the same file in `./library-scripts` directory, no?

Chuxel:
@joshspicer Ack - Committed the wrong file. Good catch.

Issue #986, opened 2021-08-13
Ubuntu container: Commands take 10 seconds to return control to bash


Rappel87:
- VSCode Version: 1.58.2
- Local OS Version: MacOS 11.5.1
- Name of Dev Container Definition with Issue: Ubuntu focal from https://github.com/microsoft/vscode-dev-containers/blob/main/containers/ubuntu/.devcontainer/base.Dockerfile

Steps to Reproduce:

1. Create a dev container for my Git project purely based on the Dockerfile above
2. Start devcontainer and type a command like `ls` in the containers terminal and the command takes 10 seconds to return control back to the terminal. This seems to be related to #867. On MacOS, the issue seems to arise from the git index. If I run `git status`, the problems disappears. On Windows, `git status` does not seem to help. 
I had a look at `__bash_prompt()` and could identify `if git ls-files --error-unmatch -m --directory --no-empty-directory -o --exclude-standard ":/*"` as root cause. I deactivated the `__bash_prompt()` and did some measurements:
```bash
vscode@87a39af14ba2:/workspaces/WLM$ time git ls-files --error-unmatch -m --directory --no-empty-directory -o --exclude-standard ":/*"
find_dup.perl
.devcontainer/devcontainer.json

real    0m10.230s
user    0m3.704s
sys     0m0.859s
vscode@87a39af14ba2:/workspaces/WLM$ time git status
Refresh index: 100% (11513/11513), done.
[...] (removed)

real    0m11.370s
user    0m3.925s
sys     0m0.948s
vscode@87a39af14ba2:/workspaces/WLM$ time git ls-files --error-unmatch -m --directory --no-empty-directory -o --exclude-standard ":/*"
find_dup.perl
.devcontainer/devcontainer.json

real    0m0.088s
user    0m0.012s
sys     0m0.035s

Chuxel:
@Rappel87 Are you using GitHub Codespaces or Remote - Containers?

//cc: @joshspicer @jkeech

Rappel87:
@Chuxel I am using the Remote - Containers VSCode extension.

janitorr:
Oops,. I think I might have just created a [dublicate](https://github.com/microsoft/vscode-dev-containers/issues/999) for this issue.

Anyhow, it seems to be caused by a large amount of branches in our repository (400+), just removing the branchName part from the prompt helped us, but it´s very manual step you need to do on every container using our repo so better alternative would be nice.

#999

[...]

PR #1161, opened 2021-11-11
Allow disable git status, drop omb stub, handle gid in different group name

Chuxel:
This PR resolves https://github.com/microsoft/vscode-dev-containers/issues/986 by allowing users to disable the git stauts part of the default prompt.  

It is modeled after Oh My Zsh!'s use of git config to handle this same situation so you can vary the behavior by repo and you do not need to update global env vars to change it. This also allows it to survive a "rebuild" if set for a repository.

To update:
```
git config codespaces-theme.hide-status 1
```

Or globally (e.g. from dotfiles):
```
git config --global codespaces-theme.hide-status 1
```

This works with both the bash and ZSH prompts, though OMZ has its own setting that also works.

I also: 

1. Removed the old "Oh My Bash" stub which was removed 9 months ago since enough time has passed.

2. Handled scenarios where the non-root user's default group is not the same as the username rather than hard coding it (this was a problem initally for the jupyter-datascience-notebook image).

### DIFF
--- a/script-library/common-alpine.sh
+++ b/script-library/common-alpine.sh
@@ -120,8 +120,9 @@ fi
 # Create or update a non-root user to match UID/GID.
 if id -u ${USERNAME} > /dev/null 2>&1; then
     # User exists, update if needed
-    if [ "${USER_GID}" != "automatic" ] && [ "$USER_GID" != "$(id -G $USERNAME)" ]; then 
-        groupmod --gid $USER_GID $USERNAME 
+    if [ "${USER_GID}" != "automatic" ] && [ "$USER_GID" != "$(id -g $USERNAME)" ]; then 
+        default_group_name="$(id -gn $USERNAME)"
+        groupmod --gid $USER_GID ${default_group}
         usermod --gid $USER_GID $USERNAME
     fi
     if [ "${USER_UID}" != "automatic" ] && [ "$USER_UID" != "$(id -u $USERNAME)" ]; then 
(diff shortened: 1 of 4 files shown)


joshspicer (approved):

joshspicer:
This is great! We still don't want to close that linked issue though

Chuxel:
@joshspicer In terms of leaving the issue open, are you thinking about maintaining visibility or going down the path of a broader fix?

jacobwoffenden:
Hi @Chuxel,

I've been `script-library/common-debian.sh` in my devcontainer build, and I think this has introduced the following error

```
docker run -it --rm ghcr.io/maglevlabs/sre@sha256:54aa01385a51c83eb8dc005bb40b891cc562956dc2cfc6fdeb9237a386a09c20
zsh: parse error: condition expected: !=                                                                                                                                     
sre ➜ ~ $ 
```

Whereas the version directly before doesn't

```
docker run -it --rm ghcr.io/maglevlabs/sre@sha256:c18e9d5d6ad4b2b36d8cfaf66ed59156dea1a5825c0c25c876e900bb6a22a02b
sre ➜ ~ $ 
```

https://github.com/microsoft/vscode-dev-containers/commit/3e77cb2501905f1de5e6e0f3dea98f2a7a697425#diff-6cf07d9e0c4858db352327b79b45cfdad0e81ca06d0b7c358b3503da6a55768aR338-R339

Chuxel:
@jacobwoffenden  Thanks. Should be resolved now.

jacobwoffenden:
Yes, thanks @Chuxel! I'll stick to the released versions from now on

PR #1163, opened 2021-11-14
fix(script-library): kubectl zsh completion file permission

hannoeru:
Allow user to replace `.oh-my-zsh` folder.
Useful when install `dotfiles` repo in `codespace`, it run `dotfiles` repo's install script with non-root user.

### DIFF
--- a/script-library/kubectl-helm-debian.sh
+++ b/script-library/kubectl-helm-debian.sh
@@ -140,10 +140,14 @@ fi
 kubectl completion bash > /etc/bash_completion.d/kubectl
 
 # kubectl zsh completion
-zsh_completion_dir="/home/${USERNAME}/.oh-my-zsh/completions"
-mkdir -p "$zsh_completion_dir"
-kubectl completion zsh > "$zsh_completion_dir/_kubectl"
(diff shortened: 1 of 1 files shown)


Chuxel:
Thanks - this makes sense, however, I noticed `group_name` is not set anywhere. Were you meaning to do something like `group_name=$(id -g $USERNAME)`?

script-library/kubectl-helm-debian.sh
```diff
@@ -140,8 +140,13 @@ fi
 kubectl completion bash > /etc/bash_completion.d/kubectl
 
 # kubectl zsh completion
-mkdir -p /home/${USERNAME}/.oh-my-zsh/completions
-kubectl completion zsh > /home/${USERNAME}/.oh-my-zsh/completions/_kubectl
+if [ ! -z $USERNAME  ]; then
```
Chuxel, line 143: USERNAME will always be set, but may be root, so this will do what you want I think:
```suggestion
if [ "$USERNAME" != "root"  ]; then
```

Chuxel:
LGTM!

PR #1167, opened 2021-11-16
Add cmake-tools to default C++ container?

jureid:
@esweet431 - I'm wondering if we should add CMake Tools to VS Code's default C++ dev container. The Dockerfile installs cmake. @bobbrow - Do you know if we do this for Codespaces?

### DIFF
--- a/containers/cpp/.devcontainer/devcontainer.json
+++ b/containers/cpp/.devcontainer/devcontainer.json
@@ -13,7 +13,8 @@
 
 	// Add the IDs of extensions you want installed when the container is created.
 	"extensions": [
-		"ms-vscode.cpptools"
+		"ms-vscode.cpptools",
+		"ms-vscode.cmake-tools"
 	],
 
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.


bobbrow:
It makes sense to me.  I thought CodeSpaces included it, but if they use this container, perhaps my memory is wrong.

esweet431:
Yeah, if we're installing cmake.exe then I agree it makes sense to install cmake-tools. I'm assuming the dockerfile already includes cpptools?

esweet431:
> I'm assuming the dockerfile already includes cpptools?

Nvm sorry, I see it in the JSON

jureid:
> It makes sense to me. I thought CodeSpaces included it, but if they use this container, perhaps my memory is wrong.

I also thought they included it but I can't find it in their configuration files: https://github.com/microsoft/vscode-dev-containers/tree/main/containers/codespaces-linux

Adding @Chuxel to review this PR. Also, Chuck, do you know if the default Codespace configuration includes cmake-tools?

Chuxel (approved):

PR #1174, opened 2021-11-24
Fix Node script checksum verification

neilvilela:
The checksum verification is comparing the node-debian.sh to the
common-debian.sh checksum, which fails if the `COMMON_SCRIPT_SHA`
environment variable is set.

This PR fixes the issue, comparing the node-debian.sh checksum to
the `NODE_SCRIPT_SHA` environment variable.


Chuxel:
Thanks!

PR #1175, opened 2021-11-26
Use variable for architecture when installing rust.

jbreiding:
Recently encountered a failure adding the rust layer on an M1 Mac.

This was a fairly simple fix.


Chuxel:
Good catch. Thanks!

PR #1176, opened 2021-11-28
Add Python devcontainer with CUDA support

### LINKED ISSUE #180: Add Pytorch example
It may be a little tricky, but I think should be possible, it can probably be based on [this nvidia container](https://docs.nvidia.com/deeplearning/frameworks/user-guide/index.html):

twsl:
Adds a devcontainer based on the Python devcontainer with CUDA support.
Uses the nvidia/cuda image as base image.
Enables developers with focus on ML/DL to leverage devcontainers.

Could fix #180


twsl:
Hey @Chuxel, can I get some feedback on the PR?

mimoralea:
Please, feedback on this would be appreciated!

Chuxel:
A few thoughts looking at this that partly caused some delay in response:

1. One challenge is that this would only work on Linux and WSL and specifically with an Nvidia GPU. The challenge is that there is really not a great way to set that expectation in the existing user experience. It would need to be directly in the description for it to appear.  There is an [upcoming capability](https://github.com/microsoft/dev-container-spec/issues/6) that could allow for two definitions that you can pick from - one GPU, one w/o a GPU - but I'm assuming that wouldn't work here, correct?  Otherwise, it might make sense to wait for that to be available.

2. As this is setup right now, it seems to assume that Microsoft would pick up pre-building the image as well. That's a higher bar for merging since it has long term implications for support, patching, etc. It might be better to refactor this so that the "base.Dockerfile" is the Dockerfile and updating the README to make this a bit clearer.  Generally we've been adding `(Community)` by community managed definitions both in the README and devcontainer.json name as well.

3. Is installing Node.js needed in this case if the focus is ML?

twsl:
Hey @Chuxel 
the first issue you raised is just the harsh truth: NVIDIA and CUDA dominate the field of machine learning. Most people in the field who could leverage such a container are already used to that and I don't think it would cause wrong expectations.
I'd like to add macOS support though, but I do not have any experience with that at all and do not own a mac for testing.
The mentioned capability could simplify the workload because this is intended as a an addition to the base python container. 
However, in addition to using the GPU, the base image would need to change and without an NVIDIA GPU that supports CUDA, this image would just be the base python container with additional, useless drivers. So i think, a dedicated image might make more sense.

What exactly is required for the refactor, any specific changes you have in mind? Feel free to request changes in a review.

Node.js has been added for parity with the base python devcontainer, so a developer can simply switch to the cuda devcontainer without breaking existing code. Workloads such as data visualization can require node.js.

mkarikom:
@twsl your base.Dockerfile is working brilliantly, thanks!

sidecus:
@Chuxel @twsl any update on this?

PR #1178, opened 2021-11-30
Move to Microsoft OpenJDK

### LINKED ISSUE #1050: Consider moving to Microsoft OpenJDK builds
Currently the "java" definition uses AdoptOpenJDK.  Microsoft now has builds for JDK 11 and 16 we could use instead.

https://docs.microsoft.com/en-us/java/openjdk/containers

@akaroml @testforstephen - Would this be the VS Code Java team's go forward recommendation?

//cc: @joshspicer

testforstephen:
Signed-off-by: Jinbo Wang <jinbwan@microsoft.com>

Fixes #1050

Use debian as the base image, and install Microsoft OpenJDK 11/17 via tar.gz download url on the docker image.

### DIFF
--- a/containers/java/.devcontainer/base.Dockerfile
+++ b/containers/java/.devcontainer/base.Dockerfile
@@ -1,34 +1,53 @@
 # This base.Dockerfile uses separate build arguments instead of VARIANT
+ARG TARGET_JAVA_VERSION=11
 ARG BASE_IMAGE_VERSION_CODENAME=bullseye
 FROM mcr.microsoft.com/vscode/devcontainers/base:${BASE_IMAGE_VERSION_CODENAME}
+
+ARG TARGET_JAVA_VERSION
+ENV JAVA_HOME /usr/lib/jvm/msopenjdk-${TARGET_JAVA_VERSION}
+ENV PATH "${JAVA_HOME}/bin:${PATH}"
+# Default to UTF-8 file.encoding
+ENV LANG en_US.UTF-8
+
+# Install Microsoft OpenJDK
+RUN arch="$(dpkg --print-architecture)" \
+	&& case "$arch" in \
+		"amd64") \
+			jdkUrl="https://aka.ms/download-jdk/microsoft-jdk-${TARGET_JAVA_VERSION}-linux-x64.tar.gz"; \
+			;; \
+		"arm64") \
+			jdkUrl="https://aka.ms/download-jdk/microsoft-jdk-${TARGET_JAVA_VERSION}-linux-aarch64.tar.gz"; \
+			;; \
+		*) echo >&2 "error: unsupported architecture: '$arch'"; exit 1 ;; \
+	esac \
+	\
+	&& wget --progress=dot:giga -O msopenjdk.tar.gz "$jdkUrl" \
+	&& mkdir -p "$JAVA_HOME" \
+	&& tar --extract \
+		--file msopenjdk.tar.gz \
+		--directory "$JAVA_HOME" \
+		--strip-components 1 \
+		--no-same-owner \
+	&& rm msopenjdk.tar.gz* \
+	\
+	&& ln -s ${JAVA_HOME} /docker-java-home
 
 # Copy library scripts to execute
 COPY library-scripts/*.sh library-scripts/*.env /tmp/library-scripts/
 
 ARG USERNAME=vscode
-# JDK Version
-ARG TARGET_JAVA_VERSION=11
 # [Option] Install Maven
 ARG INSTALL_MAVEN="false"
 ARG MAVEN_VERSION=""
 # [Option] Install Gradle
 ARG INSTALL_GRADLE="false"
 ARG GRADLE_VERSION=""
-ENV LANG en_US.UTF-8
 ENV SDKMAN_DIR="/usr/local/sdkman"
 ENV PATH="${SDKMAN_DIR}/candidates/java/current/bin:${PATH}:${SDKMAN_DIR}/candidates/maven/current/bin:${SDKMAN_DIR}/candidates/gradle/current/bin"
-# Install Java tools such as JDK, Maven and Gradle
-RUN bash /tmp/library-scripts/java-debian.sh $TARGET_JAVA_VERSION "${SDKMAN_DIR}" "${USERNAME}" "true" \
+RUN bash /tmp/library-scripts/java-debian.sh "none" "${SDKMAN_DIR}" "${USERNAME}" "true" \
 	&& if [ "${INSTALL_MAVEN}" = "true" ]; then bash /tmp/library-scripts/maven-debian.sh "${MAVEN_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
 	&& if [ "${INSTALL_GRADLE}" = "true" ]; then bash /tmp/library-scripts/gradle-debian.sh "${GRADLE_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
 	&& apt-get clean -y && rm -rf /var/lib/apt/lists/*
-
-# Add symlinks to the installed JDK
-RUN ls ${SDKMAN_DIR}/candidates/java | grep "^${TARGET_JAVA_VERSION}.*" > /tmp/library-scripts/jdkname \
-	&& jdkName="$(cat /tmp/library-scripts/jdkname)" \
-	&& ln -s ${SDKMAN_DIR}/candidates/java/${jdkName} /docker-java-home \
-	&& if [ ! -d "/usr/lib/jvm" ]; then mkdir /usr/lib/jvm ; fi \
-	&& ln -s ${SDKMAN_DIR}/candidates/java/${jdkName} /usr/lib/jvm/${jdkName}
 
 # [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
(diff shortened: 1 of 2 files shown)


testforstephen:
cc @brunoborges as well.

containers/java/.devcontainer/base.Dockerfile
```diff
@@ -1,7 +1,10 @@
 # This base.Dockerfile uses separate build arguments instead of VARIANT
-ARG TARGET_JAVA_VERSION=11
 ARG BASE_IMAGE_VERSION_CODENAME=bullseye
-FROM openjdk:${TARGET_JAVA_VERSION}-jdk-${BASE_IMAGE_VERSION_CODENAME}
+FROM debian:${BASE_IMAGE_VERSION_CODENAME}
```
Chuxel, line 3: We can actually remove the "common" script from here by using `mcr.microsoft.com/vscode/devcontainers/base:${BASE_IMAGE_VERSION_CODENAME}`. These are the `debian` and `ubuntu` images with the common script already run and the user as `vscode`.

containers/java/.devcontainer/base.Dockerfile
```diff
@@ -15,8 +18,7 @@ ARG USERNAME=vscode
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 RUN bash /tmp/library-scripts/common-debian.sh "${INSTALL_ZSH}" "${USERNAME}" "${USER_UID}" "${USER_GID}" "${UPGRADE_PACKAGES}" "true" "true" \
```
Chuxel, line 20: Same comment as above, using the `base` image would allow us to skip this step.

Chuxel (commented):
A couple initial suggestions.

containers/java/.devcontainer/base.Dockerfile
```diff
@@ -26,18 +28,26 @@ ARG INSTALL_GRADLE="false"
...
-    && if [ "${INSTALL_GRADLE}" = "true" ]; then bash /tmp/library-scripts/gradle-debian.sh "${GRADLE_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
-    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+# Install Java tools such as JDK, Maven and Gradle
+RUN bash /tmp/library-scripts/java-debian.sh $TARGET_JAVA_VERSION "${SDKMAN_DIR}" "${USERNAME}" "true" \
+	&& if [ "${INSTALL_MAVEN}" = "true" ]; then bash /tmp/library-scripts/maven-debian.sh "${MAVEN_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
+	&& if [ "${INSTALL_GRADLE}" = "true" ]; then bash /tmp/library-scripts/gradle-debian.sh "${GRADLE_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
+	&& apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
+# Add symlinks to the installed JDK
+RUN ls ${SDKMAN_DIR}/candidates/java | grep "^${TARGET_JAVA_VERSION}.*" > /tmp/library-scripts/jdkname \
```
juzuluag, line 38: Wouldn't be better to use the image mcr image as suggested by [ms docs](https://docs.microsoft.com/en-us/java/openjdk/containers#use-a-different-base-os-image) ?

containers/java/.devcontainer/base.Dockerfile
```diff
@@ -26,18 +28,26 @@ ARG INSTALL_GRADLE="false"
...
-    && if [ "${INSTALL_MAVEN}" = "true" ]; then bash /tmp/library-scripts/maven-debian.sh "${MAVEN_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
-    && if [ "${INSTALL_GRADLE}" = "true" ]; then bash /tmp/library-scripts/gradle-debian.sh "${GRADLE_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
-    && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+# Install Java tools such as JDK, Maven and Gradle
+RUN bash /tmp/library-scripts/java-debian.sh $TARGET_JAVA_VERSION "${SDKMAN_DIR}" "${USERNAME}" "true" \
+	&& if [ "${INSTALL_MAVEN}" = "true" ]; then bash /tmp/library-scripts/maven-debian.sh "${MAVEN_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
+	&& if [ "${INSTALL_GRADLE}" = "true" ]; then bash /tmp/library-scripts/gradle-debian.sh "${GRADLE_VERSION:-latest}" "${SDKMAN_DIR}" ${USERNAME} "true"; fi \
+	&& apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
+# Add symlinks to the installed JDK
```
juzuluag, line 37:
```suggestion
ARG TARGET_JAVA_VERSION=17
ARG BASE_IMAGE_VERSION_CODENAME=bullseye
FROM debian:${BASE_IMAGE_VERSION_CODENAME}

ARG TARGET_JAVA_VERSION
ENV LANG en_US.UTF-8
ENV JAVA_HOME /usr/lib/jvm/msopenjdk-${TARGET_JAVA_VERSION}-amd64
ENV PATH "${JAVA_HOME}/bin:${PATH}"
COPY --from=mcr.microsoft.com/openjdk/jdk:17-ubuntu ${JAVA_HOME} ${JAVA_HOME}
```

This is coming from [MS docs](https://docs.microsoft.com/en-us/java/openjdk/containers#use-a-different-base-os-image)

testforstephen: Since we need to support arm64 architecture as well, we can‘t copy amd64 JDK from the mcr JDK using hardcoding. The tool like sdkman supports installing native arm64 JDKs.

containers/java/.devcontainer/base.Dockerfile
brunoborges: I do recommend to use the `COPY --from=...` instruction instead of SDKMan in this case because we do not manage SDKMan manifest, and there is likely to be a time gap between our release and whenever SDKMan updates the manifest.

containers/java/.devcontainer/base.Dockerfile
brunoborges: @testforstephen but aren't Linux containers on macOS M1 laptops running as regular x64 containers?

testforstephen: https://github.com/microsoft/vscode-dev-containers/blob/36d591b3df00e3d5277c8d806bc57ac28293b5e1/containers/java/definition-manifest.json#L25-L29

The manifest defines multiple architectures support for Java images, the build pipeline will build docker images with buildx for multiple architecture support.

Add @Chuxel for more information about the underlying build pipeline of vscode-dev-containers.

testforstephen: @brunoborges One question for you, does your team publish arm64 JDK to sdkman?

sdkman supports native arm64 JDKs since 5.10.0, see https://github.com/sdkman/sdkman-cli/issues/830#issuecomment-766328300. The sdkman integrated in Java template is the latest version 5.13.1.

brunoborges: It is a 3rd-party package manager we do not manage.

testforstephen: oh, I see, thanks for clarification. I previously thought sdkman's MS OpenJDK was maintained by you guys.

Chuxel: @brunoborges @testforstephen No containers on Apple Silicon macs default to arm64. Docker Desktop can emulate x86, but this is slow. We do build an arm64 version of the bullseye variant for this reason. They get uploaded with the same identifier and the architecture is auto selected.

Is there not an arm64 base image variant for MS JDK? With the recent shifts to MacBook Pros, there's been quite a bit of interest in it.

brunoborges: Yes, we do provide JDK for Linux ARM.

But I'm not sure we have arm packages published on packages.microsoft.com.

The tar.gz can be downloaded from the website though, and simply extracted (and configured).

testforstephen: The latest commit uses Microsoft's download url to get MS OpenJDK tar.gz file and install it on the image.

Chuxel: Nice - Is there a sha or signature we should verify for the download?  It looks like the name of the package could be used to get the sha/sig.  It doesn't seem like there's one for just `17`, but this exists:

```
https://aka.ms/download-jdk/microsoft-jdk-17.0.1.12.1-linux-x64.tar.gz.sig
```

testforstephen: Appending `.sha256sum.txt` to the URL  can download the sha256 checksum file.

https://aka.ms/download-jdk/microsoft-jdk-17-linux-x64.tar.gz.sha256sum.txt

It's coming from [MS Docs](https://docs.microsoft.com/en-us/java/openjdk/download-major-urls#checksums).

Chuxel:
LGTM!!

PR #1184, opened 2021-12-01
[R] Remove aditional terminal profile

eitsupi:
The extension has been changed to add a terminal profile `R Teminal` (https://github.com/REditorSupport/vscode-R/pull/851), so the profile `radian` defined in `devcontainer.json` is no longer needed.

image


Chuxel:
Thanks!!

PR #1191, opened 2021-12-02
Add arm64 support to Hugo container

5t111111:
Recently `javascript-node`, the base image of Hugo container, [supports arm64 architecture](https://github.com/microsoft/vscode-dev-containers/blob/04c2d3ff7950d3f71fd1716ac131497f21c3fea5/containers/javascript-node/README.md).

> x86-64, arm64/aarch64 for bullseye variants

That means you can use Hugo container with arm64 by specifying kind of `16-bullseye` as `NODE_VERSION` argument. However, since it downloads amd64 binary version of `hugo` executable for the moment, hugo does not work natively in arm64 container.

This pull request makes it possible to switch the downloadable binary version by checking container OS processor architecture using `umame`.


Chuxel:
@aarynsmith - This LGTM. Make sense from your perspective as well?

PR #1195, opened 2021-12-06
Fix .npmignore to include common-debian.sh

scottcwang:
mit-scheme's .npmignore was excluding library-scripts/common-debian.sh, which meant that the image couldn't build.

This PR also removes nonexistent files from the .npmignore.


Chuxel:
Thanks!

PR #1202, opened 2021-12-10
Symlink legacy JAVA_HOME location

Chuxel:
The new Microsoft OpenJDK version of the Java image works great! However, one thing I noticed is the current openjdk:bullseye image we have been using has a different JAVA_HOME location.

This PR tweaks JAVA_HOME to match the current images in the event someone has added this exact path to their settings.json for the various Java VS Code extensions instead of the common `/docker_java_home` folder.

### DIFF
--- a/containers/java/.devcontainer/base.Dockerfile
+++ b/containers/java/.devcontainer/base.Dockerfile
@@ -4,7 +4,7 @@ ARG BASE_IMAGE_VERSION_CODENAME=bullseye
 FROM mcr.microsoft.com/vscode/devcontainers/base:${BASE_IMAGE_VERSION_CODENAME}
 
 ARG TARGET_JAVA_VERSION
-ENV JAVA_HOME /usr/lib/jvm/msopenjdk-${TARGET_JAVA_VERSION}
+ENV JAVA_HOME /usr/local/openjdk-${TARGET_JAVA_VERSION}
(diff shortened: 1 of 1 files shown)


containers/java/.devcontainer/base.Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG BASE_IMAGE_VERSION_CODENAME=bullseye
 FROM mcr.microsoft.com/vscode/devcontainers/base:${BASE_IMAGE_VERSION_CODENAME}
 
 ARG TARGET_JAVA_VERSION
-ENV JAVA_HOME /usr/lib/jvm/msopenjdk-${TARGET_JAVA_VERSION}
+ENV JAVA_HOME /usr/local/openjdk-${TARGET_JAVA_VERSION}
```
testforstephen, line 7: On linux, the path `/usr/lib/jvm` is a common location to install JDKs. How about using the old path as a symlink to it for compatibility? e.g. `ln -s ${JAVA_HOME} /usr/local/openjdk-${TARGET_JAVA_VERSION}`

https://github.com/microsoft/vscode-dev-containers/blob/908a56bdba329cdb1c3c3dbe6635b518f6375b21/containers/java/.devcontainer/base.Dockerfile#L38-L41

Issue #920, opened 2021-06-24
Create a "language + database"  definition for all pre-built images where this applies


Chuxel:
The database selected can vary based on what is most popular in the community.  The following are currently missing one:

- [ ] Java
- [ ] Go
- [ ] C++
- [ ] Anaconda
- [ ] Miniconda
- [ ] Rust
- [x] Ruby (though PR #703 should get reviewed here)
- [x] Node
- [x] Python
- [x] PHP
- [x] C#

PR #1206, opened 2021-12-14
Adding PostGreSQL + Java DB and Language Container pair

KanishkT123:
# [Work Item ID](https://github.com/microsoft/vscode-dev-containers/issues/920)

## Description

---

This PR adds:
1. New Database and Language pair for PostgreSQL and Java. 
2. Tests to verify that the Database is running and accepts commands from the Java container.
3. Supporting documentation for running and editing the containers locally

## PR Checklist

---

- [X] I have added tests to cover my changes.
- [X] I ran the lint checks which produced no new errors nor warnings for my changes.

---

- [ ] Yes
- [X] No

## Testing

---

> - Instructions for testing and validation of your code:

All testing can be carried out using the existing test runner, `test.sh`. The [README.md](https://github.com/microsoft/vscode-dev-containers/blob/50419c07d7a7597fe9d553c644097ae1e85d3406/containers/java-postgres/README.md) also contains explicit testing instructions.

>   - What OS was used for testing.
Windows 10, Windows 11. 
>   - Which test sets were used.
N/A
>   - Description of test scenarios that you have tried.
Tested building and compilation, setting up database, connecting to DB Container, logging into DB, sending and receiving data from DB.

---

These changes are primarily additive, taken from the existing Java devcontainer.


dnastrain:
Context to the reviewers: this `java-postgres` container is derived from the existing `java` container, following the 'deltas' between the `python-3` and `python-3-postgres` containers as a 'template'.

containers/java-postgres/README.md
```diff
@@ -0,0 +1,132 @@
...
+7. From here, you can add breakpoints or edit the contents of the `test-project` folder to do further testing.
+
+## Testing the PostgreSQL container
+
+The `docker-compose` file sets up a networked PostgreSQL database that is accessible from the Java Dev Container. The port is forwarded to `localhost:5432` by default, but can be changed in the [devcontainer.json](.devcontainer\devcontainer.json).
+
+1. After starting the Dev Container as above, you can use the **Debug Console** to test the PostgreSQL Database.
+2. The [AppTest.java](test-project\src\test\java\com\mycompany\app\AppTest.java) contains a Test Method, `testIP` which will ping the Postgres Database using it's default container name, `postgresdb`.
+3. Running this test will let you know that the PostgreSQL DB is accessible. **This does not make or authorize a connection to the database, only checks for connectivity between containers.**
+4. Alternatively, running `./test.sh` will also run all the connectivity tests and verify that the PostgresDB is actually accessible.
```
esgraham, line 126: I was able to run the test.sh bash script, but I had to find it first (under the test-project folder).

containers/java-postgres/README.md
```diff
@@ -0,0 +1,132 @@
...
+5. After the folder has opened in the container, press <kbd>F5</kbd> to start the project.
+6. You should see "Hello Remote World!" in the a Debug Console after the program executes.
+7. From here, you can add breakpoints or edit the contents of the `test-project` folder to do further testing.
+
+## Testing the PostgreSQL container
+
+The `docker-compose` file sets up a networked PostgreSQL database that is accessible from the Java Dev Container. The port is forwarded to `localhost:5432` by default, but can be changed in the [devcontainer.json](.devcontainer\devcontainer.json).
+
+1. After starting the Dev Container as above, you can use the **Debug Console** to test the PostgreSQL Database.
+2. The [AppTest.java](test-project\src\test\java\com\mycompany\app\AppTest.java) contains a Test Method, `testIP` which will ping the Postgres Database using it's default container name, `postgresdb`.
```
esgraham, line 124: I could not get the testIP method to run from the Debug Console.  Either the launch file is missing the debug info, or there are some steps missing here.

esgraham (commented):
I was able to build the container, run the container and see Java and postgres installed.  I was also able to connect to the postgres db through the VS code SQL console.  I was also able to run the test.sh file.

containers/java-postgres/README.md
KanishkT123: Ah, let me add a link to it. I've got links to other things, so I might as well do it here too.

bderusha (changes requested):
Overall this worked great for me and was easy to spin up!  Nice work! A few quick things and this is good to go for me

containers/java-postgres/test-project/test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Prep
+echo -e "\nGetting Maven wrapper..."
+curl -sSL https://github.com/takari/maven-wrapper/archive/maven-wrapper-0.5.5.tar.gz| tar -xzf - 
+mv maven-wrapper-maven-wrapper-0.5.5/mvnw mvnw
+mv maven-wrapper-maven-wrapper-0.5.5/.mvn .mvn
+rm -rf mv maven-wrapper-maven-wrapper-0.5.5
+
+# Definition specific tests
+checkExtension "vscjava.vscode-java-pack"
```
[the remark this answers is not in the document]
dnastrain: nope, was just an oversight, adding and re-testing now

containers/java-postgres/test-project/test-utils.sh
```diff
@@ -0,0 +1,149 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
dnastrain: thx @bderusha, I think the comments are addressed now, and @KanishkT123 can mark them as resolved

bderusha (approved):
:shipit:

containers/java-postgres/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# [Choice] Java version (use -bullseye variants on local arm64/Apple Silicon): 11, 17, 11-bullseye, 17-bullseye, 11-buster, 17-buster
```
gfmatthews, line 1: non-blocking nit: Just drop the "local" bit? This change is intended for any arm64 setup
```suggestion
# [Choice] Java version (note: use -bullseye variants on arm64/Apple Silicon): 11, 17, 11-bullseye, 17-bullseye, 11-buster, 17-buster
```

containers/java-postgres/.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+services:
+  app:
+    container_name: javadev
+    build: 
+      context: .
+      dockerfile: Dockerfile
+      args:
+        # Update 'VARIANT' to pick an version of Java: 11, 17.
+        # Append -bullseye or -buster to pin to an OS version.
+        # Use -bullseye variants on local arm64/Apple Silicon.
```
gfmatthews, line 15: nit: Do we need the "local" qualifier? 
```suggestion
        # Use -bullseye variants on arm64/Apple Silicon.
```

containers/java-postgres/README.md
gfmatthews: +1 I was also able to get this to work

gfmatthews (approved):
Super cool to get this on my machine - looks great! I have a terminology question that perhaps @esgraham can help approve / deny in my comments around how we refer to arm64 hardware but that's it.

containers/java-postgres/.devcontainer/Dockerfile
dnastrain: hi @gfmatthews and @esgraham, that line of text was actually taken verbatim from the `Dockerfile` from the [java folder](https://github.com/microsoft/vscode-dev-containers/blob/main/containers/java/.devcontainer/)

containers/java-postgres/.devcontainer/docker-compose.yml
dnastrain: Well, that line of text was actually taken verbatim from the `docker-compose.yml` in the [python-3-postgres folder](https://github.com/microsoft/vscode-dev-containers/blob/main/containers/python-3-postgres/.devcontainer/)

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,44 @@
...
+import java.net.UnknownHostException;
+
+public class AppTest
+{
+    App newApp = new App();
+    public AppTest() {
+    }
+
+    @Test
+    public void testApp()
```
juzuluag, line 24: nit: Are these necessary ? Seems like it is not testing anything

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,44 @@
...
+    {
+        assertTrue( true );
+    }
+
+    @Test
+    public void testMore()
```
juzuluag, line 30: Same comment as testApp()

containers/java-postgres/test-project/src/main/java/com/mycompany/app/App.java
```diff
@@ -0,0 +1,34 @@
+/*-------------------------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+ *-------------------------------------------------------------------------------------------------------------*/
+
+package com.mycompany.app;
+
+import static javax.xml.XMLConstants.XML_NS_PREFIX;
+import java.io.*;
```
juzuluag, line 9: nit: Avoid using *. Import specific classes only.

juzuluag (commented):
In general LGTM. Just few nits. Good job.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
dnastrain: that was an existing @Test from the `java` folder that we left in place.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
dnastrain: that was an existing @Test taken from the `java` folder that we left in place.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
juzuluag: Technically you don't need `main` at all in the `App.java` file. Just having a class with the static method is good enough.

containers/java-postgres/.devcontainer/Dockerfile
esgraham: I searched through the repository and the setup is listed as "local arm64/Apple Silicon" on all dockerfiles, so we can leave as is.

containers/java-postgres/.devcontainer/docker-compose.yml
esgraham: I searched through the repository and the setup is listed as "local arm64/Apple Silicon" on all dockerfiles, so we can leave as-is.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,44 @@
...
+    {
+        assertTrue( true );
+    }
+
+    @Test
+    public void testIP()
+    {
+        try {
+            assertTrue( App.pingAddress("postgresdb") );
```
xtophs, line 39: What is this test asserting? It seesms to establish that `postgresdb` is a routable name. It doesn't establish that there's a PostgresDb listening on that IP adress. If it's a valid test to assert that the containers are working as expected, then maybe change the error message below?

containers/java-postgres/test-project/src/main/java/com/mycompany/app/App.java
```diff
@@ -0,0 +1,34 @@
...
+import java.net.*;
+
+
+
+public class App 
+{
+    public static void main( String[] args )
+    {
```
xtophs, line 17: How come we're not following the pattern in `dotnet-mssql`? The `dotnet-mssql` app establishes that the database is operational and returns the names of the available databases. 

See: `https://github.com/microsoft/vscode-dev-containers/blob/d7c373f5b497ce8dba496e47b77a25ff34d0cf9e/containers/dotnet-mssql/test-project/Program.cs#L20`

containers/java-postgres/test-project/src/main/java/com/mycompany/app/App.java
```diff
@@ -0,0 +1,34 @@
...
+        System.out.println( "Hello Remote World!" );
+        System.out.println("The XML namespace prefix is: " + XML_NS_PREFIX);
+
+    }
+
+    public static boolean pingAddress(String ipAddress) throws UnknownHostException, IOException
+    {
+        InetAddress postgresAddress = InetAddress.getByName(ipAddress);
```
xtophs, line 25: why is this in `App.java` if it's not invoked by the app itself?

containers/java-postgres/test-project/.gitignore
```diff
@@ -0,0 +1 @@
+target
```
xtophs, line 1: How about adding a standard Java `.gitignore` file that lists non-code files, e.g. *.class.

See here for an example: https://github.com/github/gitignore/blob/main/Java.gitignore

containers/java-postgres/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+			{
+			  "name": "Container database",
+			  "driver": "PostgreSQL",
+			  "previewLimit": 50,
+			  "server": "localhost",
+			  "port": 5432,
+			  // NOTE: database/username/password should match values in docker.compose.yml
+			  "database": "postgres",
+			  "username": "postgres",
+			  "password": "postgres"
```
xtophs, line 19: hardcoding a SQL password feels wrong. The `dotnet-mssql` passes an argument to the `Dockerfile`. Let's follow that pattern.

containers/java-postgres/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+	"extensions": [
+		"vscjava.vscode-java-pack",
+		"mtxr.sqltools",
+		"mtxr.sqltools-driver-pg"
+	],
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [5432],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
```
xtophs, line 35: Do we need those comments? 

The 1st one is explained in `README.md`, the 2nd one is not. Take out or explain in `README.md`?

containers/java-postgres/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+	// "postCreateCommand": "java -version",
+
+	// Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
```
xtophs, line 38: It's already uncommented. Change the comment or comment out the line?

containers/java-postgres/README.md
```diff
@@ -0,0 +1,132 @@
+# Java & PostgreSQL
```
xtophs, line 1: Let's reorder the content of the file to get the reader started quickly.

It's probably most common that a reader is looking for how to start the containers and test that they are running properly. Let's move that content up.

Customization content can come after that. Please take another pass at the contaent abotu customization option and add some structure with headers and/or bulleted lists. Examples when options are needed would make a nice addition.

xtophs (changes requested):

containers/java-postgres/.devcontainer/devcontainer.json
dnastrain: hi @xtophs, I believe we already follow what both `python-3-postgres` and `dotnet-mssql` do -- the "db password" is present in both the `devcontainer.json` and in the `docker-compose.yml` files

containers/java-postgres/.devcontainer/devcontainer.json
dnastrain: hi @xtophs, this is modeled after `python-3-postgres` version, and I think the text in the comment is technically accurate, by suggesting the action required: "comment out connect as root instead"

containers/java-postgres/README.md
dnastrain: hi @xtophs, I think this `java-postgres` readme does follow the suggested [container-template/docker-compose](https://github.com/microsoft/vscode-dev-containers/tree/main/container-templates/docker-compose) version/format, even though ours was also structured after the `java` and `python-3-postgres` ones

containers/java-postgres/test-project/.gitignore
dnastrain: hi @xtophs, agree that could be done -- the current version came from the [java/test-project](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/java/test-project) folder, and we didn't alter it

containers/java-postgres/.devcontainer/devcontainer.json
KanishkT123: +1 This is the existing pattern in almost all the other files as well.

containers/java-postgres/.devcontainer/devcontainer.json
KanishkT123: `postCreateCommand` is a common Docker feature and the comments here are again taken from the existing templates for other `devcontainer.json` files. I'm not sure this actually needs more explanation, since the comment outlines exactly what the command would do.

containers/java-postgres/README.md
KanishkT123: Done!

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
KanishkT123: I agree with you, but again, we're `main`-ly just following the templates that already exist. Most of our changes are additive.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
KanishkT123: +1

containers/java-postgres/test-project/src/main/java/com/mycompany/app/App.java
KanishkT123: I'll change this, good catch

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
KanishkT123: Great point. I'll change the error message to be more prescriptive, I think that would cover it. We figured that the only real test we needed was to make sure that the container networking was set up correctly, and so this test establishes, as you said, that `postgresdb` can be reached.

containers/java-postgres/test-project/.gitignore
KanishkT123: We'll add the gitignore in. Small, QOL improvement.

containers/java-postgres/README.md
KanishkT123: What if we keep the order and add a table of contents? The structure should match the other files to prevent confusion, but by having a ToC we'll be able to help a reader jump to whatever they think is most important.

containers/java-postgres/test-project/src/main/java/com/mycompany/app/App.java
KanishkT123: On further reflection, @dnastrain and I have agreed to try and quickly run an operational test within `App.Java` to see if the PostgreSQL database is up and running.

containers/java-postgres/test-project/src/main/java/com/mycompany/app/App.java
KanishkT123: We'll be changing App.java to add new tests/capability, so that will cover this

containers/java-postgres/README.md
KanishkT123: I added the directions suggested. @dnastrain and I have both tested this with using Right click -> Run Java, and by running individual tests using the green play button.

containers/java-postgres/test-project/src/main/java/com/mycompany/app/App.java
KanishkT123: Done!

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
KanishkT123: Changed the error messages and added more tests to be more specific about what we test.

containers/java-postgres/test-project/src/main/java/com/mycompany/app/App.java
KanishkT123: Done!

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,125 @@
...
+               "postgres", "postgres");
+            return true;
+         } catch (Exception e) {
+            e.printStackTrace();
+            System.err.println(e.getClass().getName()+": "+e.getMessage());
+            return false;
+         }
+    }
+
+    public static boolean listDatabases(String ipAddress, String username, String password){
```
juzuluag, line 53: Seems like `username` and `password` are not used. Hardcoded values are used as part of the `.getConnection()` method.

containers/java-postgres/.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+        VARIANT: 11-bullseye
+        # Options
+        INSTALL_MAVEN: "true"
```
juzuluag, line 18: Isn't this "forcing" the container to install always MAVEN?

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,125 @@
...
+    public void testMore()
+    {
+        assertTrue( true );
+    }
+
+    @Test
+    public void testIP()
+    {
+        try {
+            assertTrue( pingAddress(System.getenv("POSTGRES_IP")) );
```
juzuluag, line 100: Don't see the benefit of invoking static methods into the test method. It would be simpler to just have here the full implementation.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,125 @@
...
+        try {
+            Class.forName("org.postgresql.Driver");
+            c = DriverManager
+               .getConnection("jdbc:postgresql://postgresdb/postgres",
+               "postgres", "postgres");
+            c.setAutoCommit(false);
+            stmt = c.createStatement();
+            ResultSet rs = stmt.executeQuery( "select * from pg_database;" );
+            System.out.println("List of databases in this cluster.");
+            while (rs.next()){
```
juzuluag, line 66: Suggestion: Instead of iterating the result, perhaps with just one call to 'poke' the db would be enough.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
KanishkT123: Ah, good catch. Thank you.

containers/java-postgres/.devcontainer/docker-compose.yml
KanishkT123: It is, but we need maven for the tests to pass. There's no other way to download JDBC right?

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
KanishkT123: I could do that.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
KanishkT123: Well, we're poking the DB with the login anyway and testing the container. This way we can iterate from an actual result set right?

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,125 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.Statement;
+
+public class AppTest
+{
+
+
+    public static boolean pingAddress(String ipAddress) throws UnknownHostException, IOException
```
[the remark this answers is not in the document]
KanishkT123: Correct, but the reason it's there is to have more granular testing. 

PingAddress checks connectivity, login checks credentials and database listing checks to see if it accepts commands. 

So each test builds on the previous.

containers/java-postgres/.devcontainer/docker-compose.yml
dnastrain: upon additional testing, the `test.sh` configures maven first before running the tests, so @juzuluag is correct that we don't need to require maven earlier

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
juzuluag: Yes, but my comment is more towards overhead of iterating and printing the content.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
dnastrain: Fixed

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
dnastrain: hi @juzuluag , part of the reason the 3 methods are static was we wanted them to live in App.java, and be used both by "F5 Debug" of the main app, and be reused the Test-cases -- I ran into an issue getting `./test.sh` to work correctly though, which some research makes me think was apparently due the result of how maven builds the jar..., since the error only would occur when running `test.sh`, or `java -jar target/my-app-1.0-SNAPSHOT.jar` manually, but not when using right-click "Run Java" or "F5 Debug" ... :

🧪 Testing test-project
Hello Remote World!
The XML namespace prefix is: xml
java.lang.ClassNotFoundException: org.postgresql.Driver
at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
at java.base/java.lang.Class.forName0(Native Method)
at java.base/java.lang.Class.forName(Class.java:315)
at com.mycompany.app.App.dbLogin(App.java:55)
at com.mycompany.app.App.main(App.java:30)

containers/java-postgres/README.md
```diff
@@ -0,0 +1,146 @@
...
+  context: ..
+  dockerfile: .devcontainer/Dockerfile
+    args:
+      # Update 'VARIANT' to pick an version of Java: 11, 17.
+      # Append -bullseye or -buster to pin to an OS version.
+      # Use -bullseye variants on local arm64/Apple Silicon.
+      VARIANT: 11-bullseye
+```
+
+You also can connect to PostgreSQL from an external tool when using VS Code by updating `.devcontainer/devcontainer.json` as follows:
```
gfmatthews, line 48: So if I'm a developer using this and wanted to populate my fancy new PostgreSQL database with test data, I'd likely do this using an external tool right? Would it be worth adding a sentence or two in here explaining how someone would go about populating some test data into their dev container instance?

containers/java-postgres/README.md
```diff
@@ -0,0 +1,146 @@
...
+  GRADLE_VERSION: "7.2"
+```
+
+### Installing Node.js
+
+Given JavaScript front-end web client code written for use in conjunction with a Java back-end often requires the use of Node.js-based utilities to build, this container also includes `nvm` so that you can easily install Node.js. You can enable installation and change the version of Node.js installed or disable its installation by updating the `args` property in `.devcontainer/docker-compose.yml`.
+
+```yaml
+args:
+  VARIANT: 11
```
gfmatthews, line 101: Wait, what does the variant arg do here in this context?? The node version is below but what's this?

containers/java-postgres/README.md
```diff
@@ -0,0 +1,146 @@
...
+  NODE_VERSION: "10" # Set to "none" to skip Node.js installation
+```
+
+### Adding the definition to your folder
+
+1. If this is your first time using a development container, please see getting started information on [setting up](https://aka.ms/vscode-remote/containers/getting-started) Remote-Containers or [creating a codespace](https://aka.ms/ghcs-open-codespace) using GitHub Codespaces.
+
+2. Start VS Code and open your project folder or connect to a codespace.
```
gfmatthews, line 109: I would propose cutting this section for the most part and instead referring to the existing public web docs. It's possible that the user flows would change over time and make this documentation go out of date.

containers/java-postgres/test-project/test-utils.sh
```diff
@@ -0,0 +1,149 @@
...
+        [ ! -e $HOME/.vscode-test-server/extensions/${EXTN_ID}* ] || \
+        [ ! -e $HOME/.vscode-remote/extensions/${EXTN_ID}* ]
+    do
+        sleep 1s
+        (( RETRY_COUNT++ ))
+        echo -n "."
+    done
+
+    if [ ${RETRY_COUNT} -lt ${TIMEOUT_SECONDS} ]; then
+        echo -e "\n✅ Passed!"
```
gfmatthews, line 82: Not sure if this code was boilerplate and copy pasted over, but, would this be an opportunity for us to switch to a "log" function similar to what we discussed in our bash style review? There's a few other instances where output is directly logged to the console in this file.

containers/java-postgres/README.md
KanishkT123: We could look into doing that. Do you think it needs to be tool specific?

containers/java-postgres/README.md
KanishkT123: I'm going to be honest, I don't know. This part of the README is taken directly from the existing documentation. Let me take a look and see why it's there. @dnastrain any ideas?

containers/java-postgres/README.md
KanishkT123: I would be against this. Dan and I have been trying to make this document additive instead of subtractive, and we don't want it to state anything different from the current documentation. 

In other words, this is copied from the existing java container.

containers/java-postgres/.devcontainer/devcontainer.json
xtophs: Best practice is to not have unused code, Please remove.

containers/java-postgres/test-project/test-utils.sh
KanishkT123: This code is boilerplate, definitely. I thought about making this neater and more flexible, but I am currently against it because it would be scope creep and would be different from how testing is done in every single other Java package. At this point, consistency, even if it's less "right" is better than inconsistency, in my opinion.

Suggestion: An issue that reworks the test-utils.sh in every Java container file.

containers/java-postgres/.devcontainer/devcontainer.json
KanishkT123: Done.

containers/java-postgres/README.md
KanishkT123: I found it. It's actually in a comment above this snippet.


      # Update 'VARIANT' to pick an version of Java: 11, 17.
      # Append -bullseye or -buster to pin to an OS version.
      # Use -bullseye variants on local arm64/Apple Silicon.

containers/java-postgres/README.md
KanishkT123: Added some guidance on populating test data into the dev container using PSQL.

containers/java-postgres/README.md
gfmatthews: If you've considered it and decided against it, I'm okay with that resolution

containers/java-postgres/.devcontainer/devcontainer.json
dnastrain: sync'ing more offline -- the password only lives in the `devcontainer.json` and `docker-compose.yml` now, so is no longer in the java code.

containers/java-postgres/README.md
```diff
@@ -0,0 +1,151 @@
+# Java & PostgreSQL
+
+- [Java & PostgreSQL](#java--postgresql)
+  - [Summary](#summary)
```
Kpopovs1, line 4: Table of contents idea is great. Makes the section easier to navigate!

containers/java-postgres/README.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+### Installing Node.js
+
+Given JavaScript front-end web client code written for use in conjunction with a Java back-end often requires the use of Node.js-based utilities to build, this container also includes `nvm` so that you can easily install Node.js. You can enable installation and change the version of Node.js installed or disable its installation by updating the `args` property in `.devcontainer/docker-compose.yml`.
+
+```yaml
+args:
+  VARIANT: 11
+  NODE_VERSION: "10" # Set to "none" to skip Node.js installation
```
Kpopovs1, line 107: qq: does *NODE_VERSION* accept specific versions of node? Also does it accept the keyword latest to know when to get the latest version? If yes, may be worth to specify which versions, or that it accepts other args other than version number and "none".

If this is not relevant, or if it's already specified elsewhere in the docs, feel free to ignore or link the other content.

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,86 @@
...
+    {
+        String host = System.getenv("POSTGRES_HOSTNAME");
+
+        InetAddress postgresAddress = InetAddress.getByName(host);
+        System.out.println("Sending Ping Request to " + host);
+        if (postgresAddress.isReachable(5000)){
+            System.out.println("Successfully Reached: " + host);
+            return;
+        }
+        System.out.println("Could not reach or connect to: " + host);
```
Kpopovs1, line 55: qq from my limited Java knowledge: How does this test `pass/fail` if either:
- There isn't an assertion statement on the host being reachable,
- or there is no exception being explicitly thrown if we cannot connect to the host.

containers/java-postgres/test-project/test-utils.sh
```diff
@@ -0,0 +1,149 @@
...
+USERNAME=${1:-vscode}
+
+if [ -z $HOME ]; then
+    HOME="/root"
+fi
+
+FAILED=()
+
+echoStderr()
+{
```
Kpopovs1, line 12: Nit: for consistency, it seems like you are following the `{` format with no newline:

```
function() {

}
```

containers/java-postgres/test-project/test-utils.sh
```diff
@@ -0,0 +1,149 @@
...
+        return 0
+    else
+        echoStderr -e "\n❌ Extension $EXTN_ID not found."
+        FAILED+=("$LABEL")
+        return 1
+    fi
+}
+
+checkCommon()
+{
```
Kpopovs1, line 92: Same comment as above here as well

Kpopovs1 (commented):
Looks great! 🚀

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
juzuluag: @Kpopovs1, the test method is throwing the generic exception (look at the signature), so whatever method call fails inside the test, it is reported as an error.

juzuluag (approved):
+1

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
juzuluag: Oh ok, but @Kpopovs1 , you're right. line 55 should be an assertFalse, since it could be the case that it is not reachable, but it doesn't timeout.
@dnastrain , @KanishkT123 , seems like that is missing here on line 55

containers/java-postgres/test-project/test-utils.sh
dnastrain: `test-utils.sh` is a 'shared' file that's the same in all containers, so we've not made any edits to it

containers/java-postgres/test-project/test-utils.sh
dnastrain: `test-utils.sh` is a 'shared' file that's the same in all containers, so we've not made any edits to it

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
dnastrain: hi @Kpopovs1, the code in place is actually what we landed on upon reviewing it with @juzuluag, where we intentionally are letting the maven+junit test runner interpret test-case exceptions as failed-test results, rather than catch them and then 're-raise a failure' to junit -- and testing w/ juzuluag we also confirmed that exceptions are thrown (by java) upon connectivity failures, which are correctly reported as 'test case failures' by `./test.sh`

containers/java-postgres/README.md
dnastrain: hi @Kpopovs1, well, this MD section showing `NODE_VERSION: "10"` was left to match the 'main' java container version -- all that changed was the location of the arg, which moved to _docker-compose.yml_ from _devcontainer.json_
and yes, one can specify "NODE_VERSION": "lts/*" to 'get latest'

containers/java-postgres/test-project/test-utils.sh
Kpopovs1: gotcha!! Thank you.

containers/java-postgres/.devcontainer/devcontainer.json
dnastrain: I opted to add info about `postCreateCommand` to the README.md, so the details are more 'upfront' for how to use it.

containers/java-postgres/README.md
dnastrain: upon further reflection, I updated the comment in the README.md, so how to use 'latest' is more clear

containers/java-postgres/test-project/src/test/java/com/mycompany/app/AppTest.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Test
+    public void testIP() throws Exception
+    {
+        String host = System.getenv("POSTGRES_HOSTNAME");
+
+        InetAddress postgresAddress = InetAddress.getByName(host);
+        System.out.println("Sending Ping Request to " + host);
+        if (postgresAddress.isReachable(5000)){
```
juzuluag, line 51:
```suggestion
assertTrue(postgresAddress.isReachable(5000));
```
And remove 53 and below.

KanishkT123: Made changes to these tests to use more Asserts in general. Thanks!

containers/java-postgres/.devcontainer/devcontainer.json
KanishkT123: I believe this has been declared to be out of scope for now, unless I'm mistaken? Do we think this can be resolved @dnastrain  and @xtophs ?

containers/java-postgres/README.md
```diff
@@ -0,0 +1,157 @@
+# Java & PostgreSQL
+
+- [Java & PostgreSQL](#java--postgresql)
```
Chuxel, line 3: The metadata is parsed, so this index needs to go beneath it.

containers/java-postgres/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+You also can connect to PostgreSQL from an external tool when using VS Code by updating `.devcontainer/devcontainer.json` as follows:
+
+```json
+"forwardPorts": [ "5432" ]
+```
+
+Once the PostgreSQL container has port forwarding enabled, it will be accessible from the Host machine at `localhost:5432`. The [PostgreSQL Documentation](https://www.postgresql.org/docs/14/index.html) has:
+
+1. [An Installation Guide for PSQL], a CLI tool to work with a PostgreSQL database.
```
Chuxel, line 56: Was there intended to be a link here given `[ ]`? If so it appears to be missing.

Chuxel (commented):
A couple of minor edits to the readme, otherwise LGTM!

containers/java-postgres/README.md
KanishkT123: Ah, thank you. I can do that!

containers/java-postgres/README.md
KanishkT123: There was, thank you. Let me fix that.

Chuxel:
LGTM!  Merging!

PR #1207, opened 2021-12-14
Document dotnet Linux ARM64 image build and test

saquib-rashid-msft:
# Task: 22998

## Description

---

Document the process of building and testing Linux ARM64 specific dotnet images.

## PR Checklist

---

- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] My code follows the code style of this project.
- [ ] I ran the lint checks which produced no new errors nor warnings for my changes.
- [x] I have checked to ensure there aren't other open Pull Requests for the same update/change.

---

- [ ] Yes
- [x] No

## Testing

---

n/a

---

n/a


build/README.md
```diff
@@ -692,3 +692,49 @@ After everything builds successfully, the packaging process kicks off and perfor
...
+- `--build-arg`: for this example, we will be targeting specific dotnet versions (e.g. 6.0, 5.0, and 3.1) and Linux versions and distros.
+
+```bash
+docker buildx build --build-arg VARIANT=6.0.100-bullseye-slim-arm64v8 --platform linux/arm64 -t dotnet-arm64  --load -f base.Dockerfile .
+```
+
+Once the build is complete, run the image using the below example. Note that the dotnet directory is mounted. The dotnet directory includes test scripts which will be used in the subsequent steps.
+
+```bash
+docker run -v ~/home/usr/local/repo/vscode-dev-containers/containers/dotnet/:/workspace -it dotnet-arm64 bash
```
juzuluag, line 716: A suggestion (to avoid the `warning` message from docker), pass the platform type for instance
```sh
docker run -v ... --platform linux/arm64 -it dotnet-arm64 ...
```

build/README.md
```diff
@@ -692,3 +692,49 @@ After everything builds successfully, the packaging process kicks off and perfor
...
+
+For ARM64 on Linux, the architecture will show as `aarch64`.
+
+### Test
+Once in the container, run the `test.sh` script from the `workspace/test-project` directory:
+
+```bash
+./test.sh
+```
+
```
juzuluag, line 733: Probably worth to mention that if the test runs on a non VSCode attached scenario there would be an expected error, since `vscode-server` folder won't be found

saquib-rashid-msft: ah, yes, I think I had that there in the original doc but this revised one, I accidently removed it.

bderusha:
The technical documentation bits here looks good to me.  My only real piece of feedback is that I found the `dotnet` examples specificity a bit out of place for this doc.  I understand that's the context that this work was being done in, but I think it would be clearer and easier to follow if it was written more generally.

dnastrain (approved):
Overall LGTM, nothing to add beyond what @bderusha already pointed out.

build/README.md
```diff
@@ -692,3 +692,49 @@ After everything builds successfully, the packaging process kicks off and perfor
...
+- `--build-arg`: for this example, we will be targeting specific dotnet versions (e.g. 6.0, 5.0, and 3.1) and Linux versions and distros.
+
+```bash
+docker buildx build --build-arg VARIANT=6.0.100-bullseye-slim-arm64v8 --platform linux/arm64 -t dotnet-arm64  --load -f base.Dockerfile .
+```
+
+Once the build is complete, run the image using the below example. Note that the dotnet directory is mounted. The dotnet directory includes test scripts which will be used in the subsequent steps.
+
+```bash
+docker run -v ~/home/usr/local/repo/vscode-dev-containers/containers/dotnet/:/workspace --platform linux/arm64 -it dotnet-arm64 bash
```
xtophs, line 716: the host path looks incorrect with /home/usr/local inside ~.  Maybe just have a placeholder or an environment variable like `$REPODIR`?

build/README.md
```diff
@@ -692,3 +692,49 @@ After everything builds successfully, the packaging process kicks off and perfor
...
+For ARM64 on Linux, the architecture will show as `aarch64`.
+
+### Test
+Once in the container, run the `test.sh` script from the `workspace/test-project` directory:
+
+```bash
+./test.sh
+```
+
+For Linux ARM64 and the Dotnet container, the tests should all pass except for the `vscode-server` test. This test requires that VS Code attach to the running container instance. Using VS Code and the `Remote-Containers` extension, attach to the running instance of the image previously created, re-run the tests and they should all pass, including `vscode-server`.
```
xtophs, line 734: How about making this a little more repeatable and discoverable? Maybe modify the test script to emit the above message  (and moving `vscode-server` to a single `check` instead of including it in `checkMultiple`. Chuck probably has some ideas how to do that the right way and move it up into the template.

xtophs (changes requested):

build/README.md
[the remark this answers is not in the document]
juzuluag: Fixed

build/README.md
juzuluag: Fixed

juzuluag (commented):
Hi @xtophs , I just addressed your comments.

build/README.md
```diff
@@ -692,3 +692,49 @@ After everything builds successfully, the packaging process kicks off and perfor
      📄 devcontainer.json
      📄 Dockerfile
 ```
+
+## Linux ARM64 Specific Builds
+
+The below provides a mechanism to build and test against targeted platforms (e.g. Linux ARM64). The following example uses the the dotnet container image for reference.
+
+### Build
```
gfmatthews, line 700: Let's assume one has gone through this process and validated a new container image for ARM64. How does one then go about adding this to the official build process?

build/README.md
```diff
@@ -692,3 +692,49 @@ After everything builds successfully, the packaging process kicks off and perfor
...
+
+### Test
+
+Once in the container, run the `test.sh` script from the `workspace/test-project` directory:
+
+```bash
+/workspace/test-project/test.sh
+```
+
+For Linux ARM64 and the Dotnet container, the tests should all pass except for the `vscode-server` test. This test requires that VS Code attaches to the running container instance. Using VS Code and the `Remote-Containers` extension, attach to the running instance of the image previously created, re-run the tests and they should all pass, including `vscode-server`.
```
gfmatthews, line 735: So, the tests don't *all* pass if you run them in the container directly, but if you run them using the extension they all work right? Wouldn't it be clearer (and simpler) to just specify that you should use the remote-containers feature to attach to the image and then run the tests?

build/README.md
juzuluag: Hi @gfmatthews , good question. AFAIK, devcontainer team is the only capable of adding it to the build process. But it would be good to get confirmation from @Chuxel

build/README.md
juzuluag: Agree. I'll remove that case. Thanks.

build/README.md
Chuxel: Note that adding an architecture for an existing image name (e.g. mcr.microsoft.com/vscode/devcontainers/dotnet) reuses the same image name, so only the architecture property in definition-manifest.json is needed.

gfmatthews (approved):

build/README.md
```diff
@@ -692,3 +692,48 @@ After everything builds successfully, the packaging process kicks off and perfor
...
+
+```bash
+ uname -m
+```
+
+For ARM64 on Linux, the architecture will show as `aarch64`.
+
+### Test
+
+For Linux ARM64 and the Dotnet container requires that VS Code attaches to the running container instance. Using VS Code and the `Remote-Containers` extension, attach to the running instance of the image previously created, run the tests and they should all pass.
```
Kpopovs1, line 729: For context, you could add a link to the `Remote-Containers` extension documentation so the readers have an idea of how to attach a running container and run the tests (and add the link to the references below as well).

juzuluag: Great suggestion @Kpopovs1 , will do.

saquib-rashid-msft:
As per request from @esgraham , moving the PR to ready for review.
cc @juzuluag

Chuxel:
LGTM!

PR #1208, opened 2021-12-15
Update rust-debian.sh

Rose2161:


Chuxel:
This appears to be spam.

Rose2161:
No it's not

Aisha Redl-Sherwood, CEO/CFO
503-857-9072
Cryptob3auty I.S.
P.O. Box 1524
McMinnville, Or 97128

Rose2161:
No it's not spam!

Chuxel:
The net effect here would be to update the user in the script which is not something we'd want to do. The script will automatically detect whatever user has a UID/GID of 1000 in the image if one already exists and the other names are not present.

PR #1209, opened 2021-12-15
Fix OMZ folder ownership after installing completions

felipecrs:
My devcontainer was failing to install my dotfiles because of this since many days, and I spent several days to track this change to here. That was mainly because I didn't pin the revision of the script to use, so I was always using the latest.

The problem is that, this script creates the OMZ folder as root since https://github.com/microsoft/vscode-dev-containers/pull/1163, and forgets to change its ownership back to the original non-root user.

Also, this PR addresses some ShellCheck warnings.

### DIFF
--- a/script-library/kubectl-helm-debian.sh
+++ b/script-library/kubectl-helm-debian.sh
@@ -140,12 +140,13 @@ fi
 kubectl completion bash > /etc/bash_completion.d/kubectl
 
 # kubectl zsh completion
-if [ "$USERNAME" != "root"  ]; then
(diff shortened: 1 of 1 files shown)


felipecrs:
BTW, I already tested, and it works as expected.

felipecrs:
Also, for later, I would suggest some enhancements. E.g.:

- Should not do it if ZSH is not installed
- Should not do it if OMZ is not installed

Chuxel:
Ahh - good catch.  Thanks!

PR #1211, opened 2021-12-16
Fix kubectl script when USERNAME is not set

felipecrs:
I noticed this in one of my images: https://github.com/felipecrs/docker-images/runs/4541686579?check_suite_focus=true#step:7:4000


Chuxel:
Yeah looking at the block of code, its missing username detection too.  Will add that.

PR #1215, opened 2021-12-20
Fix OMZ folder ownership after installing completions

felipecrs:
This reapplies https://github.com/microsoft/vscode-dev-containers/pull/1209, which got reverted as part of https://github.com/microsoft/vscode-dev-containers/commit/6985da4ccee756a468cc68b659177576c9aee4c4.


Chuxel:
Thanks!!

PR #1216, opened 2021-12-20
Allow beta versions of Go

stellirin:
Very useful for testing Generics in v1.18beta1

### DIFF
--- a/script-library/docs/go.md
+++ b/script-library/docs/go.md
@@ -24,7 +24,7 @@ Or as a feature:
 
 |Argument|Feature option|Default|Description|
 |--------|--------------|-------|-----------|
-|Go version|`version` | `latest`| Version of Go to install. Use `latest` to install the latest released version. Partial version numbers are accepted (e.g. `1.17`).|
+|Go version|`version` | `latest`| Version of Go to install. Use `latest` to install the latest released version. Use `beta` to install the latest beta version. Partial version numbers are accepted (e.g. `1.17`).|
 |GOROOT| | `/usr/local/go`| Location to install Go. |
 |GOPATH| | `/go`| Location to use as the `GOPATH`. Tools are installed under `${GOPATH}/bin` |
 |Non-root user| | `automatic`| Specifies a user in the container other than root. A value of `automatic` will cause the script to check for a user called `vscode`, then `node`, `codespace`, and finally a user with a UID of `1000` before falling back to `root`. |
--- a/script-library/go-debian.sh
+++ b/script-library/go-debian.sh
@@ -71,14 +71,16 @@ find_version_from_git_tags() {
     if [ "$(echo "${requested_version}" | grep -o "." | wc -l)" != "2" ]; then
         local escaped_separator=${separator//./\\.}
         local last_part
-        if [ "${last_part_optional}" = "true" ]; then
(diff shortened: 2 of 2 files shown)


bamurtaugh:
Thanks for opening @stellirin!

@Chuxel Does this sound reasonable to you? I was able to verify it works (added PR's version of `go-debian.sh` to my `.devcontainer/library-scripts`, signified `beta` in Dockerfile; running `go version` output `go version go1.18beta2 linux/amd64`).

Chuxel:
@bamurtaugh Yeah I think so.  This updates a utility script we tend to reuse in a number of different spots, so there's a slight risk of it getting undone by accident. We could instead add a "suffix" param to that particular function and use it from the go script. Then the go script would do what is here if "beta" is the value.

Thoughts @joshspicer? I know with features you're thinking about moving more towards utility functions.

e.g. this would add the suffix
```bash
# Figure out correct version of a three part version number is not passed
find_version_from_git_tags() {
    local variable_name=$1
    local requested_version=${!variable_name}
    if [ "${requested_version}" = "none" ]; then return; fi
    local repository=$2
    # Normally a "v" is used before the version number, but support alternate cases
    local prefix=${3:-"tags/v"}
    # Some repositories use "_" instead of "." for version number part separation, support that
    local separator=${4:-"."}
    # Some tools release versions that omit the last digit (e.g. go)
    local last_part_optional=${5:-"false"}
    # Some repositories may have tags that include a suffix (e.g. actions/node-versions)
    local version_suffix_regex=$6

    local escaped_separator=${separator//./\\.}
    local break_fix_digit_regex
    if [ "${last_part_optional}" = "true" ]; then
        break_fix_digit_regex="(${escaped_separator}[0-9]+)?"
    else
        break_fix_digit_regex="${escaped_separator}[0-9]+"
    fi    
    local version_regex="[0-9]+${escaped_separator}[0-9]+${break_fix_digit_regex}${version_suffix_regex//./\\.}"
    # If we're passed a matching version number, just return it, otherwise look for a version
    if ! echo "${requested_version}" | grep -E "^${versionMatchRegex}$" > /dev/null 2>&1; then
        local version_list="$(git ls-remote --tags ${repository} | grep -oP "${prefix}\\K${version_regex}$" | tr -d ' ' | tr "${separator}" "." | sort -rV)"
        if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ] || [ "${requested_version}" = "lts" ]; then
            declare -g ${variable_name}="$(echo "${version_list}" | head -n 1)"
        else
            set +e
            declare -g ${variable_name}="$(echo "${version_list}" | grep -E -m 1 "^${requested_version//./\\.}([\\.\\s]|${version_suffix_regex//./\\.}|$)")"
            set -e
        fi
        if [ -z "${!variable_name}" ] || ! echo "${version_list}" | grep "^${!variable_name//./\\.}$" > /dev/null 2>&1; then
            echo -e "Invalid ${variable_name} value: ${requested_version}\nValid values:\n${version_list}" >&2
            exit 1
        fi
    fi
    echo "Adjusted ${variable_name}=${!variable_name}"
}
```

joshspicer:
Yea, I think making it more generic, to be used in other places, will be very useful in the future!

PR #1219, opened 2021-12-23
Dev container for terraform development

Shashindran:
A simple development environment for terraform. Terraform commands can be executed from within VS Code terminal


bamurtaugh:
Thanks for opening this @Shashindran. We do have a Terraform definition: https://github.com/microsoft/vscode-dev-containers/tree/main/containers/azure-terraform; would that meet your needs for Terraform development?

PR #1229, opened 2022-01-04
Fix "Comment out connect as root instead" typo

artnc:
I was pretty confused by this line until I realized it was a typo.


Chuxel:
Ha - thanks! Amazing the number of times we've looked at that and not seen it.

PR #1237, opened 2022-01-05
Do not recommend Debian 9/stretch

Chuxel:
Debian 9 / Stretch is no longer getting security patches as it is EOL and it's community LTS support runs out in June of this year.  https://wiki.debian.org/LTS

We should not be recommending it to people at this point. We can keep pre-building the stretch image for existing use until June as long as there isn't a critical security issue that is "won't fix".


Chuxel:
//cc @2percentsilk as FYI

PR #1261, opened 2022-01-21
Codespaces: Change reference from .env to .env-secrets

samruddhikhandale:
Closes [#4095](https://github.com/github/codespaces/issues/4095)

In order to support multi-line secrets, we now encode secrets and write to  `/workspaces/.codespaces/shared/.env-secrets` file. We encode to this file, and decode while exporting env variables.

The codespaces image is however referencing to old `.env` file which prompts errors while connecting via ssh. Hence, changing the reference.

### DIFF
--- a/containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
+++ b/containers/codespaces-linux/.devcontainer/library-scripts/sshd-debian.sh
@@ -104,10 +104,14 @@ if [ "${CODESPACES}" != "true" ] || [ "${VSCDC_FIXED_SECRETS}" = "true" ] || [ !
     # Not codespaces, already run, or secrets already in environment, so return
     return
 fi
-if [ -f /workspaces/.codespaces/shared/.env ]; then
-    set -o allexport
-    . /workspaces/.codespaces/shared/.env
-    set +o allexport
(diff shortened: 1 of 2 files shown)


joshspicer:
FYI @chuxel @chrmarti  we don't want to release this version of the `universal` image until we release some Codespaces changes.  Can do the universal image release myself once we're ready (early next week)

cc / @samruddhikhandale @jkeech

Chuxel:
@joshspicer Good to know.  We should probably bump up the breakfix versions of the other images next week for endgame to pick up latest patches so maybe we can do that at the same time.

joshspicer:
The agent shipped yesterday with the relevant change.  Working with @samruddhikhandale for final validation

chrmarti:
@joshspicer Only catching up now, I have published 0.215.0 of vscode-dev-containers earlier today. Hope that wasn't too early.

joshspicer:
all good - I didn't bump the manifest ID

samruddhikhandale:
Validated in dev with the latest tag. Looks good 🟢 

image

joshspicer:
https://github.com/microsoft/vscode-dev-containers/commit/b4e90acace4f0251c75673b04379115a67007dc4

PR #1299, opened 2022-02-09
bump oryx in codespaces-linux to 20220212.1

joshspicer:
should pick up https://github.com/microsoft/Oryx/issues/1207


joshspicer:
smoke test still has `node --version` as 14

PR #1305, opened 2022-02-11
[Static Web Apps] Fix Node.js and Core Tools version mismatch

### LINKED ISSUE #1306: [Static Web Apps] Node.js and Core Tools version mismatch
- VSCode Version: N/A
- Local OS Version: 11.6.1
- Local chip architecture: x86
- Reproduces in: Both
- Name of Dev Container Definition with Issue: Azure Static Web Apps

Steps to Reproduce:

1. Use the default Azure Static Web Apps definition
2. Try to run a Functions API server with `func start` => Error

The default image image uses Node.js v16 with Core Tools v3, whereas Node 16 needs Core Tools v4 to work.

sinedied:
This PR brings a few changes:

- It updates the base image to `mcr.microsoft.com/azure-functions/python:4-python3.9-core-tools`, updating Python/Core Tools versions and allowing support for M1 macs
- It changes the default Node.js version to 16 instead of LTS to avoid unwanted upgrades and mismatch with core tools version
- It adds a `CORE_TOOLS_VERSION` param to allow switching to a different core tools version that the default one, for example if you wish to use a Node 14 version
- It fixes the mismatch between the previous default Node version (16) and Core Tools (v3) installed in the previous image

Fixes #1306


Chuxel:
LGTM! @anthonychu @TsuyoshiUshio Does this look good to you?

anthonychu:
Looks good. Keep in mind that Node.js 16 isn’t available in Static Web Apps functions yet. But hopefully by the time this change is released to stable it will!

Chuxel:
Sounds good. Merging!

PR #1307, opened 2022-02-14
include `devcontainer-collection.json`

joshspicer:
related to https://github.com/github/codespaces/issues/6384 and https://github.com/microsoft/dev-container-spec/issues/7


joshspicer:
Keeping `publish-definitions` disabled in the action for now

PR #1308, opened 2022-02-14
[codespaces] Add JupyterLab feature

bratsche:
https://github.com/github/codespaces/issues/6507


Chuxel:
For this to be in the script library, it will need more setup content to ensure the base dependencies are installed. e.g. python is not guaranteed to be in many images.  If this is just going to be in the codespaces-linux image, then it makes sense to leave the script there.

bratsche:
Closing this one for now, in favor of https://github.com/microsoft/vscode-dev-containers/pull/1323 which just adds JupyterLab to codespaces-linux.

PR #1317, opened 2022-02-22
Default to non-ARM64 stable Ruby 3.1 release

olivierlacan:
There’s only [one minor branch left with official support under Ruby 2.x ](https://www.ruby-lang.org/en/downloads/)(2.7) and it will likely be slated for end of life fairly soon, so it feels odd to default to a version that is 2 release behind latest stable. 

Also unless I’m mistaken Codespaces don’t run on ARM64/Apple Silicon so I’m not sure why bullseye is the default as well.


Chuxel:
@olivierlacan These images are also available locally via Remote - Containers or just general use, which is why buillseye is the default, and in fact the normal numbered versions also are bullseye in alignment with the `ruby` base image.

That said, to switch to Ruby 3 as the default, you'll need to update the docker-compose.yaml file as for this particular definition since that overrides what is in the Dockerfile.

olivierlacan:
@Chuxel Oh ok, I honestly have no idea what Remote - Containers are since I’m using all of this on an iPad at the moment exclusively. Might be a good idea to clarify what “local” means because it’s fairly ambiguous. 

So there’s no … harm in using an ARM-friendly default image essentially?

olivierlacan:
By the way I updated the commit to retain bullseye but use Ruby 3.1 in both Dockerfile and docker-compose.

Chuxel:
@olivierlacan Yes, agreed that the GitHub Codespaces docs do not do a good job of explaining that you can use these same devcontainer.json files locally or via a remote SSH host. That is actually where all these definitions originate. 

ARM64 will only be used if you are running an ARM64 device. This is something Docker images support natively.

LGTM!

PR #1318, opened 2022-02-23
Bump break fix versions for release 1.65

Chuxel:
//cc @joshspicer @chrmarti 

To merge whenever we feel like we're ready - prepped it for VS Code release next week if nothing else.


chrmarti:
This is to trigger a full rebuild of the base images, correct?

Chuxel:
Yep. Merging this will re-trigger a full build of the base images to pick up latest security patches on release.

C++ already is ready to go given there was a PR merged there that affects the Dockerfile and should be matched up.  The others just need a version bump to keep currentl.

Chuxel:
Merging for the next release.

PR #1323, opened 2022-02-24
Add JupyterLab to codespaces-linux

bratsche:
This adds JupyterLab into the codespaces-linux image.

Fixes https://github.com/github/codespaces/issues/6507


containers/codespaces-linux/.devcontainer/library-scripts/jupyterlab-debian.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
joshspicer, line 1: See the readme in this folder - anything here that isn't in `/library-scripts` will be deleted with newer copies from that directory. 

If we're not taking this the whole way and authoring a feature, I guess we might as well just add that statement as a line in the dockerfile itself?

containers/codespaces-linux/.devcontainer/base.Dockerfile
```diff
@@ -70,13 +70,14 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
...
 
-# Install Python, PHP, Ruby utilities
+# Install Python, PHP, Ruby utilities, and JupyterLab
 RUN bash /tmp/scripts/python-debian.sh "none" "/opt/python/latest" "${PIPX_HOME}" "${USERNAME}" "true" \
     # Install rvm, rbenv, any missing base gems
     && chown -R ${USERNAME} /opt/ruby/* \
     && bash /tmp/scripts/ruby-debian.sh "none" "${USERNAME}" "true" "true" \
     # Link composer
     && ln -s $(which composer.phar) /usr/local/bin/composer \
+    && bash /tmp/scripts/jupyterlab-debian.sh \
```
joshspicer, line 80: Would be great to add a line the test script under `/containers/codespaces-linux` to validate that we installed jupityer and it's available.  

That test script is run as the "smoke test" CI on the is PR. The PR is expected to fail right now due to some unresolved oryx issues, but you'll at least be able to see the result of any lines you add

bratsche: @joshspicer Thanks! I removed the jupyter-debian.sh and just do the `pip install ...` from the base.Dockerfile now.

Issue #1232, opened 2022-01-04
dotnet ARM64 support


juzuluag:
Current dotnet containers are only supporting x86. Based on some work done, `dotnet` images can be build to support arm64, but still there some dependency on Omnisharp Extension in order to have a complete user experience.
This issue was opened to track the OmniSharp dependency.

------
Testing with [Omnisharp 1.24.0-beta1](https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.24.0-beta1) with various images, it was possible to debug inside the dotnet arm64 container.

```sh
Starting OmniSharp server at 1/4/2022, 9:17:03 PM
    Target: /root

OmniSharp server started with .NET 6.0.101
.
    Path: /root/.vscode-server/extensions/ms-dotnettools.csharp-1.24.0/.omnisharp/1.38.0-net6.0/OmniSharp.dll
    PID: 5473

[info]: OmniSharp.Stdio.Host
        Starting OmniSharp on Unknown 0.0 (Unknown)
[info]: OmniSharp.Services.DotNetCliService
        Checking the 'DOTNET_ROOT' environment variable to find a .NET SDK
....
```
**Images**:
- 6.0-bullseye-slim-arm64v8 
- 5.0-bullseye-slim-arm64v8
- 3.1-bullseye-arm64v8

In case of dotnet 6.0, changes need to be added to the container's vscode settings to use 6.0 features.
```text
.vscode-server/data/Machine/settings.json
```

```json
{
	"omnisharp.useModernNet": true
}
```

Relates to: <Codespaces | Remote - Containers | Both>

juzuluag:
Reference: https://github.com/microsoft/vscode-dev-containers/issues/558

tksrc:
Since OmniSharp 1.24 is now released - do we expect the dotnet vscode images to be released for ARM as well?

Chuxel:
@juzuluag I know you hit some problems with 1.24 and arm64 Linux here (which is what the container would run), correct?  Or has that been resolved since the beta?

juzuluag:
> @juzuluag I know you hit some problems with 1.24 and arm64 Linux here (which is what the container would run), correct? Or has that been resolved since the beta?

Hi @Chuxel. I faced issues using 1.23.18-beta2 and not with 1.24.0-beta1. I haven't try the latest ([1.24.0](https://github.com/OmniSharp/omnisharp-vscode/releases/tag/v1.24.0))

PR #1325, opened 2022-02-24
Add dotnet arm64 support

JenMadiedo:
#1232  dotnet arm64 support

Description
---
This PR includes:

 - Updated `definition-manifest.json` in the dotnet container to pull both the amd-64 and arm-64 images for all variants
 - Updated dotnet, dotnet-postgres, and dotnet-fsharp `README.md` to reflect new support for arm64

PR Checklist
---
- [ ] I have added tests to cover my changes.
- [ ] I ran the lint checks which produced no new errors nor warnings for my changes.

Does this introduce a breaking change?
---
- [ ] Yes
- [X] No

Testing
---
- Manual testing that arm64 dotnet image could be used for a dotnet console app
- Manual testing that arm64 dotnet image could be used in conjunction with postgres and with f# to successfully run apps. 

Other information or known dependencies
---

### DIFF
--- a/containers/dotnet-fsharp/README.md
+++ b/containers/dotnet-fsharp/README.md
@@ -9,7 +9,7 @@
 | *Contributors* | The VS Code Team, F# team |
 | *Categories* | Languages |
 | *Definition type* | Dockerfile |
-| *Published image architecture(s)* | x86-64 |
+| *Published image architecture(s)* | x86-64, arm64/aarch64 |
 | *Works in Codespaces* | Yes |
 | *Container host OS support* | Linux, macOS, Windows |
 | *Container OS* | Ubuntu |
--- a/containers/dotnet-postgres/README.md
+++ b/containers/dotnet-postgres/README.md
@@ -9,7 +9,7 @@
 | *Contributors* | [Berkays](https://github.com/Berkays)  |
 | *Categories* | Core, Languages |
 | *Definition type* | Docker Compose |
-| *Published image architecture(s)* | x86-64 |
+| *Published image architecture(s)* | x86-64, arm64/aarch64 |
 | *Available image variants* | [See C# (.NET) definition](../dotnet). |
 | *Works in Codespaces* | Yes |
 | *Container host OS support* | Linux, macOS, Windows |
--- a/containers/dotnet/README.md
+++ b/containers/dotnet/README.md
@@ -11,7 +11,7 @@
 | *Definition type* | Dockerfile |
 | *Published images* | mcr.microsoft.com/vscode/devcontainers/dotnet |
 | *Available image variants* | 3.1 / 3.1-focal, 5.0 / 5.0-focal, 6.0 /6.0-bullseye, 6.0-focal, 5.0-bullseye, 3.1-bullseye ([full list](https://mcr.microsoft.com/v2/vscode/devcontainers/dotnet/tags/list)) |
-| *Published image architecture(s)* | x86-64 |
+| *Published image architecture(s)* | x86-64, arm64/aarch64 |
 | *Works in Codespaces* | Yes |
 | *Container host OS support* | Linux, macOS, Windows |
 | *Container OS* | Ubuntu (`-focal`), Debian (`-bullseye`) |
--- a/containers/dotnet/definition-manifest.json
+++ b/containers/dotnet/definition-manifest.json
@@ -11,7 +11,7 @@
 		"variantTags": {
 			"6.0-focal": [
 				"dotnet:${VERSION}-6.0",
-				"dotnet:${VERSION}-6.0"
+				"dotnetcore:${VERSION}-6.0"
 			],
 			"6.0-bullseye-slim": [
 				"dotnet:${VERSION}-6.0",
--- a/containers/dotnet/definition-manifest.json
+++ b/containers/dotnet/definition-manifest.json
@@ -7,15 +7,12 @@
 		"tags": [
 			"dotnet:${VERSION}-${VARIANT}"
 		],
-		"architectures": {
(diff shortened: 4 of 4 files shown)


containers/dotnet/definition-manifest.json
```diff
@@ -7,7 +7,12 @@
 		"tags": [
 			"dotnet:${VERSION}-${VARIANT}"
 		],
+		"architectures": ["linux/amd64", "linux/arm64"],
 		"variantTags": {
+			"6.0-focal": [
+				"dotnet:${VERSION}-6.0",
+				"dotnetcore:${VERSION}-6.0"
```
dnastrain, line 14: am not 100% sure, but I don't think this `dotnetcore:<version>-6.0` container + tag is published, based on the [mcr list here](https://mcr.microsoft.com/v2/vscode/devcontainers/dotnetcore/tags/list) for 'vscode/devcontainers/dotnetcore'

JenMadiedo: Good catch, thought I saw it but it seems to only go to 5

Chuxel: Yeah - we kept it for older versions of C# for compatibility after the change. New ones like 6.0 we dropped it since it's not .NET Core.

containers/dotnet/definition-manifest.json
```diff
@@ -7,7 +7,11 @@
 		"tags": [
 			"dotnet:${VERSION}-${VARIANT}"
 		],
+		"architectures": ["linux/amd64", "linux/arm64"],
 		"variantTags": {
+			"6.0-focal": [
```
Chuxel, line 12: This is not needed. It's covered by the variants and the primary `tag` property.  We don't add dotnetcore to new versions of .NET given that brand is gone. The others are there for back compat post change.

The .NET image defaults to debian bullseye, which is why this is also defaulting to bullseye for pure "6.0".

containers/dotnet-fsharp/README.md
```diff
@@ -9,7 +9,7 @@
 | *Contributors* | The VS Code Team, F# team |
 | *Categories* | Languages |
 | *Definition type* | Dockerfile |
-| *Published image architecture(s)* | x86-64 |
+| *Published image architecture(s)* | x86-64, arm64/aarch64 |
```
Chuxel, line 12: We'll need to qualify which images have arm64/aarch64 support since the focal ones will not.  See the python-3 definition for how this is described.

containers/dotnet-postgres/README.md
```diff
@@ -9,7 +9,7 @@
 | *Contributors* | [Berkays](https://github.com/Berkays)  |
 | *Categories* | Core, Languages |
 | *Definition type* | Docker Compose |
-| *Published image architecture(s)* | x86-64 |
+| *Published image architecture(s)* | x86-64, arm64/aarch64 |
```
Chuxel, line 12: See previous comment about this statement. It will need to be qualified.

containers/dotnet/README.md
```diff
@@ -11,7 +11,7 @@
 | *Definition type* | Dockerfile |
 | *Published images* | mcr.microsoft.com/vscode/devcontainers/dotnet |
 | *Available image variants* | 3.1 / 3.1-focal, 5.0 / 5.0-focal, 6.0 /6.0-bullseye, 6.0-focal, 5.0-bullseye, 3.1-bullseye ([full list](https://mcr.microsoft.com/v2/vscode/devcontainers/dotnet/tags/list)) |
-| *Published image architecture(s)* | x86-64 |
+| *Published image architecture(s)* | x86-64, arm64/aarch64 |
```
Chuxel, line 14: Ditto here.

containers/dotnet/definition-manifest.json
```diff
@@ -7,7 +7,11 @@
 		"tags": [
 			"dotnet:${VERSION}-${VARIANT}"
 		],
+		"architectures": ["linux/amd64", "linux/arm64"],
```
Chuxel, line 10: I just tried and building an arm64 image for the focal variants fails on macOS and will also in our CI.  You'll see an error about GPG key missing and this is due to a curl segfault that has no resolution.

See python-3 for how this was set up to change architecture by variant.

Chuxel (commented):
A few issues I'm seeing - e.g. focal bombs on macOS arm64 Linux and will also in our CI due to the problem described here. No other images include arm64 for focal (or debian buster, but that's N/A here). 

https://github.com/microsoft/vscode-dev-containers/issues/558#issuecomment-905117722

It's even listed as an issue in Docker Desktop instructions.

JenMadiedo:
> A few issues I'm seeing - e.g. focal bombs on macOS arm64 Linux and will also in our CI due to the problem described here. No other images include arm64 for focal (or debian buster, but that's N/A here).
> 
> [#558 (comment)](https://github.com/microsoft/vscode-dev-containers/issues/558#issuecomment-905117722)
> 
> It's even listed as an issue in Docker Desktop instructions.

@Chuxel got it, we weren't sure since we saw that there are focal arm images published for dotnet, but that all makes sense, I've updated accordingly to qualify and remove the architectures for arm

Chuxel:
@JenMadiedo Yeah that's the strange bit, if you run actual Linux on arm64, it works fine (e.g. an arm server in the cloud or an actual arm device). It's specifically M1/Apple Silicon macs using Docker Desktop, or if you're emulating linux/arm64 on linux x86 - which is the typical CI case. The problem is that these particular images target developers, and all new macs are now all M1... so it's a frustrating set of circumstances that the most likely combination would fail. That's why we moved to bullseye as fast as we could where we provided arm support.

Anyway, LGTM!

Issue #1122, opened 2021-10-21
Missing some SDK/CLI "features" for remote containers


Rabadash8820:
There are a couple glaring CLIs/SDKs missing from the remote container "features" menu in VS Code:

- AWS CLI
- AWS SAM CLI
- AWS CDK CLI
- Google Cloud CLI (I don't actually use this one, but I guarantee others will want it)
- Dotnet CLI
- MSBuild CLI
- NuGet CLI

I was hoping to make PRs for these, but the shell scripts for the Azure CLI and others were kind of daunting. If there are any approachable docs or templates, I'd still be open to trying to implement some of these myself.

Relates to: Remote - Containers

takekazuomi:
I also want to add bicep. Waiting for documents and templates.

esgraham:
@bderusha and @Kpopovs1 working on issue

esgraham:
Dotnet CLI has been added to the "Feature" area of dev containers.  Documentation has been created that details how to add other SDKs or CLIs.

Pull Requests associated to Issue:  
(Dotnet CLI added to Feature) https://github.com/microsoft/vscode-dev-containers/pull/1210
(Documentation Added):  https://github.com/microsoft/vscode-dev-containers/pull/1246

Issue #1291, opened 2022-02-04
[Announcement] State of vscode-dev-containers contributions


bamurtaugh:
Hi folks! We wanted to give you the latest update on this repository.

We're currently thinking about how to best scale the growth of dev container definitions, as there's increasing interest in new definitions and configurations. This includes how to best manage existing and future community contributions. We'll be having open discussions on this in [an issue](https://github.com/microsoft/dev-container-spec/issues/7) in the new [dev container spec repo](https://github.com/microsoft/dev-container-spec/issues).

We are also evolving the idea of a dev container feature, with the goal of being better able to mix, match, and share config that can be acquired remotely. You can view and join the discussion in [this issue](https://github.com/microsoft/dev-container-spec/issues/8), and read more about features in [this doc](https://code.visualstudio.com/docs/remote/containers#_dev-container-features-preview).

Please know that we tremendously value your contributions to and usage of dev containers and their definitions! In the near-term, we'll look to accept changes in this repo with a high-priority user impact/need. For the long term, we're still discussing the best process to accept and manage contributions. Since there may be fewer responses in this repository as we figure out this plan, we wanted to make you all aware and assure you that we see and appreciate your work.

Thank you for your patience and engagement!

bamurtaugh:
Hi folks,

Thank you again for your contributions and patience. We wanted to share the latest plans for dev container contributions.

The configuration files you see in this repo are moving over to a repo in our new `devcontainers` org, which will likely be https://github.com/devcontainers/templates. In this repo, you'll find a markdown file that lists community-contributed templates. If you'd like to add a new template, you can open a PR to this markdown file and add your repo containing a dev container configuration to the list. As there is a growing number of potential community contributions, this model provides more community autonomy and scalability.

Similar to [vscode-dev-containers](https://github.com/microsoft/vscode-dev-containers), you may also open issues and PRs to contribute to the templates that are directly managed within this new repo.

We look forward to continue making dev containers a top-notch experience, and we value your continued contributions and support. Thank you!

samruddhikhandale:
Closing this in favor of a latest update mentioned in https://github.com/microsoft/vscode-dev-containers/issues/1589

PR #1326, opened 2022-02-26
AWS CLI feature

Rabadash8820:
Adds a container feature for the missing AWS CLI (as mentioned in #1122). I think I followed the contribution guidelines correctly, but please let me know if there are issues with this PR. In particular:

- Am I targeting the correct branch (`main`)?
- I wasn't sure if this was a use case for just a container _feature_ or a whole new _container_. I've started with the former.
- There are some other non-essential changes in the first commit. If these are out of scope then I can drop/revert that commit.
- I have a hard-coded GPG public key and fingerprint in the install script; wasn't sure if this should be added to the `shared/settings.env` file in some way.
- I put `VS Code and GitHub Codespaces teams` in the `Maintainers` sections, in the hopes that that team _would_ take over this feature, since it seems like a pretty essential one. But if not, I guess I'm willing to put my username instead.
- I'm not sure if I included all the appropriate containers in the `include` section of `devcontainer-features.json`. Seems to me like the association should really be the _other_ way around, with _containers_ specifying which _features_ they include, but still.
- I hope to add features and/or containers for the SAM CLI and CDK CLIs also, but wanted to make sure that I'm doing things correctly here before opening more PRs.


script-library/awscli-debian.sh
```diff
@@ -0,0 +1,99 @@
...
+
+install() {
+    local scriptZipFile=awscli.zip
+    local scriptSigFile=awscli.sig
+
+    # See Linux install docs at https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
+    if [ "${AWSCLI_VERSION}" != "latest" ]; then
+        local versionStr=-${AWSCLI_VERSION}
+    fi
+    local scriptUrl=https://awscli.amazonaws.com/awscli-exe-linux-x86_64${versionStr}.zip
```
sktan, line 79: @Rabadash8820 it might be useful to determine the architecture running the devcontainer so that it will work for both x64 and ARM architectures:

```
# Install x86 or ARM version of awscliv2 based on current machine architecture
if [ "$(uname -m)" == "x86_64" ]; then
  local scriptUrl=https://awscli.amazonaws.com/awscli-exe-linux-x86_64${versionStr}.zip
else
  local scriptUrl=https://awscli.amazonaws.com/awscli-exe-linux-aarch64${versionStr}.zip
fi
```

Rabadash8820: Great suggestion! Just updated. Also included an error message for machines that are not x84-64 or ARM64.

Chuxel:
This is great. Yeah walking through your questions:

> Am I targeting the correct branch (main)?

Yep!

> I wasn't sure if this was a use case for just a container feature or a whole new container. I've started with the former.

Yeah given how this would be consumed, I think a feature would make sense.

> I have a hard-coded GPG public key and fingerprint in the install script; wasn't sure if this should be added to the shared/settings.env file in some way.

There is a utility function that will go and check that file (assuming it can access it) to see if there is a more up-to-date file and get the specified variable.  This helps insulate situations where a GPG key, signature, etc are invalidated and need to be updated. See [here](https://github.com/microsoft/vscode-dev-containers/blob/1971f2a7808178886076e65043c573d4d8f3e7c9/script-library/shared/utils.sh#L100).

> I'm not sure if I included all the appropriate containers in the include section of devcontainer-features.json. Seems to me like the association should really be the other way around, with containers specifying which features they include, but still.

Yeah there is both an "include" and "exclude" option. The challenge with "exclude" is that the testing combinations get out of hand quickly.

> I put VS Code and GitHub Codespaces teams in the Maintainers sections, in the hopes that that team would take over this feature, since it seems like a pretty essential one. But if not, I guess I'm willing to put my username instead.

Right now, we're looking at ways to better enable self-publishing and we have found it is best to have the contributor involved in ongoing work even if we're still keeping an eye on things. This allows individuals and organizations to publish updates on a schedule that makes sense for the specific scenario - folks like yourself will work day to day in particular ecosystem and will best know when things shift. (This is what @bamurtaugh summarized in https://github.com/microsoft/vscode-dev-containers/issues/1291.)

Rabadash8820:
@Chuxel Thanks for the responses, and for the link to that other repo; very interesting to see how this space is evolving!

So to clarify:
1. I did see that `receive_gpg_keys` util method, but I was unclear how it works. I'm not _super_ experienced with GPG...it looks like that method is downloading keys from a public key server, but idk which server(s), if any, hold the Amazon key.
2. Is the `include` metadata good as is? When, if ever, would it need to change?
3. Should I put myself as the Maintainer or leave it as is?

Chuxel:
@Rabadash8820 We chatted about this internally and agree that having us help maintain this particular one does make sense given its criticality.  That said, if it's okay with you, we'd add you on as an additional contributor/maintainer given the primary folks maintaining this repository do not live in AWS day to day, so we may not be proactively aware of shifts that happen that would affect the feature.

The include metadata looks just fine!

> I did see that `receive_gpg_keys` util function, but I was unclear how it works. I'm not super experienced with GPG...it looks like that function is downloading keys from a public key server, but idk which server(s), if any, hold the AWS CLI key.

Yeah, everyone manages GPG keys a bit differently from what we've seen. Sometimes it's a key server, sometimes it's a file, other times it's the data as you have here.  For AWS, it looks like they have the hard coded key on a cursory glance.

What you can do to future proof a bit is add that key into [this file](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/shared/settings.env). The GPG_KEY_SERVERS variable is an example of one with newlines in it. In the script, you then use utility function from [here](https://github.com/microsoft/vscode-dev-containers/blob/1971f2a7808178886076e65043c573d4d8f3e7c9/script-library/shared/utils.sh#L100) to fetch it. If it needs to change down the road, we just update the .env file and anyone using the script gets the updated key. Here's how using that function could look.

```bash
AWS_CLI_KEY="-----BEGIN PGP PUBLIC KEY BLOCK-----
mQINBF2Cr7UBEADJZHcgusOJl7ENSyumXh85z0TRV0xJorM2B/JL0kHOyigQluUG
ZMLhENaG0bYatdrKP+3H91lvK050pXwnO/R7fB/FSTouki4ciIx5OuLlnJZIxSzx
PqGl0mkxImLNbGWoi6Lto0LYxqHN2iQtzlwTVmq9733zd3XfcXrZ3+LblHAgEt5G
TfNxEKJ8soPLyWmwDH6HWCnjZ/aIQRBTIQ05uVeEoYxSh6wOai7ss/KveoSNBbYz
gbdzoqI2Y8cgH2nbfgp3DSasaLZEdCSsIsK1u05CinE7k2qZ7KgKAUIcT/cR/grk
C6VwsnDU0OUCideXcQ8WeHutqvgZH1JgKDbznoIzeQHJD238GEu+eKhRHcz8/jeG
94zkcgJOz3KbZGYMiTh277Fvj9zzvZsbMBCedV1BTg3TqgvdX4bdkhf5cH+7NtWO
lrFj6UwAsGukBTAOxC0l/dnSmZhJ7Z1KmEWilro/gOrjtOxqRQutlIqG22TaqoPG
fYVN+en3Zwbt97kcgZDwqbuykNt64oZWc4XKCa3mprEGC3IbJTBFqglXmZ7l9ywG
EEUJYOlb2XrSuPWml39beWdKM8kzr1OjnlOm6+lpTRCBfo0wa9F8YZRhHPAkwKkX
XDeOGpWRj4ohOx0d2GWkyV5xyN14p2tQOCdOODmz80yUTgRpPVQUtOEhXQARAQAB
tCFBV1MgQ0xJIFRlYW0gPGF3cy1jbGlAYW1hem9uLmNvbT6JAlQEEwEIAD4WIQT7
Xbd/1cEYuAURraimMQrMRnJHXAUCXYKvtQIbAwUJB4TOAAULCQgHAgYVCgkICwIE
FgIDAQIeAQIXgAAKCRCmMQrMRnJHXJIXEAChLUIkg80uPUkGjE3jejvQSA1aWuAM
yzy6fdpdlRUz6M6nmsUhOExjVIvibEJpzK5mhuSZ4lb0vJ2ZUPgCv4zs2nBd7BGJ
MxKiWgBReGvTdqZ0SzyYH4PYCJSE732x/Fw9hfnh1dMTXNcrQXzwOmmFNNegG0Ox
au+VnpcR5Kz3smiTrIwZbRudo1ijhCYPQ7t5CMp9kjC6bObvy1hSIg2xNbMAN/Do
ikebAl36uA6Y/Uczjj3GxZW4ZWeFirMidKbtqvUz2y0UFszobjiBSqZZHCreC34B
hw9bFNpuWC/0SrXgohdsc6vK50pDGdV5kM2qo9tMQ/izsAwTh/d/GzZv8H4lV9eO
tEis+EpR497PaxKKh9tJf0N6Q1YLRHof5xePZtOIlS3gfvsH5hXA3HJ9yIxb8T0H
QYmVr3aIUes20i6meI3fuV36VFupwfrTKaL7VXnsrK2fq5cRvyJLNzXucg0WAjPF
RrAGLzY7nP1xeg1a0aeP+pdsqjqlPJom8OCWc1+6DWbg0jsC74WoesAqgBItODMB
rsal1y/q+bPzpsnWjzHV8+1/EtZmSc8ZUGSJOPkfC7hObnfkl18h+1QtKTjZme4d
H17gsBJr+opwJw/Zio2LMjQBOqlm3K1A4zFTh7wBC7He6KPQea1p2XAMgtvATtNe
YLZATHZKTJyiqA==
=vYOk
-----END PGP PUBLIC KEY BLOCK-----
"

get_common_setting AWS_CLI_KEY
echo "${AWS_CLI_KEY}" > "${awsCliPublicKeyFile}"
gpg --quiet --import "${awsCliPublicKeyFile}"
```

Rabadash8820:
@Chuxel That all sounds good! I also asked about when, if ever, the `include`/`exclude` metadata will need to be updated. Do you have recommendations on that?

Regarding the GPG key, I'm not really sure I see the value in adding this key to the shared settings file. TBH, most of the settings in that file aren't really "shared", and look like they are only relevant to one or two specific feature install scripts. So in theory, if no other feature is going to need the AWS CLI GPG key, why should it be a shared setting?

Rabadash8820:
This is a bit out of scope for this PR, but I'm wondering if there should be separate utility functions for the different GPG scenarios that you described:

1. `receive_gpg_keys`: looks up a key from various keyservers, as the current function does (maybe with additional ability to provide a feature-specific keyserver)
2. `import_gpg_key`: has options to provide key as a file path, or as hard-coded text (presumably from a variable)

joshspicer:
> Regarding the GPG key, I'm not really sure I see the value in adding this key to the shared settings file. TBH, most of the settings in that file aren't really "shared", and look like they are only relevant to one or two specific feature install scripts. 

The shared file is mostly there to keep us sane maintaining all these definitions/features (it's nice to have it in one spot).  Also, if these settings are centralized, we may be able to (down the line) do something more advanced with the tooling to push out fixes for "commonly breakable" things. (At least that's how I see it :) )

If nothing else, it makes the script file itself look a bit cleaner

joshspicer:
> * `receive_gpg_keys`: looks up one or more keys from various keyservers, as the current function does (maybe with additional ability to provide a feature-specific keyserver)
> * `import_gpg_key`: has options to import a single key from a file path, or from hard-coded text (presumably from a variable)

I think the only issue with this idea is that every vendor does seems to use a slightly different means of validating their binary/package/etc..

Rabadash8820:
@joshspicer Works for me. 🤷‍♂️ Just updated to use `get_common_setting` for the GPG key/fingerprint. This was the first instance of an entire GPG key in the shared settings file, so I went with the naming convention `*_GPG_KEY_MATERIAL`. Also I wasn't quite sure how to test this; before I could just copy the feature install script to the `.devcontainer` folder of a new workspace, but now the dependence on the shared settings/utils files is throwing me off a bit. The CI tests should catch things though, right?

> I think the only issue with this idea is that every vendor does seems to use a slightly different means of validating their binary/package/etc..

I mean, I think "receiving" and "importing" cover the two primary GPG use cases, but I agree that if these shared functions are only used by like one or two features...might not be worth it.

script-library/awscli-debian.sh
```diff
@@ -0,0 +1,76 @@
...
+    local sigFilePath=$2
+
+    get_common_setting AWSCLI_GPG_KEY
+    get_common_setting AWSCLI_GPG_KEY_MATERIAL
+    local awsCliPublicKeyFile=aws-cli-public-key.pem
+    echo "${AWSCLI_GPG_KEY_MATERIAL}" > "${awsCliPublicKeyFile}"
+    gpg --quiet --import "${awsCliPublicKeyFile}"
+    (
+    ) > "${awsCliPublicKeyFile}"
+    gpg --quiet --import "${awsCliPublicKeyFile}"
```
Chuxel, line 32: I think this is left over content - I'm getting an error from here when running the script.

script-library/awscli-debian.sh
```diff
@@ -0,0 +1,76 @@
...
+
+set -e
+
+AWSCLI_VERSION=${1:-"latest"}
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    exit 1
+fi
+
```
Chuxel, line 20: To get this model to work, you'd still include the key as the default along with the get_common_setting script.  Add the following:
```suggestion
AWSCLI_GPG_KEY=FB5DB77FD5C118B80511ADA8A6310ACC4672475C
AWSCLI_GPG_KEY_MATERIAL="-----BEGIN PGP PUBLIC KEY BLOCK-----

mQINBF2Cr7UBEADJZHcgusOJl7ENSyumXh85z0TRV0xJorM2B/JL0kHOyigQluUG
ZMLhENaG0bYatdrKP+3H91lvK050pXwnO/R7fB/FSTouki4ciIx5OuLlnJZIxSzx
PqGl0mkxImLNbGWoi6Lto0LYxqHN2iQtzlwTVmq9733zd3XfcXrZ3+LblHAgEt5G
TfNxEKJ8soPLyWmwDH6HWCnjZ/aIQRBTIQ05uVeEoYxSh6wOai7ss/KveoSNBbYz
gbdzoqI2Y8cgH2nbfgp3DSasaLZEdCSsIsK1u05CinE7k2qZ7KgKAUIcT/cR/grk
C6VwsnDU0OUCideXcQ8WeHutqvgZH1JgKDbznoIzeQHJD238GEu+eKhRHcz8/jeG
94zkcgJOz3KbZGYMiTh277Fvj9zzvZsbMBCedV1BTg3TqgvdX4bdkhf5cH+7NtWO
lrFj6UwAsGukBTAOxC0l/dnSmZhJ7Z1KmEWilro/gOrjtOxqRQutlIqG22TaqoPG
fYVN+en3Zwbt97kcgZDwqbuykNt64oZWc4XKCa3mprEGC3IbJTBFqglXmZ7l9ywG
EEUJYOlb2XrSuPWml39beWdKM8kzr1OjnlOm6+lpTRCBfo0wa9F8YZRhHPAkwKkX
XDeOGpWRj4ohOx0d2GWkyV5xyN14p2tQOCdOODmz80yUTgRpPVQUtOEhXQARAQAB
tCFBV1MgQ0xJIFRlYW0gPGF3cy1jbGlAYW1hem9uLmNvbT6JAlQEEwEIAD4WIQT7
Xbd/1cEYuAURraimMQrMRnJHXAUCXYKvtQIbAwUJB4TOAAULCQgHAgYVCgkICwIE
FgIDAQIeAQIXgAAKCRCmMQrMRnJHXJIXEAChLUIkg80uPUkGjE3jejvQSA1aWuAM
yzy6fdpdlRUz6M6nmsUhOExjVIvibEJpzK5mhuSZ4lb0vJ2ZUPgCv4zs2nBd7BGJ
MxKiWgBReGvTdqZ0SzyYH4PYCJSE732x/Fw9hfnh1dMTXNcrQXzwOmmFNNegG0Ox
au+VnpcR5Kz3smiTrIwZbRudo1ijhCYPQ7t5CMp9kjC6bObvy1hSIg2xNbMAN/Do
ikebAl36uA6Y/Uczjj3GxZW4ZWeFirMidKbtqvUz2y0UFszobjiBSqZZHCreC34B
hw9bFNpuWC/0SrXgohdsc6vK50pDGdV5kM2qo9tMQ/izsAwTh/d/GzZv8H4lV9eO
tEis+EpR497PaxKKh9tJf0N6Q1YLRHof5xePZtOIlS3gfvsH5hXA3HJ9yIxb8T0H
QYmVr3aIUes20i6meI3fuV36VFupwfrTKaL7VXnsrK2fq5cRvyJLNzXucg0WAjPF
RrAGLzY7nP1xeg1a0aeP+pdsqjqlPJom8OCWc1+6DWbg0jsC74WoesAqgBItODMB
rsal1y/q+bPzpsnWjzHV8+1/EtZmSc8ZUGSJOPkfC7hObnfkl18h+1QtKTjZme4d
H17gsBJr+opwJw/Zio2LMjQBOqlm3K1A4zFTh7wBC7He6KPQea1p2XAMgtvATtNe
YLZATHZKTJyiqA==
=vYOk
-----END PGP PUBLIC KEY BLOCK-----"

get_common_setting() {
    if [ "${common_settings_file_loaded}" != "true" ]; then
        curl -sfL "https://aka.ms/vscode-dev-containers/script-library/settings.env" 2>/dev/null -o /tmp/vsdc-settings.env || echo "Could not download settings file. Skipping."
        common_settings_file_loaded=true
    fi
    if [ -f "/tmp/vsdc-settings.env" ]; then
        local multi_line=""
        if [ "$2" = "true" ]; then multi_line="-z"; fi
        local result="$(grep ${multi_line} -oP "$1=\"?\K[^\"]+" /tmp/vsdc-settings.env | tr -d '\0')"
        if [ ! -z "${result}" ]; then declare -g $1="${result}"; fi
    fi
    echo "$1=${!1}"
}
```

This will update the setting if one is found and otherwise use what is embedded in the script.  This is particularly critical for testing.

script-library/awscli-debian.sh
```diff
@@ -0,0 +1,76 @@
...
+    # See Linux install docs at https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
+    if [ "${AWSCLI_VERSION}" != "latest" ]; then
+        local versionStr=-${AWSCLI_VERSION}
+    fi
+    arch=$(uname -m)
+    if [ "$arch" != "x86_64" -a "$arch" != "aarch64" ]; then
+        echo "AWS CLI does not support machine architecture '$arch'. Please use an x86-64 or ARM64 machine."
+        exit 1
+    fi
```
Chuxel, line 55: `uname -m` can come back with a number of differrent values for arm64 unfortunatley... arm64 and armv8l are two other examples in addition to aarch64.

Using `dpkg --print-architecture` gives you two reliable values on debian - either amd64 or arm64 is more reliable.  This should do what you are trying to do I think:
```suggestion
    arch=$(dpkg --print-architecture)
    case "${arch}" in
        amd64) arch=x86_64 ;;
        arm64) arch=aarch64 ;;
        *)
            echo "AWS CLI does not support machine architecture '$arch'. Please use an x86-64 or ARM64 machine."
            exit 1
  esac
```

script-library/awscli-debian.sh
```diff
@@ -0,0 +1,76 @@
...
+        echo "Could not verify GPG signature of AWS CLI install script. Make sure you provided a valid version."
+        exit 1
+    fi
+
+    unzip "${scriptZipFile}"
+    ./aws/install
+
+    rm -rf ./aws
+}
+
```
Chuxel, line 71: I noticed that this script fails with the standard debian and ubuntu images due to the fact that curl and gpg are not in them. You can add the following to make sure these are installed if they're missing:

```suggestion
# Function to run apt-get if needed
apt_get_update_if_needed()
{
    if [ ! -d "/var/lib/apt/lists" ] || [ "$(ls /var/lib/apt/lists/ | wc -l)" = "0" ]; then
        echo "Running apt-get update..."
        apt-get update
    else
        echo "Skipping apt-get update."
    fi
}

# Checks if packages are installed and installs them if not
check_packages() {
    if ! dpkg -s "$@" > /dev/null 2>&1; then
        apt_get_update_if_needed
        apt-get -y install --no-install-recommends "$@"
    fi
}

export DEBIAN_FRONTEND=noninteractive

check_packages curl ca-certificates gnupg2 dirmngr
```

Chuxel (commented):
@Rabadash8820 Did an in-depth review and have some suggestions.  You can use "commit suggestions" if they make sense.  If it's easier we can also make these tweaks post merge. 

One thing I noticed is that the regression test appears to be giving a false positive which is why  some of this did not appear.

Chuxel:
Pulled in the regression test fix so you can see errors as the CI jobs run now as well.

script-library/awscli-debian.sh
Rabadash8820: Ok, I think I follow... I don't love having this giant base64 blob in two places, but I get that the script needs a default. Why would I add the `get_common_setting` function though, if it's already in `shared/utils.sh`? I see that this function exists in the related `azcli-debian.sh` script, but I don't get why its necessary there either. The body is identical to the shared function.

script-library/awscli-debian.sh
Rabadash8820: Fixed in eb14d6934bf0022eef3791f94e251e734dae9c69

script-library/awscli-debian.sh
Rabadash8820: Good idea. Fixed in eb14d6934bf0022eef3791f94e251e734dae9c69.

script-library/awscli-debian.sh
Rabadash8820: So I actually just noticed that these `apt_get_update_if_needed` and `check_packages` functions are already defined in the `shared/utils.sh` script. Why do they need to be defined again here (and in the Azure CLI install script)?

script-library/awscli-debian.sh
Rabadash8820: Oops. 😅 Fixed in eb14d6934bf0022eef3791f94e251e734dae9c69.

script-library/awscli-debian.sh
Chuxel: We do not download and source that file in separately right now.  Users do tend to go to these files directly and curl the script in without the utility as well, so we try to keep them self-contained.  So, there's a legacy here we're dealing with TBH.

As the features functionality stabilizes, we'll shuffle things around, but there's a bit of a dual use here today.

script-library/awscli-debian.sh
Chuxel: We do not download and source that file in separately right now.  Users do tend to go to these files directly and curl the script in without the utility as well, so we try to keep them self-contained.  So, there's a legacy here we're dealing with TBH.

As the features functionality stabilizes, we'll shuffle things around, but there's a bit of a dual use here today.

So, this is point-in-time.

Chuxel: Oh and on the blob, yes - the idea with the central file is for out of band updates. There are pros-and-cons with having a default value here. Technically we could separate that part of the PR out, merge that first, then merge the script itself.  That would be the alternative.

Rabadash8820: Fair enough, I can accept "because legacy" lol. Added `get_common_setting` in 72b8a9ab3e94a66c389bd7f05fc9e90003e0419d.

Rabadash8820:
Awesome, thanks @Chuxel, this was fun. :P Any idea when I can expect to see this feature live in the VS Code list?

Chuxel:
@Rabadash8820 It will appear in the UX at the latest with the next VS Code release (start of next month) - if there's "recovery builds" it could go earlier. 

If you switch to the preview version of the extension (which is the default for VS Code Insiders), you'll see it sooner. Typically whenever there's a release of this repository, we generally get it into insiders the next day which will happen multiple times during the development iteration.

Over time as we decouple release processes a bit, the updates in stable will happen sooner as well.

joshspicer:
Additionally, next time we [create a release](https://github.com/microsoft/vscode-dev-containers/releases) from this repo, you'll be able to reference the feature directly from a `devcontainer.json` with our experimental "remote `features`" syntax

##### `devcontainer.json`:
```json
"features": {
   "microsoft/vscode-dev-containers/aws-cli": "latest"
}

```

This syntax will pull directly from the latest GitHub release, whereas (at the moment) the shorthand `aws-cli` is pulling from a cache that we release alongside the Remote-Containers/Codespaces extensions.

Rabadash8820:
Awesome, thanks for the info @Chuxel @joshspicer. Looking forward to the future of devcontainers!

Chuxel:
Deployed!

PR #1314, opened 2022-02-19
Add R - Jupyter - R Markdown - Data Science - Machine Learning DevConatiner

R-icntay:
An environment to perform Data Science and Machine Learning in R with support for .R scripts, Jupyter Notebooks, and R Markdown Notebooks.

PR #1327, opened 2022-02-28
[R] Support for R package pre-installed images and arm64 images

eitsupi:
Suggested by https://github.com/microsoft/vscode-dev-containers/pull/1314#issuecomment-1049423896.

1. Allows the base image to be changed from the VSCode UI.
  We can choose an image with the tidyverse package installed, which is considered to be in high demand (`rocker/tidyverse`), or a huge image with a large number of R packages and jupyter installed (`rocker/binder`).
2. Since a user `rstudio` exists in `rocker/tidyverse` and `rocker/binder` as a user with UID 1000, change the user name from `vscode` to `rstudio` to match this.
3. Allow the build args to control the installation of jupyter.
4. For arm64 builds (only exist for `rocker/r-ver`), the R packages are source installation, so add the necessary dependency packages in that case (`libfontconfig1-dev` and `libcairo2-dev`).

cc @R-icntay

### DIFF
--- a/containers/r/.devcontainer/Dockerfile
+++ b/containers/r/.devcontainer/Dockerfile
@@ -1,18 +1,19 @@
-# R version: 4, 4.1, 4.0
-ARG VARIANT="4"
-FROM rocker/r-ver:${VARIANT}
+# [Choice] R version: 4, 4.1, 4.0
+ARG VARIANT=4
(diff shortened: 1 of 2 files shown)


Chuxel:
LGTM!

Issue #1052, opened 2021-09-16
[PHP] Debugging Error Xdebug: [Step Debug] Could not connect to debugging client #11


stefnats:
- VSCode Version: 1.60.1
- Local OS Version: macOS 11.5.2
- Name of Dev Container Definition with Issue: PHP

Hello,

i'm using this container with the following config (`devcontainer.json`):

```
// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
// https://github.com/microsoft/vscode-dev-containers/tree/v0.191.0/containers/php
{
	"name": "PHP",
	"build": {
		"dockerfile": "Dockerfile",
		"args": { 
			// Update VARIANT to pick a PHP version: 8, 8.0, 7, 7.4, 7.3
			"VARIANT": "8",
			"NODE_VERSION": "lts/*"
		}
	},
	
	// Set *default* container specific settings.json values on container create.
	"settings": { 
		"php.validate.executablePath": "/usr/local/bin/php"
	},

	// Add the IDs of extensions you want installed when the container is created.
	"extensions": [
		"felixfbecker.php-debug",
		"bmewburn.vscode-intelephense-client",
		"mrmlnc.vscode-apache",
		"calebporzio.better-phpunit"
	],

	// Use 'forwardPorts' to make a list of ports inside the container available locally.
	// "forwardPorts": [8080],

	// Use 'portsAttributes' to set default properties for specific forwarded ports. More info: https://code.visualstudio.com/docs/remote/devcontainerjson-reference.
	"portsAttributes": {
		"8000": {
			"label": "Hello Remote World",
			"onAutoForward": "notify"
		}
	},

	// Use 'otherPortsAttributes' to configure any ports that aren't configured using 'portsAttributes'.
	// "otherPortsAttributes": {
	// 		"onAutoForward": "silent"
	// },

	// Use 'postCreateCommand' to run commands after the container is created.
	// "postCreateCommand": "sudo chmod a+x \"$(pwd)\" && sudo rm -rf /var/www/html && sudo ln -s \"$(pwd)\" /var/www/html"

	// Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
	"remoteUser": "vscode"
}
```

When i run any php process i receive this error

```
Xdebug: [Step Debug] Could not connect to debugging client. Tried: localhost:9000 (through xdebug.client_host/xdebug.client_port) :-(
```

[...]

Issue #695, opened 2020-12-19
Xdebug version 3.0...  doesn't work in php Dev Container 


vinser:
- VSCode Version: 1.52.0
- Local OS Version: Windows 10 Pro 20H2 build 19042.630
- Name of Dev Container Definition with Issue: vscode-dev-containers/containers/php/

Steps to Reproduce: 
1. Start apache2ctl start
2. Set breakpoint and launch Xdebug (f5)
3. Open localhost:8080. Debugger will not catch breakpoint.

Workaround:
1. Use [base.Dockerfile](https://github.com/microsoft/vscode-dev-containers/blob/master/containers/php/.devcontainer/base.Dockerfile)
2. Replace code:

`# Install xdebug
RUN yes | pecl install xdebug \
    && echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" > /usr/local/etc/php/conf.d/xdebug.ini \
    && echo "xdebug.remote_enable=on" >> /usr/local/etc/php/conf.d/xdebug.ini \
    && echo "xdebug.remote_autostart=on" >> /usr/local/etc/php/conf.d/xdebug.ini \
    && rm -rf /tmp/pear`

with something like this:

`# Install xdebug
RUN pecl install xdebug && docker-php-ext-enable xdebug \
    && echo "xdebug.mode = debug" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini \
    && echo "xdebug.start_with_request = yes" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini \
    && echo "xdebug.client_host = host.docker.internal" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini \
    && echo "xdebug.discover_client_host = false" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini \
    && echo "xdebug.client_port = 9003" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini \
    && echo "xdebug.log = /var/www/html/xdebug.log" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
`

3. Don't forget to change port 9000 to 9003 in launch.json 

It works for me. 

It will be nice to see mcr.microsoft.com/vscode/devcontainers/php container update.

Edit12/26/2020: 
[...]

PR #1329, opened 2022-02-28
Switch to correct Xdebug port 9003

mgenereu:
Xdebug was hard to use with the existing setup.  This switches a few configuration items to make it easier to use the debugger.

- Port 80 and 8080 was causing confusion especially when auto-forwarding the ports and opening browsers automatically.
- Port 9003 was offering to auto-forward.  This never needs to forward as the VS Code debugger runs on localhost in the container.
- XDebug is on 9003 with the new version.  Using 9000 is just begging for confusion.
- Apache is always complaining about the ServerName so localhost is set as the default.
- The Apache VirtualHost is set to the same port as the main configuration.  This not matching caused side effects.

Also, regarding #1052, I use this setup with the VS Code debugger running whenever I call PHP directly or via Apache so I don't get the warning.

### DIFF
--- a/containers/php/.devcontainer/devcontainer.json
+++ b/containers/php/.devcontainer/devcontainer.json
@@ -2,7 +2,7 @@
 	"name": "PHP",
 	"build": {
 		"dockerfile": "Dockerfile",
-		"args": { 
+		"args": {
 			// Update VARIANT to pick a PHP version: 8, 8.1, 8.0, 7, 7.4
 			// Append -bullseye or -buster to pin to an OS version.
 			// Use -bullseye variants on local on arm64/Apple Silicon.
@@ -10,25 +10,30 @@
 			"NODE_VERSION": "lts/*"
 		}
 	},
-	
 	// Set *default* container specific settings.json values on container create.
-	"settings": { 
+	"settings": {
 		"php.validate.executablePath": "/usr/local/bin/php"
 	},
-
 	// Add the IDs of extensions you want installed when the container is created.
 	"extensions": [
 		"xdebug.php-debug",
 		"bmewburn.vscode-intelephense-client",
 		"mrmlnc.vscode-apache"
 	],
-
-	// Use 'forwardPorts' to make a list of ports inside the container available locally.
-	"forwardPorts": [8080],
-
+	// Use 'portsAttributes' to control auto forwarding of ports
+	"portsAttributes": {
+		"8080": {
+			"label": "Application",
+			"onAutoForward": "openBrowser",
+			"requireLocalPort": true
+		},
+		"9003": {
+			"label": "PHP Xdebug",
+			"onAutoForward": "ignore"
+		}
+	},
 	// Use 'postCreateCommand' to run commands after the container is created.
 	// "postCreateCommand": "sudo chmod a+x \"$(pwd)\" && sudo rm -rf /var/www/html && sudo ln -s \"$(pwd)\" /var/www/html"
-
 	// Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
 	"remoteUser": "vscode"
 }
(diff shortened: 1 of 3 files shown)


Chuxel:
Hmmm - Re: switching to 9003, the challenge here is that updating the image can affect people that already have things set up for port 9000. So, it's a bit of a breaking change. Everything else seems like it would be a clean and a good update tho!

mgenereu:
> it's a bit of a breaking change.

Is there a way to announce breaking changes?  People were already having trouble using it.  Is there a place to put documentation and announcements for the specific container? I would add additional documentation on how to use this container for PHP if I knew where to put it.

mgenereu:
I can't figure out why the smoke test is failing.  I looked for hidden characters with `od` but it looks like healthy JSON.

mgenereu:
Speaking of breaking changes, I forgot about this line that the upstream container recommends for development.  That's another big one that solved PHP problems downstream.
```
cp "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
```

mgenereu:
I just made a new `launch.json` file and it automatically populated with port 9003.  This might not be that breaking per #695? Thoughts, @Chuxel?  With a little feedback, I can continue to refine.

containers/php/.devcontainer/devcontainer.json
```diff
@@ -2,33 +2,37 @@
...
 	],
-
-	// Use 'forwardPorts' to make a list of ports inside the container available locally.
-	"forwardPorts": [8080],
-
+	// Use 'portsAttributes' to control auto forwarding of ports
+	"portsAttributes": {
+		"8080": {
+			"label": "Application",
+			"onAutoForward": "notify",
```
Chuxel, line 27: There is a hanging comma here that is causing the CI to fail.

Chuxel (commented):
> I just made a new launch.json file and it automatically populated with port 9003. This might not be that breaking per https://github.com/microsoft/vscode-dev-containers/issues/695? Thoughts, @Chuxel? With a little feedback, I can continue to refine.

The problem here is that the image (mcr.microsoft.com/vscode/devcontainers/php) can be used independant of updates to the definition.  Any changes made to the image would affect existing users that have a .devcontainer folder in their project.  What #695 was about was actually changing what was in the image to work - which is why that was clean.

So the question becomes - Is a port update worth that problem?  We'd basically have to add some text in the terminal telling people to make changes since they otherwise would not know. We have a mechanism to do it, but it will drive issues and questions.

For port `80` for example, we could actually add `onAutoForward` with a value of `ignore`  in devcontainer.json, which avoids that problem. Within the container, port 80 along with 8080 would then still be available which wouldn't break anything existing.

One possible way to manage `9003` is to have a separate build argument that indicates what should happen that is used in both the `base.Dockerfile` (used to generate the image) and `Dockerfile` that could be used to set alternate options. If these are not set, then the config is as it is now, but if it is set, then it would take whatever the build arguments were set to.

> Speaking of breaking changes, I forgot about this line that the upstream container recommends for development. That's another big one that solved PHP problems downstream.
>
> > cp "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"

Yeah it sounds like this is mainly log output differences given the description which shouldn't break people I would expect. What problems did you see it resolve down stream? Assuming its not removing functionality vereses the default config, that's a safer update IMHO.

Also +@joshspicer since PHP config in the default Codespaces image is the same as this.

containers/php/.devcontainer/devcontainer.json
mgenereu: Didn't know `JSON with Comments` was a relaxed mode for parsing JSON.  That was a new one for me.

mgenereu:
I get the existing developer and downstream breakage but as a new developer using the current extensions, things were broken and I had to do some digging.  I'll close this PR here to document for others what fixes made the developer experience better for me.

Issue #704, opened 2021-01-03
Ruby containers should not have rbenv and rvm installed


jarrodldavis:
- VSCode Version: 1.52.1
- Local OS Version: macOS Big Sur 11.1
- Name of Dev Container Definition with Issue: `ruby`, `ruby-rails`

Steps to Reproduce:

1. Generate a new Ruby on Rails app
2. Add `ruby-rails` development container definition
3. Reopen folder in container
4. Open a terminal and `cd` into the workspace directory (e.g. `cd .` or `cd /workspace` from a new Integrated Terminal)
5. Observe that `rvm` complains about the Ruby version specified in the `Gemfile` isn't installed, even though the development container specifically has (the correct version of) Ruby installed (see #690).

Why are `rbenv` and `rvm` installed on Ruby development containers? We can already select the version of Ruby we want installed by using a particular Docker image tag, so there's no need for a Ruby version manager; they generate unnecessary noise and are difficult to remove from the pre-built containers. At the very least, there should be a mechanism to conditionally prevent loading the `rvm` and `rbenv` initialization scripts using an environment variable.

Chuxel:
These were not in the containers originally in the containers, but developers requested them be added so they can take advantage of the related config files. So, removing them is not ideal.

That said, we can, however, "eat" these messages and prevent them from appearing or allow disabling them as you mention.

Chuxel:
This ENV based approach would also apply to the codespaces-linux image.

csutter:
This is also causing issues with the Ruby VS Code extension for me, which seems to detect rbenv/rvm being present and tries to load Ruby through them (instead of using the container's system Ruby): `Required ruby-3.0.2 is not installed.`

I'm moving towards using custom devcontainer `Dockerfile`s based directly off the upstream ecosystem images (e.g. `ruby:3`, because the "kitchen sink" approach of these official VS Code doesn't really work for me personally. Having the ability to more aggressively disable a lot of this stuff might tempt me back 😃

slaytanic:
> This is also causing issues with the Ruby VS Code extension for me, which seems to detect rbenv/rvm being present and tries to load Ruby through them (instead of using the container's system Ruby): `Required ruby-3.0.2 is not installed.`

I had the same issue and solved it by adding `RUN rm /etc/profile.d/rvm.sh` to my `Dockerfile`

simmerz:
[...]

PR #1333, opened 2022-03-03
Issue 704: Adding enabling and disabling of RBENV and RVM packages

KanishkT123:
# [Work Item ID](#704)

## Description

---

Allows for RBENV and RVM tools to be enabled or disabled in the Ruby dev container.

## PR Checklist

---

- [X] I have added tests to cover my changes.
- [X] I ran the lint checks which produced no new errors nor warnings for my changes.

---

- [ ] Yes
- [X] No

> If this introduces a breaking change, please describe the impact and migration path for existing applications below.

## Testing

---

> - Instructions for testing and validation of your code:
All smoke testing is done in the CI pipeline of the VS Code Remote Containers Repository. A sample run can be found here: 
https://github.com/microsoft/vscode-dev-containers/runs/5378725362?check_suite_focus=true

PR has been reviewed and approved internally by @juzuluag , @bderusha  and 
@dnastrain 

---

> - Any other information or known dependencies that is important to this PR.

1. Changes to `ruby-debian.sh` script that modifies the `bashrc` and `zshrc` files in a devcontainer to allow conditional enabling of `rbenv` and `rvm`.
2. Changes to `devcontainer.json` to include two new ENV Variables for built containers: `RBENV_ENABLED` and `RVM_ENABLED` which are read by the `bashrc` and `zshrc` to enable/disable RBENV and RVM
3. Default enables `rbenv` and `rvm` in `base.Dockerfile` in order to reduce workflow disruption.
4. Default disables `rbenv` and `rvm` in `devcontainer.json` in order to reduce tooling overhead for devcontainer users.


containers/ruby/.devcontainer/library-scripts/ruby-debian.sh
```diff
@@ -247,25 +247,48 @@ git clone --depth=1 \
...
+    updaterc 'if [ "${RBENV_ENABLED}" != "false" ]; then
+        chown -R ${USERNAME} /home/${USERNAME}/.rbenv
+        ln -s /usr/local/share/ruby-build /home/${USERNAME}/.rbenv/plugins/ruby-build > /dev/null 2>&1
+    else
+        rm -f /home/${USERNAME}/.rbenv/plugins/ruby-build > /dev/null 2>&1
+    fi'
+fi
+
+updaterc 'if [ "${RBENV_ENABLED}" != "false" ]; then
+        sudo ln -s /usr/local/share/rbenv/bin/rbenv /usr/local/bin > /dev/null 2>&1
```
Chuxel, line 283: This will not work in the majority of images since `sudo` is typically not available unless explicitly installed.  I don't know that there's really a problem with it being in the path in either the rvm or rbenv case is there?  

If neither are sourced I wouldn't expect to see conflicts. Or is there something I'm not aware of?

If being in the path is a problem, perhaps a better approach is to reference the full path in the next line instead and remove the symlink entirely.

```
eval "$(/usr/local/share/rbenv/bin/rbenv init -)"
```

You could also explicitly update the PATH in this scenario if that's needed (export PATH=$PATH:/usr/local/share/rbenv/bin)

containers/ruby/.devcontainer/library-scripts/ruby-debian.sh
```diff
@@ -247,25 +247,48 @@ git clone --depth=1 \
...
+        eval "$(rbenv init -)"
+        sudo ln -s /usr/local/share/ruby-build /root/.rbenv/plugins/ruby-build > /dev/null 2>&1
+    else
+        rm -f  /root/.rbenv/plugins/ruby-build > /dev/null 2>&1
+        rm -f  /usr/local/bin/rbenv > /dev/null 2>&1
```
Chuxel, line 288: See previous comment on removing the symlinks. I would only remove things for the current user. Permissions can get messed up here.  

Honestly, I also would just leave the .rbenv folder.  It does no harm to be there.

containers/ruby/.devcontainer/library-scripts/ruby-debian.sh
```diff
@@ -247,25 +247,48 @@ git clone --depth=1 \
...
     -c receive.fsck.zeroPaddedFilemode=ignore \
     https://github.com/rbenv/ruby-build.git /usr/local/share/ruby-build
+
 mkdir -p /root/.rbenv/plugins
-ln -s /usr/local/share/ruby-build /root/.rbenv/plugins/ruby-build
+
 if [ "${USERNAME}" != "root" ]; then
     mkdir -p /home/${USERNAME}/.rbenv/plugins
-    chown -R ${USERNAME} /home/${USERNAME}/.rbenv
-    ln -s /usr/local/share/ruby-build /home/${USERNAME}/.rbenv/plugins/ruby-build
```
Chuxel, line 264: I think these lines are still needed to avoid symlinking using sudo.

containers/ruby/.devcontainer/library-scripts/ruby-debian.sh
```diff
@@ -247,25 +247,48 @@ git clone --depth=1 \
...
 git clone --depth=1 \
     -c core.eol=lf \
     -c core.autocrlf=false \
     -c fsck.zeroPaddedFilemode=ignore \
     -c fetch.fsck.zeroPaddedFilemode=ignore \
     -c receive.fsck.zeroPaddedFilemode=ignore \
     https://github.com/rbenv/ruby-build.git /usr/local/share/ruby-build
+
 mkdir -p /root/.rbenv/plugins
-ln -s /usr/local/share/ruby-build /root/.rbenv/plugins/ruby-build
```
Chuxel, line 260: I would just leave this in place so you do not need to mess with it in the bashrc file.

Chuxel (commented):
Some comments on initial review.

containers/ruby/.devcontainer/library-scripts/ruby-debian.sh
KanishkT123: There is not an explicit problem, I think the only reason we did it this way is to try to keep the user path as clean as possible. However, it's not actually necessary, so we could remove it.

KanishkT123: I do like the idea of just updating the PATH directly though, so we'll do that.

Chuxel:
So reviewing this - the changes look good... but one thing that is flummoxing me is it still seems like somehow rvm is getting sourced when I set `"RVM_ENABLED":"false"`  Are you seeing that as well?

Chuxel:
Ah - found it. It looks like rvm is installing a file at `/etc/profile.d/rvm.sh` that we didn't expect. This is getting picked up when VS Code starts rather than when a new terminal is created. 

I suspect this will need to have the same condition added to it.

KanishkT123:
> Ah - found it. It looks like rvm is installing a file at `/etc/profile.d/rvm.sh` that we didn't expect. This is getting picked up when VS Code starts rather than when a new terminal is created.
> 
> I suspect this will need to have the same condition added to it.

I'm not sure exactly how this works. Do you mean modifying the `rvm.sh` file directly, or do you mean maybe having the .bashrc file delete the `rvm.sh` file in `profile.d`?

Chuxel:
@KanishkT123 Yeah, that is a reasonable way to handle it I think.  I don't think this originally happened but got added somewhere along the way by the rvm installer.

PR #1335, opened 2022-03-05
Add JupyterLab as a feature

bratsche:
If Python is not installed, error out and tell the user that it's missing for now. The intent here is to improve this to auto-install Python later, but this seems like a good first step.


Chuxel:
Wouldn't it make sense to check to see if it's already there before installing?  I'd also recommend checking this with the python-3 and python-3-anaconda definitions since I gartunee people will use it with that.

You should probably also deal with cached wheels, etc here. This will leave a lot of unneeded content in the image (making it even bigger).  Adding `--no-cache-dir` at a minimum to the pip install would make sense.

@dynamicwebpaige Does it make sense to do this as a straight pip install (verses pipx)?   I kind of assume so, but thought I'd ask.

Chuxel:
A few other thoughts here in thinking about it. 

Features are installed using the "root" user. The problem here is that you could end up inadvertently end up doing one of two things:

1. Locking up the pip cache and management files that can only be edited by root
2. Installing in the user's pip folder, which in this case would be "root" and therefore not accessible by a non-root user like codespace / vscode / node, etc

I'd verify what happens on disk after this is installed to be sure nothing major breaks.  You also may want to set `umask 0002` so that any group access is retained... or detect the non-root user and fire pip there.

Also - for codespaces-linux, be sure to verify the scenario where a version of python is dynamically acquired by Oryx.  If you use "pipx" you'd be abstracted from those kinds of changes.  But I'm not sure what would happen if you install when the default version of python in the image and then a different one ends up getting used to.  I'd be worried it would just totally break.

script-library/container-features/src/devcontainer-features.json
```diff
@@ -1121,6 +1121,24 @@
...
+			"documentationURL": "https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/jupyterlab.md",
+			"options": {
+				"version": {
+					"type": "string",
+					"proposals": ["latest", "3.6.2"],
+					"default": "latest",
+					"description": "Select or enter a jupyterlab version."
+				}
+			},
+			"buildArg": "_VSC_INSTALL_JUPYTERLAB",
```
joshspicer, line 1137: `buildArg` is actually a [legacy attritbute](https://github.com/microsoft/vscode-remote-containers/blob/main/src/configuration/containerFeaturesConfiguration.ts#L22) now, we auto generate the build environment variables programmatically via the `id` (for consistency)

We fall back to buildArg right now, but would be nice to follow the programatic `_BUILD_ARG_${idSafe}=true` for new features.  

We also need to include this somewhere in the `feature-scripts.env`, or it won't be installed by the [`install.sh`](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/container-features/src/install.sh)script (Which is the only thing implicitly called by the tooling to install).   I'd suggest placing it underneath the python scripts, since this file actually enforces install order (although we eventually want to move that logic into the tooling itself to simplify things (https://github.com/microsoft/dev-container-spec/issues/16)

script-library/container-features/src/devcontainer-features.json
```diff
@@ -1121,6 +1121,24 @@
...
+			"extensions": [],
```
joshspicer, line 1138: I'd imagine we'd want some vscode extensions installed by default, namely the pylance/python extension and the Jupyter extension

script-library/jupyterlab-debian.sh
```diff
@@ -0,0 +1,28 @@
...
+#-------------------------------------------------------------------------------------------------------------
+#
+# Docs: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/jupyterlab.md
+# Maintainer: The VS Code and Codespaces Teams
+#
+# Syntax: ./jupyter-debian.sh
+
+set -e
+
+JUPYTER_LAB_VERSION=${1:-"latest"}
```
joshspicer, line 14: I think it would make more sense to just call this `VERSION`, as it lines up with the option name in the `devcontainer-features.json` file above.

And then somewhere around this part of the file: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/container-features/src/feature-scripts.env#L14

Add in something like:

```
_VSC_INSTALL_JUPYTERLAB="-jupyterlab-debian.sh ${_BUILD_ARG_JUPYTERLAB_VERSION:-latest}"
```

Again, that `_BUILD_ARG...` is auto-generated by the user setting that version property in their `devcontainer.json`, and code-gen'd here: https://github.com/microsoft/vscode-remote-containers/blob/main/src/node/containerFeatures.ts#L180-L193

script-library/container-features/src/devcontainer-features.json
```diff
@@ -1121,6 +1121,24 @@
...
+			"include": [
```
joshspicer, line 1139: `include` and `exclude` need a bit of work, but today this would only have this feature show up for `codespaces-linux` and none of our other base definitions.  I don't think there's any reason to be so restrictive?

joshspicer:
If it's helpful, @Rabadash8820 's PR adding in the `aws-cli` is a great example of contributing a new feature https://github.com/microsoft/vscode-dev-containers/pull/1326

script-library/container-features/src/devcontainer-features.json
bratsche: @joshspicer When I try to follow the pattern from the example template repo and do:

```sh
if [ ! -z ${_BUILD_ARG_JUPYTERLAB} ]; then
   .. install
fi
```

It doesn't seem to work. But everything is working fine if I do the install without checking for that. It seems like it should be getting that from the id attribute which is set to `jupyterlab`. Is that not enough?

joshspicer (approved):

bratsche:
Thanks for your comments @Chuxel!

> A few other thoughts here in thinking about it.
> 
> Features are installed using the "root" user. The problem here is that you could end up inadvertently end up doing one of two things:
> 
> 1. Locking up the pip cache and management files that can only be edited by root. My hope would be `--no-cache-dir` would avoid this, but I'd double check.
> 2. Installing in the user's pip folder, which in this case would be "root" and therefore not accessible by a non-root user like codespace / vscode / node, etc.
> 
> I'd verify what happens on disk after this is installed to be sure nothing major breaks. You also may want to set `umask 0002` so that any group access is retained... or detect the non-root user and fire pip there instead to add.

I updated it to install to the non-root user if there is one. This installs `jupyter-lab` to `~/.local/bin/jupyter-lab`

> Also - for codespaces-linux, be sure to verify the scenario where a version of python is dynamically acquired by Oryx. If you use "pipx" you'd be abstracted from those kinds of changes. But I'm not sure what would happen if you install when the default version of python in the image using pip and then a different one ends up getting used to. I'd be worried it would just totally break or, failing that, it wouldn't be in the virtual environment that was created.
> 
> e.g., add a `runtime.txt` file with `python-3.6.9` in it in the repo - that would case this to happen.
> 
> This one is **very** specific to that image... you wouldn't see it in a normal python / anaconda one. Unlike something like node, there's a lot of version specific magic tied to native compilation in pip packages.

I added the `runtime.txt` with `python-3.6.9` as you suggested. In the logs I'm seeing the following:

```
Downloading and extracting 'python' version '3.6.9' to '/opt/python/3.6.9'...
Downloaded in 1 sec(s).
Verifying checksum...
Extracting contents...
performing sha512 checksum for: python...
Done in 2 sec(s).

image detector file exists, platform is python..
/tmp/BuildScriptGenerator/0840f23b71354aec894853987bbe8595/build.sh: line 131: /etc/ld.so.conf.d/python.conf: Permission denied
Could not auto-build. Skipping.
```

And when I I run `python --version` it's showing me `3.8.12`.

I'm testing it outside of my branch to try to figure out if this is something that broke in my branch somehow.

bratsche:
That issue I'm having with `runtime.txt` seems to be happening outside my branch, fwiw.

bratsche:
Hacking the Dockerfiles to add `touch /etc/ld.so.conf.d/python.conf && chown codespace:codespace /etc/ld.so.conf.d/python.conf` fixes (well, works around) the issue above, but then I get another permission denied error after that: `/sbin/ldconfig.real: Can't create temporary cache file /etc/ld.so.cache~: Permission denied`. It seems like a kitchen sink bug with Oryx.

bratsche:
Possibly related to the `runtime.txt` issue: https://github.com/microsoft/Oryx/issues/1206

bratsche:
I worked around the issues with runtime.txt and confirmed that JupyterLab was working just fine with an Oryx-installed version of Python.

joshspicer:
Great - @bratsche and I chatted and now with additions to the PATH, smoke test passes
image

Issue #1367, opened 2022-03-16
Moby fails to install on ubuntu 21.04-based definitions


joshspicer:
To Repro:

1. In Remote-Containers or Codespaces, select the C++ definition based on ubuntu 21.04.
2. Add the `docker-in-docker` feature on `latest`.  Moby will == true by default
3. Rebuild, see error.

<img width="878" alt="image" src="https://user-images.githubusercontent.com/23246594/158597147-c555f34b-d5d4-4548-8698-3639ed81bf62.png">

If I set the base definition image to ubuntu 20.04, did not see this happen in codespaces.

joshspicer:
line in script: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docker-in-docker-debian.sh#L185

Chuxel:
We should trap and provide a better error since this distro version is not supported. 22.04 is about to come out as well which is the new LTS so the same thing will happen for a while.

joshspicer:
while it seems moby isn't offered by docker in 21.04, docker itself should be
https://docs.docker.com/engine/install/ubuntu/
![image](https://user-images.githubusercontent.com/23246594/158651015-9245fb61-d5f6-4a62-bd93-f2e5885361e6.png)

Chuxel:
Cool - We can trap each separately then.  Automatically switching has licensing implications, so we should error and then suggest updating.

joshspicer:
> Cool - We can trap each separately then. Automatically switching has licensing implications, so we should error and then suggest updating.

On it !

Chuxel:
Just as a quick follow up. Ubuntu 22.04 Jammy is slated to release on April 22nd. Hopefully the MS feeds will have the latest by then -- it's apparently in progress.

We'll stop building "impish" at this point given its already out of support.  Latest is still set to `focal`, so this only applies if you're pinning to this OS version.

Issue #1019, opened 2021-08-31
Add distro version check to scripts in script-library


Chuxel:
Relates to: Both

Currently scripts in the `script-library` are setup to assume that support exists for all the major LTS versions of Debian and Ubuntu. However, we've seen a few lag behind (notably the Azure CLI which doesn't yet have a `bullseye` package).

As a result we should add a distro check in scripts that would be affected. As we look at what we've been calling the "lego block" experience or "container features" (https://github.com/microsoft/vscode-remote-release/issues/5442), we can expect to see more mismatching than we have now and should provide good error messages when it happens.

joshspicer:
debating how I should do this. 

1. `apt-cache policy azure-cli | less` and try to parse out the `$VERSION_CODENAME` I get from /etc/os-release
2. Keep a "fail-list" that we maintain of LTS distros that we know break the script (eg: `azcli` fails if run when `$VERSION_CODENAME` == `bullseye`)
3. Some other, better/more obvious way

Did you have an idea how to accomplish this @chuxel?

Chuxel:
Yeah it probably varies a bit depending on how the bits are acquired.

> 1. `apt-cache policy azure-cli | less` and try to parse out the `$VERSION_CODENAME` I get from /etc/os-release
> 2. Keep a "fail-list" that we maintain of LTS distros that we know break the script (eg: `azcli` fails if run when `$VERSION_CODENAME` == `bullseye`)
> 3. Some other, better/more obvious way

The thing I notice is that every major release seems to require subtle updates to scripts that invariably catches me by suprise.  Bullseye for example doesn't have a `python` command by default when you install `python3-minimal` unless you install `python-is-python3`.  That caused problems with `node-gyp` in the Node definition much to my surprise.  Different versions of the script therefore will have different sets of distros they support.

I suspect we need to more tightly control when we switch to new distro versions and run a script regression on them. The arm64 changes I just made do that for the images themselves (we control what our MCR python:3 image is). We likely need to have an "allow" list rather than a "it doesn't work" list to be safe.

This will probably be even more important as we add "features" into the mix since people can mix and match any combination of things.

joshspicer:
tying this into the devcontainerv2 work, this could be useful as a more generic feature script function as a "sanity check"

joshspicer:
>  \> (notably the Azure CLI which doesn't yet have a bullseye package).

I have no problem now running the `./azcli-debian.sh` script in a bullseye container.

```bash
[...]

PR #1371, opened 2022-03-17
supported distro detection for docker-in-docker script

joshspicer:
closes https://github.com/microsoft/vscode-dev-containers/issues/1367

Provide clearer error messaging and the ability to dynamically update supported distro releases for both OSS Moby and docker installation.

For example, trying to use moby dind with hirsute will provide the following guidance:
```bash
2022-03-17 16:50:05.406Z: #7 [3/3] RUN bash /tmp/scripts/docker-in-docker-debian.sh "true" "${USERNAME}" "true"
2022-03-17 16:50:05.429Z: #7 sha256:df641496a1c8141a6497bf1279fdb806b1d428b8c0c88e0272b4b0792f370b2a
2022-03-17 16:50:06.577Z: #7 1.206 DOCKER_IN_DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES=buster bullseye bionic focal
2022-03-17 16:50:06.605Z: #7 1.206 (!) Unsupported  distribution version 'hirsute'. To resolve, either: (1) set feature option '"moby": false' , or (2) choose a compatible OS distribution
2022-03-17 16:50:06.632Z: #7 1.206 (!) Support distributions include:  buster bullseye bionic focal
2022-03-17 16:50:06.654Z: #7 ERROR: executor failed running [/bin/sh -c bash /tmp/scripts/docker-in-docker-debian.sh "true" "${USERNAME}" "true"]: exit code: 1
```

This set of codenames can be toggled as it reads from the common `get_common_setting` utilty function, which sources values from `script-library/shared/settings.env`

@chuxel  - Did I get all the supported docker/moby codenames right?

relevant to: https://github.com/microsoft/vscode-dev-containers/issues/1019


script-library/docker-in-docker-debian.sh
```diff
@@ -15,14 +15,27 @@ USE_MOBY=${3:-"true"}
 DOCKER_VERSION=${4:-"latest"} # The Docker/Moby Engine + CLI should match in version
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 DOCKER_DASH_COMPOSE_VERSION="1"
+DOCKER_IN_DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="buster bullseye bionic focal"
+DOCKER_IN_DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="buster bullseye bionic focal hirsute"
```
Chuxel, line 19: Ubuntu 21.10 "impish" also appears to be supported.

```suggestion
DOCKER_IN_DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="buster bullseye bionic focal hirsute Impish"
```

Chuxel:
Also - would it make sense to do this in docker-from-docker too?

migrations/readme-metadata-to-json.py
```diff
@@ -106,17 +106,17 @@ def template(defId, displayName, description, categories, defType, platforms, op
         options = 'lc rc'
 
     subst = '''
-		lc
```
joshspicer, line 109: Not relevant to this PR, oops :D

joshspicer: Still want it though.

script-library/docker-debian.sh
```diff
@@ -17,6 +17,9 @@ USE_MOBY=${5:-"true"}
 DOCKER_VERSION=${6:-"latest"}
 DOCKER_DASH_COMPOSE_VERSION=${7:-"v1"} # v1 or v2
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
+DOCKER_DASH_COMPOSE_VERSION="1"
```
jbreiding, line 20: I believe this line was removed in the previous commit.

Or is the pinning back to v1 intentional?

joshspicer: Ah! Thank you for catching this - removed.

PR #1372, opened 2022-03-17
[Feature] Adding Hugo as a Feature

samruddhikhandale:
Adding a new hugo feature. 

Things to look at by reviewers - 
- The `include[]` in devcontainer-features. As Hugo is built with Go, I added all the images which Go has in it's include[]. Should that be fine?


script-library/hugo-debian.sh
```diff
@@ -0,0 +1,113 @@
...
+# Install dependencies
+check_packages curl ca-certificates tar
+
+# Fetch latest version of Hugo if needed
+if [ VERSION="latest" ] || [ VERSION="lts" ]; then
+    export VERSION=$(curl -s https://api.github.com/repos/gohugoio/hugo/releases/latest | grep "tag_name" | awk '{print substr($2, 3, length($2)-4)}')
+fi
+
+# Install Hugo if it's missing
+if ! hugo version > /dev/null ; then
```
samruddhikhandale, line 92: Do we check if hugo is already installed or shall we check if hugo with $VERSION is already installed?

PR #1373, opened 2022-03-18
cache devcontainer collection

joshspicer:
@Chuxel  - I want to cache the `devcontainer-collection.json` from this repo in the npm package we embed with Remote-Containers - is this the right place?

### DIFF
--- a/build/config.json
+++ b/build/config.json
@@ -24,7 +24,8 @@
 		"LICENSE",
 		"package.json",
 		"yarn.lock",
-		".npmignore"
+		".npmignore",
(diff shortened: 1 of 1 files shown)


Chuxel:
Meaning storing it with the package or excluding it?  If you want to store it with the package, update this as well: https://github.com/microsoft/vscode-dev-containers/blob/aa031996c1413ec9cd76ca54e36cedf58153f963/package.json#L20

The spot you updated will stage the content, but this is the part that copies it into the output package.  Allows us to use content during the build process that don't land in the resulting package.

PR #1379, opened 2022-03-25
Add links to open in container

bamurtaugh:
Add the badge and link. Happy to discuss any wording or location suggestions!

cc @egamma @chrmarti


.devcontainer/README.md
```diff
@@ -1,9 +1,13 @@
 # Dev Container for vscode-dev-containers
 
+[![Open in Remote - Containers](https://img.shields.io/static/v1?label=Remote%20-%20Containers&message=Open&color=blue&logo=visualstudiocode)](https://vscode.dev/redirect?url=vscode://ms-vscode-remote.remote-containers/cloneInVolume?url=https://github.com/microsoft/vscode-dev-containers)
```
joshspicer, line 3: Oh wow, fancy!  Do we have a similar way to open up in a Codespace 🤔

Chuxel: Deep linking support would be needed for Codespaces on the web.  What this does is a redirect to a vscode:// URI.  Now the Codespaces extension could add support to handle those (since this even instlls the extension), but it wouldn't go to the web client if that's what you are thinking.

PR #1391, opened 2022-04-03
AWS CLI - GPG Fix

sktan:
Because the awscli Visual Studio Code feature does a `gpg --import`, we can't sign our commits when using it.

I've changed the behaviour to load it as a temporary keyring file so that it doesn't interfere with the default gpg keyring on setup. It should allow you to then use the forwarded GPG keyring from the host.


Rabadash8820:
Nope! Nice catch @sktan, I didn't realize that altering the public GPG keyring could cause issues like this; you taught me something new!

@Chuxel Out of curiosity, I did a [quick search](https://github.com/search?p=2&q=%22import%22+repo%3Amicrosoft%2Fvscode-dev-containers+language%3Ashell&type=Code) for other instances of `gpg --import` across this repo and got the following hits. Should probably apply a similar fix in these scripts:

- `containers/azure-terraform/.devcontainer/library-scripts/terraform-debian.sh`
- `script-library/terraform-debian.sh`
- `containers/codespaces-linux/.devcontainer/library-scripts/kubectl-helm-debian.sh`
- `containers/kubernetes-helm/.devcontainer/library-scripts/kubectl-helm-debian.sh`
- `containers/kubernetes-helm-minikube/.devcontainer/library-scripts/kubectl-helm-debian.sh`
- `script-library/kubectl-helm-debian.sh`
- `containers/go/.devcontainer/library-scripts/go-debian.sh`
- `containers/codespaces-linux/.devcontainer/library-scripts/go-debian.sh`
- `script-library/go-debian.sh`

script-library/awscli-debian.sh
```diff
@@ -92,15 +92,13 @@ verify_aws_cli_gpg_signature() {
...
     get_common_setting AWSCLI_GPG_KEY
     get_common_setting AWSCLI_GPG_KEY_MATERIAL true
-    local awsCliPublicKeyFile=aws-cli-public-key.pem
-    echo "${AWSCLI_GPG_KEY_MATERIAL}" > "${awsCliPublicKeyFile}"
-    gpg --quiet --import "${awsCliPublicKeyFile}"
+    local awsGpgKeyring=aws-cli-public-key.gpg
 
-    gpg --batch --quiet --verify "${sigFilePath}" "${filePath}"
+    echo "${AWSCLI_GPG_KEY_MATERIAL}" | gpg --dearmor > "${awsGpgKeyring}"
+    gpg --batch --quiet --no-default-keyring --keyring "./${awsGpgKeyring}" --verify "${sigFilePath}" "${filePath}"
```
Rabadash8820, line 98: Nit: I would add `./` to the value of `awsGpgKeyring` so that the references are consistent.

Chuxel:
> @Chuxel Out of curiosity, I did a [quick search](https://github.com/search?q=%22--import%22+repo%3Amicrosoft%2Fvscode-dev-containers&type=Code) for other instances of gpg --import across this repo and got the following hits.

  Generally the way these work is by setting a temp GNUPGHOME and then deleting it afterwards to avoid this problem.  e.g. https://github.com/microsoft/vscode-dev-containers/blob/4a8064f120e1dbe4c3fbc66d9d97ece499e5dbcc/script-library/go-debian.sh#L160

Rabadash8820:
Ah. Would it be better to use the same method here, for consistency? (I kinda like @sktan's method better b/c messing with env variables always feels hacky to me, but I'm indifferent).

PR #1395, opened 2022-04-07
Add swift VARIANT for swift 5.6

adam-fowler:
I am using the `5.6-focal` docker image because it has support for both x86_64 and arm64. While the `5.6` image is based off ubuntu bionic and does not have support for arm64.


0xTim:
I think we should probably comment out the run as non-root section and make it opt in since it seems to be causing issues with SwiftPM

adam-fowler:
> I think we should probably comment out the run as non-root section and make it opt in since it seems to be causing issues with SwiftPM

Can we look at that in another PR. I'm not sure about that change

0xTim:
LGTM

Chuxel:
This LGTM - But IIRC, Ubuntu focal hits libcurl segfaults on Docker Desktop on Apple Silicon macs.  Debian 11/bullseye, Ubuntu bionic and Jammy (22.04 about to come out) do not have this problem.  It's actually listed as a known issue there.  As I understand it, there's a bug with libcurl that is causing it.

PR #1398, opened 2022-04-07
Prep for Node 18 release, Node 12 deprecation

Chuxel:
Latest information seems to indicate 4/19 for the release date of Node 18 and thus deprecation of Node 12.

Given the increase of macOS machines with Apple Silicon hardware, this will also switch the default OS from "buster" (Debian 10) to "bullseye" (Debian 11) to better enable arm64 support.

We can validate and merge once the upstream Node 18 image has been released.

PR #1399, opened 2022-04-07
Prep for Ubuntu 22.04/Jammy release (including C++ definitions)

Chuxel:
Latest info says this will GA 4/21.

Since this will be very new, we can leave the default for `ubuntu` as `focal` for now and update that once we're more confident that most tools have support.


bamurtaugh:
@joshspicer Looks like Ubuntu 22.04 is now available: https://ubuntu.com/download/desktop. Are you alright merging this and coordinating release (perhaps including this and Node PR: https://github.com/microsoft/vscode-dev-containers/pull/1398)?

joshspicer:
@bamurtaugh Yep! Will let the CI run here and then check on it/merge/release later today!

PR #1415, opened 2022-04-22
Use regex to install `liblttng-ust` in Common Debian script

### LINKED ISSUE #1414: Common Debian fails on Ubuntu 22.04
Hey,

I am just testing upgrading my devcontainer to Ubuntu 22.04 and it fails with 

```bash
E: Unable to locate package liblttng-ust0
```

Steps to reproduce are

```bash
docker run -it --rm --env VSCODE_DEV_CONTAINERS_VERSION="v0.231.6" public.ecr.aws/ubuntu/ubuntu:22.04

apt update --assume-yes

apt install --assume-yes curl 

curl https://raw.githubusercontent.com/microsoft/vscode-dev-containers/${VSCODE_DEV_CONTAINERS_VERSION}/script-library/common-debian.sh --output /tmp/common-debian.sh
  
bash /tmp/common-debian.sh "true" "devcontainer" "1000" "1000" "true" "true" "false"
```

https://github.com/microsoft/vscode-dev-containers/blob/v0.231.6/script-library/common-debian.sh#L105

I will submit a PR to fix this.

jacobwoffenden:
Resolves #1414

### DIFF
--- a/script-library/common-debian.sh
+++ b/script-library/common-debian.sh
@@ -102,7 +102,7 @@ if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
         libkrb5-3 \
         libgssapi-krb5-2 \
         libicu[0-9][0-9] \
-        liblttng-ust0 \
+        liblttng-ust[0-9] \
         libstdc++6 \
         zlib1g \
         locales \


jacobwoffenden:
Oops, I can see that it is taken care of in https://github.com/microsoft/vscode-dev-containers/pull/1399

PR #1423, opened 2022-04-28
replace `git.io` links with their actual URLs

SukkaW:
All links on `git.io` will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/

Replace `git.io` links with their actual URL.

### DIFF
--- a/script-library/fish-debian.sh
+++ b/script-library/fish-debian.sh
@@ -83,9 +83,9 @@ fi
 # Install Fisher
 if [ "${INSTALL_FISHER}" = "true" ]; then
     echo "Installing Fisher..."
-    fish -c 'curl -sL https://git.io/fisher | source && fisher install jorgebucaran/fisher'
+    fish -c 'curl -sL https://raw.githubusercontent.com/jorgebucaran/fisher/main/functions/fisher.fish | source && fisher install jorgebucaran/fisher'
     if [ "${USERNAME}" != "root" ]; then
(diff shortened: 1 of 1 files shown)


script-library/fish-debian.sh
```diff
@@ -83,9 +83,9 @@ fi
 # Install Fisher
 if [ "${INSTALL_FISHER}" = "true" ]; then
     echo "Installing Fisher..."
-    fish -c 'curl -sL https://git.io/fisher | source && fisher install jorgebucaran/fisher'
+    fish -c 'curl -sL https://raw.githubusercontent.com/jorgebucaran/fisher/HEAD/functions/fisher.fish | source && fisher install jorgebucaran/fisher'
```
eitsupi, line 86: fisher's READEME uses `main` instead of `HEAD`.
jorgebucaran/fisher#713

eitsupi:
Thanks for catching this!

script-library/fish-debian.sh
SukkaW: I just replace it with the actual URL the original `git.io` redirect to.
I can update the link to the one provided in fisher's README if you wish.

SukkaW:
> @joshspicer if this change sounds alright with you, I'll merge as the links stop redirecting after today.

@bamurtaugh GitHub describes in a way that only git.io links in print documentation and research papers will retain services. Even if they have changed their mind, it is not guaranteed that the `git.io` will not stop providing service in the future.

IMHO, it is better to use the original URL.

joshspicer:
Looks good to me! Thanks @eitsupi / @SukkaW !

PR #1418, opened 2022-04-25
codespaces-linux v2 image

samruddhikhandale:
- [x] Update the oryx image tag consisting of these [proposal changes](https://github.com/github/codespaces/issues/7171#issuecomment-1081080810)
- [x] Remove languages/frameworks from the codespace base image with these [proposal changes](https://github.com/github/codespaces/issues/7171#issuecomment-1081221878)
- [x] Bump up the `codespaces-linux` version to `2.0.0`
- [x] Add python packages. #7205

PR #1424, opened 2022-04-29
Fix installation of common Python packages

jungaretti:
Modifies `$PATH` to prioritize the version of Python that we install to `/opt/python/latest` (`~/.python/latest` is a link) and cleans up the installation of common ML packages. We added these packages in #1418, but they were only installed for the `root` user.

Without modifying `$PATH`, the various Python commands point to different versions and installations:

Invalid Python commands

After modifying `$PATH`, all of the Python commands point to the latest version that we installed:

Valid Python commands

Closes: https://github.com/github/codespaces/issues/7205


containers/codespaces-linux/.devcontainer/library-scripts/jupyterlab-debian.sh
```diff
@@ -58,3 +58,6 @@ if ! jupyter-lab --version > /dev/null ; then
     sudoUserIf pip install jupyterlab=="${VERSION}" --no-cache-dir
   fi
 fi
+
+# Install common Python packages for use with JupyterLab
+sudoUserIf pip install numpy pandas scipy matplotlib seaborn scikit-learn tensorflow keras torch requests
```
eitsupi, line 63: Doesn't installing packages unrelated to JupyterLab confuse users?

containers/codespaces-linux/.devcontainer/library-scripts/python-ml-debian.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
joshspicer, line 1: An exact copy of this will also need to go in the root [script-library](https://github.com/microsoft/vscode-dev-containers/tree/main/script-library).  At release-time, the copy in each individual `containers/` folder is synced with the "original" copy in the root folder.

script-library/jupyterlab-debian.sh
```diff
@@ -43,18 +44,15 @@ sudoUserIf()
   fi
 }
 
-# If we don't yet have Python, install it now.
-if ! python --version > /dev/null ; then
+# Make sure that Python is installed correctly
+if ! $PYTHON --version > /dev/null ; then
```
joshspicer, line 48: nit: wrap in brackets in case the provided variable has a space in it, etc..

```suggestion
if ! "${PYTHON}" --version > /dev/null ; then
```

joshspicer: and in other places

joshspicer (changes requested):
Minor changes where scripts are place to ensure the image builds properly!

Also, can you add some tests "asserts" 
to https://github.com/microsoft/vscode-dev-containers/blob/8e35f9f03c511dc75dfac28da83ed1b8da4e052c/containers/codespaces-linux/test-project/test.sh

This is executed against the built image in the "smoke tests" on the PR and in CI.  

Three tests are expected to fail right now due to long-standing oryx bugs
`💥  Failed tests: oryx-install-dotnet-3.1.0 dotnet-3.1.0-installed oryx-build-jekyll`, but anything you add _should_ of course pass (along with the existing asserts, as it does right now)!

containers/codespaces-linux/test-project/test.sh
```diff
@@ -31,6 +33,18 @@ check "pydocstyle" pydocstyle --version
 check "bandit" bandit --version
 check "virtualenv" virtualenv --version
 
+# Check Python packages
+check "numpy" python -c 'import numpy'
```
jungaretti, line 37: I'm not sure what's wrong here. I've been testing these changes by copying the Dockerfile and scripts to a test codespace and rebuilding its container. Everything works as expected there:

Python test commands work in a codespace

I noticed that one of the earlier checks, `pip3 --version` has different output than it does on my codespace:

Wrong output in smoke test
Correct output in codespace

Python gets confused when `$PATH` isn't set properly. Would the shell running these tests have the same `$PATH` as [we define in `base.Dockerfile`](https://github.com/jungaretti/vscode-dev-containers/blob/28f084fae7252a088ad61cce53304c078ae7d3fe/containers/codespaces-linux/.devcontainer/base.Dockerfile#L37)?

PR #1430, opened 2022-05-03
update r extension that is used

isidorn:
The r extension has changed the id, we need to update in the devcontainer.json

fixes https://github.com/REditorSupport/vscode-R/issues/690#issuecomment-1115831368


joshspicer:
Thanks!

PR #1434, opened 2022-05-05
remove setting npm global location

joshspicer:
Justification from @mylesborins

> Hey All, I noticed a weirdness with how npm is setup in codespaces, specifically that there is a a ~/.npmrc with the contents
prefix=/home/codespace/.npm-global
While this is good for making things a bit more “consistent” across versions of node / npm it does result in making it significantly more complicated to update npm independent of Node.js when installed via nvm since npm i -g npm installs into that prefix rather then the ~/.nvm folder, so the nvm installed version always will trump the customer installed npm version due to the order in $PATH


joshspicer:
Rerunning the test. Haven't been seeing out of space problems

PR #1437, opened 2022-05-05
Refactor Python package installer

jungaretti:
Removes `python-package-debian.sh` in favor of a private Python package installer script for Codespaces.


jungaretti:
Passing smoke test except for the 3 expected failures

PR #1439, opened 2022-05-10
Add option to create JupyterLab config for user

jungaretti:
Adds a new step to the JupyterLab installer script that can create a config file. We use that new step to fix a CORS bug that affects codespaces running JupyterLab.

Closes https://github.com/github/codespaces/issues/7239


joshspicer (approved):
makes sense to me!

PR #1443, opened 2022-05-11
Add inotify-tools to elixir phoenix postgres template

elliotblackburn:
inotify-tools is required in linux environments for live code reloading with Phoenix. Users seeking to use this template to development will be met with a warning when starting a phoenix project for the first time and will need to install it manually. Adding it as a dependency to the dev container would avoid that and help people get going quicker.

I wasn't sure if it would be clearer to put this as a seperate install command, but it felt like it made most sense to put it next to build-essentials. Happy to move it if needed.

ref: https://hexdocs.pm/phoenix/installation.html#inotify-tools-for-linux-users

### DIFF
--- a/containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
+++ b/containers/elixir-phoenix-postgres/.devcontainer/Dockerfile
@@ -44,7 +44,7 @@ RUN apt-get update \
   fi \
   #
   # Install dependencies
-  && apt-get install -y build-essential \
+  && apt-get install -y build-essential inotify-tools \
   #
   # Clean up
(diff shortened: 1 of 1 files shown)


talk2MeGooseman:
> @idyll @talk2MeGooseman As you're listed as contributors for this template, would love to hear your thoughts.

I agree with this change, I have been finding myself adding this to my personal setup so sharing it for everyone seems like a solid change.

bamurtaugh:
Thanks for confirming! I'll merge this in now.

BlueHatbRit:
Thanks both!

PR #1446, opened 2022-05-13
Use Java 17 by default

jdubois:
JDK 17 is the new long-term support release, so it should be used by default.

This PR follows comment https://github.com/microsoft/vscode-remote-try-java/pull/35#discussion_r869186808 from @chrmarti @joshspicer

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -988,7 +988,7 @@
                 "Type": "other",
                 "Other": {
                     "Name": "Eclipse Temurin",
-                    "Version": "1.8.0_332",
+                    "Version": "17.0.3_7",
                     "DownloadUrl": "https://adoptium.net/"
                 }
             }
@@ -998,7 +998,7 @@
                 "Type": "other",
                 "Other": {
                     "Name": "Docker Image: mcr.microsoft.com/vscode/devcontainers/java",
-                    "Version": "11-bullseye",
+                    "Version": "17-bullseye",
                     "DownloadUrl": "https://hub.docker.com/_/openjdk"
                 }
             }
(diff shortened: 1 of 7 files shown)


cgmanifest.json
```diff
@@ -988,7 +988,7 @@
                 "Type": "other",
                 "Other": {
                     "Name": "Eclipse Temurin",
-                    "Version": "1.8.0_332",
+                    "Version": "17.0.3_7",
```
testforstephen, line 991: No need to modify java 1.8 version here.

For JDK 1.8, we recommend to use Eclipse Temurin 1.8. For higher version, we use Microsoft Build of OpenJDK.

cgmanifest.json
```diff
@@ -998,7 +998,7 @@
                 "Type": "other",
                 "Other": {
                     "Name": "Docker Image: mcr.microsoft.com/vscode/devcontainers/java",
-                    "Version": "11-bullseye",
+                    "Version": "17-bullseye",
```
testforstephen, line 1001: @Chuxel Do you know what the version in cgmanifest.json for? Taken as default image version?

cgmanifest.json
jdubois: Thanks, I just reverted this change

cgmanifest.json
bamurtaugh: @joshspicer @chrmarti Do you have insights here?

Chuxel: This is used for registering OSS - it's auto-generate based on the contents of definition-manifest.json files. Changes will just be overwritten.

jdubois: I have removed this changed, as this is useless

testforstephen (approved):
LGTM

cgmanifest.json
testforstephen: LGTM

PR #1447, opened 2022-05-13
Add `customizations` property

bamurtaugh:
Updating to use `customizations`. I just updated one definition to start out, to ensure it looks correct and we like the wording and layout. 

Once we think this looks good, I'll update the PR to update all the devcontainer.json files in this repo.

### DIFF
--- a/containers/alpine/.devcontainer/devcontainer.json
+++ b/containers/alpine/.devcontainer/devcontainer.json
@@ -6,19 +6,6 @@
 		"args": { "VARIANT": "3.15" }
 	},
 
-	// Configure tool-specific properties.
-	"customizations": {
-		// Configure properties specific to VS Code.
-		"vscode": {
-			// Add the IDs of extensions you want installed when the container is created.
-			// Note that some extensions may not work in Alpine Linux. See https://aka.ms/vscode-remote/linux.
-			"extensions": [],
-
-			// Set *default* container specific settings.json values on container create. 
-			"settings": {}
-		}
-	},
-
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [],
 
--- a/containers/alpine/.devcontainer/devcontainer.json
+++ b/containers/alpine/.devcontainer/devcontainer.json
@@ -5,13 +5,19 @@
 		// Update 'VARIANT' to pick an Alpine version: 3.13, 3.14, 3.15
 		"args": { "VARIANT": "3.15" }
 	},
-	
-	// Set *default* container specific settings.json values on container create. 
-	"settings": {},
 
-	// Add the IDs of extensions you want installed when the container is created.
-	// Note that some extensions may not work in Alpine Linux. See https://aka.ms/vscode-remote/linux.
-	"extensions": [],
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+			// Add the IDs of extensions you want installed when the container is created.
+			// Note that some extensions may not work in Alpine Linux. See https://aka.ms/vscode-remote/linux.
+			"extensions": [],
+
+			// Set *default* container specific settings.json values on container create. 
+			"settings": {}
+		}
+	},
 
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [],
--- a/containers/azure-static-web-apps/.devcontainer/devcontainer.json
+++ b/containers/azure-static-web-apps/.devcontainer/devcontainer.json
@@ -11,17 +11,23 @@
 	},
 	"forwardPorts": [ 7071, 4280 ],
 
-	// Set *default* container specific settings.json values on container create.
-	"settings": {},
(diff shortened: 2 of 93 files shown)


containers/alpine/.devcontainer/devcontainer.json
```diff
@@ -5,13 +5,19 @@
...
 
-	// Add the IDs of extensions you want installed when the container is created.
-	// Note that some extensions may not work in Alpine Linux. See https://aka.ms/vscode-remote/linux.
-	"extensions": [],
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+			// Add the IDs of extensions you want installed when the container is created.
+			// Note that some extensions may not work in Alpine Linux. See https://aka.ms/vscode-remote/linux.
```
chrmarti, line 14: Assuming the note on Alpine only goes into Alpine templates.

containers/alpine/.devcontainer/devcontainer.json
```diff
@@ -5,13 +5,19 @@
...
-	// Set *default* container specific settings.json values on container create. 
-	"settings": {},
 
-	// Add the IDs of extensions you want installed when the container is created.
-	// Note that some extensions may not work in Alpine Linux. See https://aka.ms/vscode-remote/linux.
-	"extensions": [],
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
```
chrmarti, line 12: Since "extensions" and "settings" are empty: Should we omit "vscode" (and "customizations")? There is a discoverability argument against the removal, but at the same time this approach won't scale to more tools.

containers/alpine/.devcontainer/devcontainer.json
bamurtaugh: Yes! Good call out though. I copied over the note on Alpine from the existing devcontainer.json.

containers/alpine/.devcontainer/devcontainer.json
bamurtaugh: This is a good point. Looking at several of the other devcontainer.json's in this repo, they all seem to contain `extensions` and `settings`, even if they're empty. But, this may have been because we considered these standard properties that everyone may want to use, which is less the case now (they're editor-specific). 

I think it makes sense to remove them from configurations where they're empty. In this next pass, I'll do that, plus the remaining `customizations` work for devcontainer.json's where they aren't empty.

bamurtaugh:
@chrmarti Took a pass at updating the remaining devcontainer.json's, let me know what you think.

containers/azure-static-web-apps/.devcontainer/devcontainer.json
```diff
@@ -11,17 +11,20 @@
...
-		"ms-python.python",
-		"dbaeumer.vscode-eslint"
-	],
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+			// Add the IDs of extensions you want installed when the container is created.
+			"extensions": [
+			"ms-azuretools.vscode-azurefunctions",
```
chrmarti, line 20: Nit: Extension ids might be missing one indentation level.

containers/azure-terraform/.devcontainer/devcontainer.json
```diff
@@ -25,22 +25,29 @@
...
-		"ms-azuretools.vscode-azureterraform",
-		"ms-azuretools.vscode-docker"
-	],
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+				// Set *default* container specific settings.json values on container create.
+				"settings": { 
+				"terraform.languageServer": {
```
chrmarti, line 35: Indentation level? (Or could be displayed incorrectly in the GitHub UI.)

chrmarti (approved):
Thanks!

containers/azure-static-web-apps/.devcontainer/devcontainer.json
bamurtaugh: Good catch, fixed!

bamurtaugh:
@joshspicer Any idea why the `codespace-linux` smoke test is failing? Error is: `💥  Failed tests: oryx-install-dotnet-3.1.0 dotnet-3.1.0-installed oryx-build-jekyll`

joshspicer:
> @joshspicer Any idea why the `codespace-linux` smoke test is failing? Error is: `💥 Failed tests: oryx-install-dotnet-3.1.0 dotnet-3.1.0-installed oryx-build-jekyll`

Those three are expected right now 🙃.  It's a problem with ownership in the oryx image vs our image and how we run these tests. Safe to ignore if it's only these three.  

(I keep them on so that I can double check the logs and see that it is indeed the same issue.  We have an open issue with oryx to fix this, but we're moving off oryx upstream images altogether soon so this will be a non-issue moving forward).

tuanggolt (commented):
open

PR #1465, opened 2022-05-23
Rust: Target Correct Rust Analyzer Extension

SE2Dev:
The `matklad.rust-analyzer` extension has been replaced in favor of `rust-lang.rust-analyzer`. Since the old extension no longer exists, I've updated the referenced extension in the dev container configurations.


joshspicer:
Awesome, thanks!

PR #1475, opened 2022-05-26
[R] Update info about codespaces

eitsupi:
It didn't seem to work well with the previous R definition (which used a different extension), but now it seems to work on Codespaces.


bamurtaugh:
@joshspicer @edgonmsft anything Codespaces-specific we should be aware of with this template? Or should this be okay to merge?

eitsupi:
@joshspicer @edgonmsft Any thoughts?

joshspicer:
Don't have a great idea how to test this, but if it is working - great!

Issue #1493, opened 2022-06-05
Devcontainer does not update remote user's UID to match host machine's UID when group already exists


KyleRAnderson:
- VSCode Version: 1.67.2 (user setup)
- Local OS Version: Windows_NT x64 10.0.22000, but running over Remote - SSH on an Ubuntu 20.04 host.
- Local chip architecture: x86
- Reproduces in: Remote - Containers
- Name of Dev Container Definition with Issue: Update UID feature applied to any devcontainer. In this case, I was using the C++ devcontainer.

Steps to Reproduce:

1. On the container host, create a user with a UID of something other than the default 1000 (I used 1577 in my case) and a GID of 100.
2. Create a C++ VSCode devcontainer definition with no modifications and launch the container. I tried this using the ubuntu-20.04 , ubuntu-21.04 and debian-11 variants, it doesn't seem to matter.
3. Check whether the remote user's UID has been updated to 1577. In my case, it had not been updated, causing permission issues when writing files in the bind mount from the host system.

I did a little bit of investigation into why this is the case by looking at the Dockerfile that the UID update feature uses. It looks like what is happening is that a group with ID 100 already exists in the container so the feature skips updating both the UID and GID. However, I do not understand why this logic is in place as we would still want to update the UID in this case, and I see no harm as well in assigning the vscode user to the existing GID.

joshspicer:
Perhaps @Chuxel can weigh in on this design decision.  Just want to make sure - is this purely based on curiosity, or is the current logic we have here causing a problem of some kind?

Chuxel:
This sounds like a dupe of https://github.com/microsoft/vscode-remote-release/issues/2402.  This is not about the images in this repo, but rather what is possible when doing UID/GID syncing.

The core challenge is that it's a bit unclear as to what is safe to do in this scenario - low group IDs like that are often system groups, so reusing likely isn't safe to do. I'd suggest moving the conversation to this existing issue.  However, 100 is "users" so in concept that could be made an exception.

Closing as a dupe, more discussion can happen over in the other issue.

PR #1500, opened 2022-06-13
fix git-lfs install from git source 

### LINKED ISSUE #1473: Git LFS feature fails to build on Linux/ARM64 for version 3.2+
Steps to Reproduce:

1. Add `"git-lfs": "latest"` to the `features` definition
2. Build for `linux/arm64` using the `devcontainer` cli

Error message:
```
[19231 ms] ------
 > [stage-0 3/3] RUN cd /tmp/build-features/local-cache && chmod +x ./install.sh && ./install.sh:
...
#13 13.36 git-lfs-linux-arm64-v3.2.0.tar.gz: OK
#13 13.72 ./git-lfs-debian.sh: line 163: ./install.sh: No such file or directory
```

The underlying cause is that the tarball structure changed, everything is now in a subdirectory with a name that I _think_ changes based on the version (haven't checked).

If anyone comes across this and needs to work around it, use version 3.1.4:
`"features": { "git-lfs": "3.1.4" }`

joshspicer:
closes #1473, which reported a change in the underlying file structure of the release artifacts. This manifested on "jammy", since git-lfs is not published in the jammy apt repo yet.

### DIFF
--- a/script-library/git-lfs-debian.sh
+++ b/script-library/git-lfs-debian.sh
@@ -140,6 +140,7 @@ install_using_apt() {
 
     if ! (apt-get update && apt-get install -yq git-lfs${version_suffix}); then
         rm -f /etc/apt/sources.list.d/git-lfs.list
+        echo "Could not fetch git-lfs from apt"
         return 1
     fi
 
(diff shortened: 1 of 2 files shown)


joshuarrolfwalz:
#1493

Issue #1501, opened 2022-06-13
'fish' script `"(line 197): 'case' builtin not inside of switch block"`


joshspicer:
Running script-library regressions tests, I noticed https://github.com/microsoft/vscode-dev-containers/runs/6867998358?check_suite_focus=true#step:4:19235

<img width="1422" alt="image" src="https://user-images.githubusercontent.com/23246594/173430210-45fa4b70-ac89-4c5b-8bdf-711358f0984d.png">

Line that exercises fish: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/test/regression/run-scripts.sh#L112

Note: PR #1500 temporarily disables this check: https://github.com/microsoft/vscode-dev-containers/pull/1500/files#diff-b4fcdb4d2b5b4dd46c896cb5bfa0a42e633340c263aa0238d71dce261bbd5c36R112

PR #1503, opened 2022-06-14
Fix Fish shell installation

andreiborisov:
Fixes https://github.com/microsoft/vscode-dev-containers/issues/1501

### DIFF
--- a/script-library/fish-debian.sh
+++ b/script-library/fish-debian.sh
@@ -65,7 +65,8 @@ echo "Installing fish shell..."
 if grep -q 'Ubuntu' < /etc/os-release; then
     check_packages software-properties-common
     apt-add-repository -y ppa:fish-shell/release-3
-    check_packages fish
+    apt-get update
+    apt-get -y install --no-install-recommends fish
 elif grep -q 'Debian' < /etc/os-release; then
     if grep -q 'stretch' < /etc/os-release; then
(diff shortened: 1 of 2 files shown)


eitsupi:
Does this change fix #1501?
[The log](https://github.com/microsoft/vscode-dev-containers/runs/6867998358?check_suite_focus=true#step:4:19235) show that the fish installation was not successful?

eitsupi:
I think the fish installation was successful because I see `Installing Fisher...` in the log.
What is failing is the installation of Fisher.

https://github.com/microsoft/vscode-dev-containers/blob/172a918f40e31bd24da8e64135026ec9f26c91b0/script-library/fish-debian.sh#L85-L88

image

andreiborisov:
@eitsupi the issues was `check_packages fish` not working properly and as a result the script installing ancient fish version from the system package list. It has a slightly different syntax, hence `fisher` fails to work properly since it only supports fish v3 and greater.

tldr: it should fix it.

eitsupi:
That makes sense. Thank you for working on it!

This change should be revert in this PR, I think.
https://github.com/microsoft/vscode-dev-containers/pull/1500/files#diff-b4fcdb4d2b5b4dd46c896cb5bfa0a42e633340c263aa0238d71dce261bbd5c36R112

And version check e.x. `fish -v` in this script maybe useful for debug?

andreiborisov:
@eitsupi good suggestion! Added version check.

andreiborisov:
@joshspicer seem to be good for the merge🙌🏻

joshspicer:
Thank you guys for your continued support! 🚀

Issue #1490, opened 2022-06-03
Python + Homebrew, returning "Error: Running Homebrew as root is extremely ..."


mozgbrasil:
- VSCode Version: 1.67.2
- Local OS Version: Ubuntu 22.04 LTS
- Local chip architecture: x86-64
- Reproduces in: Remote - Containers
- Name of Dev Container Definition with Issue: "Python 3"

Steps to Reproduce:

Good morning my friends

About: Python + Homebrew, returning error "Error: Running Homebrew as root is extremely ..."

I have just created a development environment using python + homebrew and the following error is returned

```
Installing Homebrew...
Cloning into '/home/linuxbrew/.linuxbrew/Homebrew'...
Cloning into '/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core'...
Error: Running Homebrew as root is extremely dangerous and no longer supported.
As Homebrew does not drop privileges on installation you would be giving all
build scripts full access to your system.
The command '/bin/sh -c cd /tmp/build-features/local-cache && chmod +x ./install.sh && ./install.sh' returned a non-zero code: 1
[395831 ms] Error: Command failed: docker build -t vsc-python-labs-55ab636a2bbdda6b755a2f7493ddd6d9-features --build-arg BASE_IMAGE=vsc-python-labs-55ab636a2bbdda6b755a2f7493ddd6d9 --build-arg IMAGE_USER=root /tmp/vsch/container-features/0.234.0-1654274760079
[395832 ms]     at IR (/home/marcio/.vscode/extensions/ms-vscode-remote.remote-containers-0.234.0/dist/spec-node/devContainersSpecCLI.js:221:995)
[395832 ms]     at zy (/home/marcio/.vscode/extensions/ms-vscode-remote.remote-containers-0.234.0/dist/spec-node/devContainersSpecCLI.js:221:933)
[395832 ms]     at processTicksAndRejections (node:internal/process/task_queues:96:5)
[395833 ms]     at async $R (/home/marcio/.vscode/extensions/ms-vscode-remote.remote-containers-0.234.0/dist/spec-node/devContainersSpecCLI.js:226:2007)
[395833 ms]     at async Zy (/home/marcio/.vscode/extensions/ms-vscode-remote.remote-containers-0.234.0/dist/spec-node/devContainersSpecCLI.js:226:3112)
[...]

PR #1504, opened 2022-06-15
Make sure to run Homebrew as a non-root user only

andreiborisov:
Fixes: https://github.com/microsoft/vscode-dev-containers/issues/1490

### DIFF
--- a/script-library/homebrew-debian.sh
+++ b/script-library/homebrew-debian.sh
@@ -39,6 +39,11 @@ elif [ "${USERNAME}" = "none" ] || ! id -u ${USERNAME} > /dev/null 2>&1; then
     USERNAME=root
 fi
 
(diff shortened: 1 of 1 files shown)


bamurtaugh:
Thanks @andreiborisov!

@joshspicer it looks like some tests didn't pass this PR - is that a concern?

andreiborisov:
I need to look into this again, it seems like there is an issue upstream that need to be fixed first

joshspicer:
Thanks, @andreiborisov!

andreiborisov:
@Michaelvsk we migrated Homebrew installation script on the Devcontainer Features spec. Please, use the new feature: https://github.com/meaningful-ooo/devcontainer-features/tree/main/src/homebrew

PR #1505, opened 2022-06-16
Add `util-linux` to `common-alpine.sh`

felipecrs:
This package comes preinstalled in Debian, and provides some very common commands on Linux. Namely, I miss the `script` command.

The list of commands can be seemed here:

https://www.mankier.com/package/util-linux


ryboe:
I built this and the `linux/amd64` image size goes from 110 MB to 117 MB. That 7 MB buys us a long list of useful commands. However, a lot of those commands, like `fstab`, `mkfs`, `cfdisk`, and `eject`, aren't relevant to codespaces 🤔 It's a tough call.

bamurtaugh:
Thanks for opening!

PR #1529, opened 2022-07-11
add terraform-docs to azure-terraform

t3mi:
Adds terraform-docs to azure-terraform container


bamurtaugh:
cc @cmendible for thoughts.

PR #1543, opened 2022-07-19
Add Alpine 3.16

ryboe:
Alpine 3.16 was released on 2022-05-23 and is currently at 3.16.1. The new version has some updated stable packages. In particular, I'm interested in the updated `fzf` package, which does not depend on `perl`.

UPDATE: I release now that I need to generate release notes in `alpine/history`. Is there a tool I should use to create `history/0.204.5.md`?

### DIFF
--- a/containers/alpine/.devcontainer/Dockerfile
+++ b/containers/alpine/.devcontainer/Dockerfile
@@ -1,5 +1,5 @@
-# [Choice] Alpine version: 3.15, 3.14, 3.13
-ARG VARIANT=3.15
+# [Choice] Alpine version: 3.16, 3.15, 3.14, 3.13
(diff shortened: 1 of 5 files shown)


bamurtaugh:
This looks reasonable to me and is in-line with our general process for new software versions, @Chuxel seem reasonable to you?

Chuxel:
Everything looks right here from my perspective. This came out faster than we'd expected. @joshspicer ?

> UPDATE: I release now that I need to generate release notes in alpine/history. Is there a tool I should use to create history/0.204.5.md?

The release process will automatically generate this file, so no need to do anything here.

ryboe:
Thanks! I tried to follow #1432 as closely as possible 🙁

PR #1544, opened 2022-07-20
correct systemctl path

maksimryndin:
On debian system path is different

image


Chuxel:
Ack.  Good catch.  Thanks!

PR #1546, opened 2022-07-21
Update Pylint version

bamurtaugh:
Update to latest version 2.14.5.

@Chuxel I only found mention of a specific version of Pylint in the cgmanifest - please let me know if I might've missed anything else I should update too.

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -204,7 +204,7 @@
                 "Type": "Pip",
                 "Pip": {
                     "Name": "pylint",
-                    "Version": "2.14.3"
+                    "Version": "2.14.5"
                 }
             }
         },
@@ -776,15 +776,6 @@
                     "Name": "Docker Image: php",
                     "Version": "7.4-apache-buster",
                     "DownloadUrl": "https://hub.docker.com/_/php"
-                }
-            }
-        },
-        {
-            "Component": {
-                "Type": "Pip",
-                "Pip": {
-                    "Name": "pylint",
-                    "Version": "2.13.9"
                 }
             }
         },


Chuxel:
@bamurtaugh - This is auto-generated when the images are published so if we want to update the version, we'd need to do another publish.

This query will tell you which python version is getting registered.

https://github.com/search?q=pylint+repo%3Amicrosoft%2Fvscode-dev-containers+filename%3Adev.md+language%3AMarkdown+language%3AMarkdown&type=Code&ref=advsearch&l=Markdown&l=Markdown

Chuxel:
Actually - looks like the kitchen sink image is failing to build which is preventing dev from updating.  @joshspicer - any idea what is going on?

https://github.com/microsoft/vscode-dev-containers/runs/7392375820?check_suite_focus=true

This is set up to "fail fast" - we could have it set up to keep processing anyway, but  the codespaces image wouldn't be accurate.

bamurtaugh:
@Chuxel I see this in the search:

image

So it sounds like we should be looking for what version of Python is being used in definitions (and what version of Pylint comes with those Python versions), rather than explicitly looking for Pylint?

Chuxel:
> @Chuxel I see this in the search:
> 
> image
> 
> So it sounds like we should be looking for what version of Python is being used in definitions (and what version of Pylint comes with those Python versions), rather than explicitly looking for Pylint?

Yep - we chatted about this offline, but there's a new search beta so it's possible you and I are on different engines which is why the query is slightly different. 😵‍💫

bamurtaugh:
With https://github.com/microsoft/vscode-dev-containers/pull/1549, we shouldn't need this manual PR to update the Pylint versions now - they should automatically update with the Codespaces fix.

I just checked https://github.com/microsoft/vscode-dev-containers/blob/main/cgmanifest.json and still see the older versions of Pylint - I'm guessing it may just take some time for the Codespaces change to take effect. I'll leave this PR open until I see the change.

Chuxel:
The update missed the weekly monday morning dev build so I retriggered. Will take a few hours to get through it all.

bamurtaugh:
Looks like the version of Pylint has automatically updated to 2.14.5 now ([ref](https://github.com/microsoft/vscode-dev-containers/commit/7a617791d1a00631ba1dc879acf77861fd061855#diff-45a0a38d395bf4427d0291d9c72bd99b57fa43e14b00e059ae170d657949f1f2R207)), so I'll close this out.

PR #1552, opened 2022-07-25
codespaces-linux - override python

joshspicer:


joshspicer:
continuing in https://github.com/microsoft/vscode-dev-containers/pull/1553

PR #1553, opened 2022-07-26
Reinstall Python to fix missing lzma module

jungaretti:
Fixes: https://github.com/github/codespaces/issues/8155

Removes Oryx's Python installation and replaces it with a fresh installation that is built from source. The fresh installation includes the `lzma` library that is required by many other Python modules.

Preserves the `/home/codespace/.python/current` -> `/opt/python/3.10.4` link that Oryx created and that exists in our published image. This PR is a drop-in replacement and will not break any scripts that rely on `/home/codespace/.python/current/bin/python`.

### Changes

- Replaced Oryx's Python installation with a fresh build from source
- Added new test for `lzma`
- Disabled Oryx tests that always fail

### DIFF
--- a/.github/workflows/smoke-codespace.yaml
+++ b/.github/workflows/smoke-codespace.yaml
@@ -19,13 +19,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Maximize build space
-      uses: easimon/maximize-build-space@master
-      with:
-        remove-dotnet: 'true'
-        remove-android: 'true'
-        remove-haskell: 'true'
-
     - name: Checkout
       id: checkout
       uses: actions/checkout@v2
--- a/.github/workflows/smoke-codespace.yaml
+++ b/.github/workflows/smoke-codespace.yaml
@@ -19,6 +19,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
+    - name: Maximize build space
+      uses: easimon/maximize-build-space@master
+      with:
+        remove-dotnet: 'true'
+        remove-android: 'true'
+        remove-haskell: 'true'
+
     - name: Checkout
       id: checkout
       uses: actions/checkout@v2
--- a/containers/codespaces-linux/.devcontainer/base.Dockerfile
+++ b/containers/codespaces-linux/.devcontainer/base.Dockerfile
@@ -68,8 +68,12 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && mkdir -p /usr/local/etc/vscode-dev-containers/ \
     && mv -f /tmp/scripts/first-run-notice.txt /usr/local/etc/vscode-dev-containers/
 
+# Remove all existing Python installations
+RUN rm -rf /opt/python
+
 # Install Python, JupyterLab, common machine learning packages, and Ruby utilities
-RUN bash /tmp/scripts/python-debian.sh "none" "/opt/python/latest" "${PIPX_HOME}" "${USERNAME}" "true" \
+# Image includes a link from /home/codespace/.python/current -> /opt/python/3.10.4
+RUN bash /tmp/scripts/python-debian.sh "3.10.4" "/opt/python/3.10.4" "${PIPX_HOME}" "${USERNAME}" "true" "true" "false" \
     # Install JupyterLab and common machine learning packages
     && PYTHON_BINARY="${PYTHON_ROOT}/current/bin/python" \
     && bash /tmp/scripts/jupyterlab-debian.sh "latest" "automatic" ${PYTHON_BINARY} "true" \
--- a/containers/codespaces-linux/.devcontainer/base.Dockerfile
+++ b/containers/codespaces-linux/.devcontainer/base.Dockerfile
@@ -68,10 +68,12 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && mkdir -p /usr/local/etc/vscode-dev-containers/ \
     && mv -f /tmp/scripts/first-run-notice.txt /usr/local/etc/vscode-dev-containers/
 
+# Remove existing Python installation from Oryx base image
+RUN rm -rf /opt/python && rm "${PYTHON_ROOT}/current"
+
 # Install Python, JupyterLab, common machine learning packages, and Ruby utilities
-RUN rm -rf /opt/python \
-    # Replace existing Python installation from Oryx image with a fresh installation
-    && bash /tmp/scripts/python-debian.sh "3.10.4" "/opt/python/3.10.4" "${PIPX_HOME}" "${USERNAME}" "true" "true" "false" \
+RUN bash /tmp/scripts/python-debian.sh "3.10.4" "/opt/python/3.10.4" "${PIPX_HOME}" "${USERNAME}" "true" "true" "false" \
+    # Preserve symbolic link from Oryx base image
     && ln -sf /opt/python/3.10.4 "${PYTHON_ROOT}/current" \
     # Install JupyterLab and common machine learning packages
     && PYTHON_BINARY="${PYTHON_ROOT}/current/bin/python" \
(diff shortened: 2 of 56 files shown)


jungaretti:
The Oryx image includes a link from `/home/codespace/.python/current` to `/opt/python/3.10.4`. We need to remove/update that link if we want to support versions beyond `3.10.4`.

jungaretti:
We should add a new test for `lzma`

jungaretti:
I'm not sure why this is failing the smoke test. I tested the new Dockerfile in a codespace using this branch: https://github.com/jungaretti/playground/tree/lzma-testing

I copied the `~/.devcontainer` directory from this PR's branch, tweaked paths, and rebuilt my container.

No error in codespace

containers/codespaces-linux/.devcontainer/library-scripts/python-debian.sh
```diff
@@ -215,9 +215,9 @@ install_from_source() {
     fi
     echo "(*) Building Python ${PYTHON_VERSION} from source..."
     # Install prereqs if missing
-    check_packages curl ca-certificates gnupg2 tar make gcc libssl-dev zlib1g-dev libncurses5-dev \
-                libbz2-dev libreadline-dev libxml2-dev xz-utils libgdbm-dev tk-dev dirmngr \
-                libxmlsec1-dev libsqlite3-dev libffi-dev liblzma-dev uuid-dev 
+    check_packages curl gdb ca-certificates gnupg2 tar make gcc libssl-dev zlib1g-dev libncurses5-dev \
+                libbz2-dev libreadline-dev libreadline6-dev libxml2-dev xz-utils libgdbm-dev libgdbm-compat-dev tk-dev dirmngr \
+                libxmlsec1-dev libsqlite3-dev libffi-dev liblzma-dev lzma zma-dev uuid-dev
```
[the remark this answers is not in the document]
joshspicer: I added in packages after reading https://devguide.python.org/getting-started/setup-building/index.html#linux

containers/codespaces-linux/.devcontainer/base.Dockerfile
```diff
@@ -68,8 +68,13 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
...
     && mv -f /tmp/scripts/first-run-notice.txt /usr/local/etc/vscode-dev-containers/
 
+# Remove existing Python installation from Oryx base image
+RUN rm -rf /opt/python && rm "${PYTHON_ROOT}/current"
+
 # Install Python, JupyterLab, common machine learning packages, and Ruby utilities
-RUN bash /tmp/scripts/python-debian.sh "none" "/opt/python/latest" "${PIPX_HOME}" "${USERNAME}" "true" \
+RUN bash /tmp/scripts/python-debian.sh "3.10.4" "/opt/python/3.10.4" "${PIPX_HOME}" "${USERNAME}" "true" "true" "false" \
+    # Preserve symbolic link from Oryx base image
+    && ln -sf /opt/python/3.10.4 "${PYTHON_ROOT}/current" \
```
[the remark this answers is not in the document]
jungaretti: > We should also install `3.9.7` to keep the current kitchensink as is.

Good idea. I'll add a second call to the installer script to install `3.9.7` to `/opt/python/3.9.7`.

> It might be nicer to add the addition sym links

How does the new feature handle symlinks?

containers/codespaces-linux/.devcontainer/base.Dockerfile
```diff
@@ -68,8 +68,13 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && mkdir -p /usr/local/etc/vscode-dev-containers/ \
     && mv -f /tmp/scripts/first-run-notice.txt /usr/local/etc/vscode-dev-containers/
 
+# Remove existing Python installation from Oryx base image
+RUN rm -rf /opt/python && rm "${PYTHON_ROOT}/current"
+
 # Install Python, JupyterLab, common machine learning packages, and Ruby utilities
-RUN bash /tmp/scripts/python-debian.sh "none" "/opt/python/latest" "${PIPX_HOME}" "${USERNAME}" "true" \
+RUN bash /tmp/scripts/python-debian.sh "3.10.4" "/opt/python/3.10.4" "${PIPX_HOME}" "${USERNAME}" "true" "true" "true" \
```
jungaretti, line 75:
```suggestion
RUN bash /tmp/scripts/python-debian.sh "3.10.4" "/opt/python/3.10.4" "${PIPX_HOME}" "${USERNAME}" "true" "true" "false" \
```

Installing with Oryx reintroduces the `lzma` bug 😞:

ModuleNotFoundError: No module named '_lzma'

We need to build Python from source to fix `lzma`

containers/codespaces-linux/.devcontainer/base.Dockerfile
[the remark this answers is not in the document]
samruddhikhandale: > How does the new feature handle symlinks?

haha. It doesn't add `3 3.10 3.9 etc` but again that's a new way of how kitchensink is built with features.

jungaretti:
Building Python from source adds a new `/usr/local/py-utils` directory that is about 194 MB. In total, this PR adds about 300 MB of disk usage.

containers/codespaces-linux/test-project/test.sh
```diff
@@ -3,6 +3,11 @@ cd $(dirname "$0")
 
 source test-utils.sh codespace
 
+echo '---'
+which python
+echo $PATH
+echo '---'
+
```
jungaretti, line 10: Should we keep this?

.github/workflows/smoke-codespace.yaml
```diff
@@ -19,6 +19,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
+    - name: Maximize build space
+      uses: easimon/maximize-build-space@master
+      with:
+        remove-dotnet: 'true'
+        remove-android: 'true'
+        remove-haskell: 'true'
+
```
jungaretti, line 28: Should we keep this?

containers/codespaces-linux/test-project/test.sh
```diff
@@ -33,6 +38,31 @@ check "pydocstyle" pydocstyle --version
...
+    # Skip directories
+    if [[ -d "$python" ]]; then
+        continue
+    fi
+    if "$python" -c 'import lzma' &>/dev/null; then
+        echo "$python: true"
+    else
+        echo "$python: false"
+    fi
+done
```
jungaretti, line 61: We can remove this

containers/codespaces-linux/test-project/test.sh
```diff
@@ -33,6 +38,31 @@ check "pydocstyle" pydocstyle --version
...
+
+check "lzma cpython test" python /opt/python/3.10.4/lib/python3.10/test/test_lzma.py
+check "lzma library included" python -c "import lzma"
```
jungaretti, line 64: Should we keep these? They always seem to fail, but I've tested several times in codespaces and Python is working as expected.

samruddhikhandale:
```suggestion
check "lzma" python -c "import lzma"
```

I am guessing if the `check` function is failing due to spaces 🤔  I don't see any other `check` arg with spaces

.github/workflows/smoke-codespace.yaml
samruddhikhandale: If this isn't helping with the test failures, then it would be nice to remove.

containers/codespaces-linux/test-project/test.sh
samruddhikhandale: Remove? unless these logging statements are helpful

samruddhikhandale:
> Building Python from source adds a new /usr/local/py-utils directory that is about 194 MB. 

@jungaretti  ℹ️  We already have `/usr/local/py-utils` in the current kitchensink

image

samruddhikhandale (approved):
Approved with some suggestions! ✨

containers/codespaces-linux/test-project/test.sh
jungaretti: Worth a shot!

PR #1557, opened 2022-07-28
Update mongo tools install for latest available, add bullseye

Chuxel:
The mongo database tools were originally not available for Debian Bullseye when this definition was last edited.  That has since changed.

In addition, arm64 support is not available for mongodb-database-tools but mongodb-mongosh is available. So this ensures the shell is always installed.


Chuxel:
//cc: @dbaeumer @burkeholland @gregvanl

edgonmsft (approved):
LGTM

PR #1573, opened 2022-08-10
Add info on using ingress in Codespaces

Chuxel:
Documents how to work around a Codespaces shortcoming when setting up ingress for Kubernetes. 

See https://twitter.com/Chuxel/status/1557151634593038339

//cc @jkeech @joshspicer

### DIFF
--- a/script-library/docs/kubectl-helm.md
+++ b/script-library/docs/kubectl-helm.md
@@ -38,7 +38,7 @@ Or as a feature:
 
 ## A note on ingress and port forwarding
 
-When configuring ingress for your Kubernetes cluster, note that by default Kubernetes will bind to a specific interface's IP rather than localhost or all interfaces. This is why you need to use the Kubernetes Node's IP when connecting - even if there's only one Node as in the case of Minikube. Port forwarding in Remote - Containers will allow you to specify `<ip>:<port>` in either the `forwardPorts` property or through the port forwarding UI in VS Code.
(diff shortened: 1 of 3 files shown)


script-library/docs/kubectl-helm.md
```diff
@@ -36,6 +36,16 @@ Or as a feature:
 |minikube SHA256| | `automatic`| SHA256 checksum to use to verify the minikube download. `automatic` will download the checksum. |
 |Non-root user| | `automatic`| Specifies a user in the container other than root that will be using the desktop. A value of `automatic` will cause the script to check for a user called `vscode`, then `node`, `codespace`, and finally a user with a UID of `1000` before falling back to `root`. |
 
+## A note on ingress and port forwarding
+
+When configuring ingress for your Kubernetes cluster, note that by default Kubernetes will bind to a specific interface's IP rather than localhost or all interfaces. This is why you need to use the Kubernetes Node's IP when connecting - even if there's only one Node as in the case of Minikube. Port forwarding in Remote - Containers will allow you to specify `<ip>:<port>` in either the `forwardPorts` property or through the port forwarding UI in VS Code.
```
bamurtaugh, line 41:
```suggestion
When configuring [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) for your Kubernetes cluster, note that by default Kubernetes will bind to a specific interface's IP rather than localhost or all interfaces. This is why you need to use the Kubernetes Node's IP when connecting - even if there's only one Node as in the case of Minikube. Port forwarding in Remote - Containers will allow you to specify `<ip>:<port>` in either the `forwardPorts` property or through the port forwarding UI in VS Code.
```

bamurtaugh (approved):
Thanks for making these updates @Chuxel! One suggestion to add a docs link.

Chuxel:
Thanks!

PR #1554, opened 2022-07-26
Option to specify nightly version of Rust

maksimryndin:

PR #1592, opened 2022-08-22
rust-debian.sh: fix set -e interaction with nightly rust detection

tomaszjonak:
Currently feature(rust) in devcontainer definition fails
container build unless user specifies nightly rust.

Issue was introduced in commit 10bf6f70c4f467be58dee5df8a7ff19cc598e09f
https://github.com/microsoft/vscode-dev-containers/pull/1554

Lack of "nightly" in RUST_VERSION will result in nonzero return code
from grep. This in turn triggers -e option set for the whole script.
As a result build silently fails.

### DIFF
--- a/script-library/rust-debian.sh
+++ b/script-library/rust-debian.sh
@@ -183,8 +183,8 @@ else
             apt_get_update_if_needed
             apt-get -y install --no-install-recommends git
         fi
-        echo ${RUST_VERSION} | grep -q "nightly"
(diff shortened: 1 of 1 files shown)


tomaszjonak:
cc [joshspicer](https://github.com/joshspicer) as an approver of original commit.

joshspicer:
Thanks for this!

joshspicer (approved):

PR #1526, opened 2022-07-11
Add bun devcontainer

bherbruck:

PR #1614, opened 2022-09-04
Add bun container definition

Wobbly-Wibbly:
Adds the possibility to play with a brand new JS/TS runtime [Bun](https://bun.sh/)


eitsupi:
duplicate of #1526

PR #1623, opened 2022-09-09
avsnr.wrong.account.py

avsnr:
You only need ID + token  My goal is to publish unencrypted information, it's for education and if you're smart about it, you'll use it.  Developer: @avsnr


jkeech:
Closing as off-topic. This is not related to dev containers or setting up a dev environment.

PR #1626, opened 2022-09-13
Add `app.github.dev` to `RAILS_DEVELOPMENT_HOSTS`

joshaber:
Codespaces is migrating port forwarding to live under `app.github.dev` so we need to update the default `RAILS_DEVELOPMENT_HOSTS` to reflect that.


containers/codespaces-linux/.devcontainer/base.Dockerfile
```diff
@@ -27,7 +27,7 @@ ENV SHELL=/bin/bash \
     PIPX_HOME="/usr/local/py-utils" \
     PIPX_BIN_DIR="/usr/local/py-utils/bin" \
     RVM_PATH="/usr/local/rvm" \
-    RAILS_DEVELOPMENT_HOSTS=".githubpreview.dev" \ 
+    RAILS_DEVELOPMENT_HOSTS=".githubpreview.dev,.app.github.dev" \
```
samruddhikhandale, line 30: Added to `universal` image in devcontainers/images https://github.com/devcontainers/images/pull/112

samruddhikhandale: We no longer build images from this repo, so this change would not have any effect. It will be nice if we could revert changes in this file 😄

joshaber: Done!

samruddhikhandale (approved):
LGTM ✨

ghiculescu:
@joshaber I thought you might want to know that this out of the box config doesn't seem to be working on new codespaces on the current Rails release. I think it has not worked since https://github.com/rails/rails/commit/0fccfb9a3097a9c4260c791f1a40b128517e7815 though it may be earlier.

Here is a test that shows the issue: https://github.com/rails/rails/compare/main...ghiculescu:rails:codespaces-bug

The `[".githubpreview.dev", ".app.github.dev"]` comes from [the Dockerfile](https://github.com/devcontainers/templates/blob/main/src/ruby-rails-postgres/.devcontainer/Dockerfile#L8) via the `RAILS_DEVELOPMENT_HOSTS` env var ([PR](https://github.com/rails/rails/pull/41560)), and `"ghiculescu-urban-lamp-asdfasdgsafda-3000.preview.app.github.dev"` is the (lightly sanitized) URL of my codespace.

It looks like because it is a sub-subdomain it isn't permitted. I mentioned this to someone else who regularly does Rails development on codespaces and he said "oh yeah you have to add config.hosts", which is fine, but given there was a lot of work put into making it work out of the box I thought you might want to know.

Thanks for all your work on Codespaces!

cc @dmilburn

ghiculescu:
I wrote all that post and then I realised the fix was just to add this config:

```ruby
config.hosts << ".preview.app.github.dev"
```

Perhaps that could be added to the template?

joshaber:
@ghiculescu great catch, thank you for pointing it out! I've opened https://github.com/devcontainers/templates/pull/83 and https://github.com/devcontainers/images/pull/262 to resolve it.

Issue #1589, opened 2022-08-19
[Announcement] August 2022 update: Dev container community contributions


samruddhikhandale:
### Latest context

Hi dev container community! We appreciate your continued patience as we work to make dev containers a great experience and open specification, and we're excited to share with you the latest plan for community contributions (which we've been crafting since https://github.com/microsoft/vscode-dev-containers/issues/1291).

### New repos

As mentioned, we recently announced the [dev container specification](https://containers.dev). As part of this, we now have a [devcontainers GitHub org](https://github.com/devcontainers), and we place repos related to the spec in this org.

We recently open sourced two new `devcontainers` repos:
- https://github.com/devcontainers/features
- https://github.com/devcontainers/images

As shown in the repos, we are hosting a specific set of features and images that were previously in this repository. Please log issues in this new location.

Also note that we will soon also have a repo for a subset of dev container definitions (a.k.a. templates) in the `containers` folder of this repository. We expect the distribution model for templates to be similar to Features - but this is not quite ready.

### How can I contribute or create new content?

While you could always publish your own dev container image, you can now publish Features as well! This means you can make updates and fixes at any cadence you like! To get started, use this [GitHub Repository Template](https://github.com/devcontainers/feature-template) to create one. It includes a GitHub Actions workflow to automate the publishing process. If you want more details, you can read more about the [features portion of spec at containers.dev](https://containers.dev/implementors/features).

If you've contributed a definition/template in the past, consider whether a Dev Container Feature may be better. Features allow your contribution to be added to any supported base image, rather than just the one in the template.

If you still want to contribute a definition/template in addition or instead, stay tuned for information on how you can self-publish these as well!

If you'd like your contributions to appear in VS Code or GitHub Codespaces UI for dev container creation, you may do the following soon (this is still a work in progress):
- Go to https://github.com/devcontainers/devcontainers.github.io
- Open a PR to modify the `feature-collections.txt` file, and add your `<registry>/<namespace>` to the end of the file

### What happens to this vscode-dev-containers repo?

We will ultimately be archiving this repository and focusing new work on the repos in the `devcontainers` org. We will not delete or hide this repo. However, we won't archive the repository until we've migrated the dev container definitions / templates as well. We expect the general model for distributing templates to be the same as Features - so stay tuned for more information on that!

### Feedback

[...]

PR #1650, opened 2022-10-05
Add docker-in-docker Azure DNS override flag

kklopfenstein:
By default the docker-in-docker script passes an Azure specific IP address to dockerd that causes Docker not to use the devcontainer's resolv.conf settings. We would like to not override DNS within Docker in some situations.

This PR adds a new `AZURE_DNS_OVERRIDE` flag that will give us the ability to disable this Azure DNS setting.

### DIFF
--- a/script-library/docker-in-docker-debian.sh
+++ b/script-library/docker-in-docker-debian.sh
@@ -7,14 +7,14 @@
 # Docs: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/docker-in-docker.md
 # Maintainer: The VS Code and Codespaces Teams
 #
-# Syntax: ./docker-in-docker-debian.sh [enable non-root docker access flag] [non-root user] [use moby] [Engine/CLI Version] [Major version for docker-compose] [azure DNS override]
+# Syntax: ./docker-in-docker-debian.sh [enable non-root docker access flag] [non-root user] [use moby] [Engine/CLI Version] [Major version for docker-compose] [azure DNS auto detection flag]
 
 ENABLE_NONROOT_DOCKER=${1:-"true"}
 USERNAME=${2:-"automatic"}
 USE_MOBY=${3:-"true"}
 DOCKER_VERSION=${4:-"latest"} # The Docker/Moby Engine + CLI should match in version
 DOCKER_DASH_COMPOSE_VERSION=${5:-"v1"} # v1 or v2
-AZURE_DNS_OVERRIDE=${6:-"true"}
+AZURE_DNS_AUTO_DETECTION=${6:-"true"}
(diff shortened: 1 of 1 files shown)


CleanCut (approved):

script-library/docker-in-docker-debian.sh
```diff
@@ -7,13 +7,14 @@
 # Docs: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/docker-in-docker.md
 # Maintainer: The VS Code and Codespaces Teams
 #
-# Syntax: ./docker-in-docker-debian.sh [enable non-root docker access flag] [non-root user] [use moby] [Engine/CLI Version] [Major version for docker-compose]
+# Syntax: ./docker-in-docker-debian.sh [enable non-root docker access flag] [non-root user] [use moby] [Engine/CLI Version] [Major version for docker-compose] [azure DNS override]
```
samruddhikhandale, line 10: Hi, thanks for the PR. 

However, we have migrated the Features to the [devcontainers/features](https://github.com/devcontainers/) repo. More details in the [announcement issue.](https://github.com/microsoft/vscode-dev-containers/issues/1589)

Do you mind opening the PR in [docker-in-docker](https://github.com/devcontainers/features/tree/main/src/docker-in-docker) Feature instead?

script-library/docker-in-docker-debian.sh
```diff
@@ -325,6 +326,10 @@ tee /usr/local/share/docker-init.sh > /dev/null \
 
 set -e
 
+AZURE_DNS_OVERRIDE=$AZURE_DNS_OVERRIDE
+EOF
+
+tee -a ./docker-init.sh > /dev/null << 'EOF'
```
joshspicer, line 332: nit: Could we keep the style of this `tee` command the same as the one above (same line breaks, using the full `/usr/local/share/docker-init.sh` file path)

joshspicer (commented):
Thank you for improving our scripts! 

As @samruddhikhandale mentioned, it would be great to additionally put up a PR on the [associated Feature](https://github.com/devcontainers/features/tree/main/src/docker-in-docker), as we have deprecated these scripts in favor of the new Features repo.  The script in the Features repo should be nearly identical.

In the Features repo, you'd want to update both the [install script](https://github.com/devcontainers/features/tree/main/src/docker-in-docker) and add in a new [`AzureDnsOverride` option](https://github.com/devcontainers/features/blob/main/src/docker-in-docker/devcontainer-feature.json#L7) so that we can surface this flag in tooling

script-library/docker-in-docker-debian.sh
kklopfenstein: Here's the features PR: https://github.com/devcontainers/features/pull/200

As discussed, we would like to make this change here first and follow up with the features change.

PR #1659, opened 2022-10-10
Updates community templates/definitions

samruddhikhandale:
Updates community templates to follow the [spec](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-templates.md).

All other templates are now hosted from the https://github.com/devcontainers/templates repository.

### DIFF
--- a/containers/azure-terraform/.devcontainer/devcontainer.json
+++ b/containers/azure-terraform/.devcontainer/devcontainer.json
@@ -5,10 +5,7 @@
 		"args": { 
 			"TERRAFORM_VERSION": "0.14.5",
 			"TFLINT_VERSION": "0.24.1",
-			"TERRAGRUNT_VERSION": "0.28.1",
-			"INSTALL_AZURE_CLI": "true",
-			"INSTALL_DOCKER": "true",
-			"NODE_VERSION": "lts/*"
+			"TERRAGRUNT_VERSION": "0.28.1"
 		}
 	},
 	"mounts": [ "source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind" ],
@@ -55,6 +52,6 @@
 	// Use 'postCreateCommand' to run commands after the container is created.
 	// "postCreateCommand": "terraform --version",
 
-	// Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
+	// Set `remoteUser` to `root` to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
 	"remoteUser": "vscode"
 }
--- a/containers/azure-terraform/devcontainer-template.json
+++ b/containers/azure-terraform/devcontainer-template.json
@@ -1,0 +1,48 @@
+{
+	"id": "azure-terraform",
+	"version": "1.0.0",
+	"name": "Azure Terraform",
+	"description": "_build, change, and version Azure infrastructure with terraform_",
+	"documentationURL": "https://github.com/microsoft/vscode-dev-containers/tree/main/containers/azure-terraform",
+	"publisher": "Community",
+	"licenseURL": "https://github.com/microsoft/vscode-dev-containers/blob/main/LICENSE",
+	"options": {
+		"installZsh": {
+			"type": "boolean",
+			"description": "Install ZSH!?",
+			"default": "true"
+		},
+		"upgradePackages": {
+			"type": "boolean",
+			"description": "Upgrade OS packages?",
+			"default": "false"
+		},
+		"installAzureCli": {
+			"type": "boolean",
+			"description": "Install Azure CLI!?",
+			"default": "true"
+		},
+		"installDockerCli": {
+			"type": "boolean",
+			"description": "Install Docker CLI!?",
+			"default": "true"
+		},
+		"nodeVersion": {
+			"type": "string",
+			"description": "Node.js version:",
+			"proposals": [
+				"none",
+				"lts/*",
+				"16",
+				"14",
+				"12",
+				"10"
+			],
+			"default": "lts/*"
+		}
+	},
+	"platforms": [
+		"Azure",
+		"Terraform"
+	]
+}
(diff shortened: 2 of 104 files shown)


eitsupi:
Hello. I was expecting this repository to be archived and was preparing to move the R definition to @rocker-org.
(So I am surprised by this PR)

Should I continue to keep the R definition on this repository?
Or should we create and maintain a new template repository in @rocker-org and deprecate the R definitions here?
My only point of contention is whether or not the template can be easily called up on the VSCode UI, and I will continue to maintain it in either case.

Thanks.

eitsupi:
Thank you for your explanation. It makes sense to me.
And thank you for doing so much work so that this does not have to be an abrupt transition!

containers/azure-terraform/devcontainer-template.json
```diff
@@ -0,0 +1,48 @@
+{
+	"id": "azure-terraform",
+	"version": "1.0.0",
+	"name": "Azure Terraform",
+	"description": "_build, change, and version Azure infrastructure with terraform_",
+	"documentationURL": "https://github.com/microsoft/vscode-dev-containers/tree/main/containers/azure-terraform",
+	"publisher": "Community",
+	"licenseURL": "https://github.com/microsoft/vscode-dev-containers/blob/main/LICENSE",
+	"options": {
+		"installZsh": {
```
Chuxel, line 10: Given the base image has ZSH and package upgrades in it, we could probably drop the installZsh and upgradePackages options here.

We also have a terraform feature, so technically this definition could just flip to using that.

containers/azure-terraform/devcontainer-template.json
```diff
@@ -0,0 +1,48 @@
...
+		"installDockerCli": {
+			"type": "boolean",
+			"description": "Install Docker CLI!?",
+			"default": "true"
+		},
+		"nodeVersion": {
+			"type": "string",
+			"description": "Node.js version:",
+			"proposals": [
+				"none",
```
Chuxel, line 34: Old list ... lts/*, 18, 16, 14 is current

containers/clojure/.devcontainer/devcontainer.json
```diff
@@ -3,17 +3,8 @@
...
-			// Use the -bullseye variants on local arm64/Apple Silicon.
-			"VARIANT": "17",
 			// Options
-			"CLOJURE_VERSION": "1.10.3",
-			"INSTALL_CLOJURE_CLI": "true",
-			"INSTALL_BOOT": "true",
-			"INSTALL_LEININGEN": "true",
-			"INSTALL_POLYLITH": "true",
-			"NODE_VERSION": "lts/*"
+			"CLOJURE_VERSION": "1.10.3"
```
Chuxel, line 7: Why not make this an option as well?

containers/azure-terraform/.devcontainer/devcontainer.json
```diff
@@ -5,10 +5,7 @@
 		"args": { 
 			"TERRAFORM_VERSION": "0.14.5",
 			"TFLINT_VERSION": "0.24.1",
-			"TERRAGRUNT_VERSION": "0.28.1",
-			"INSTALL_AZURE_CLI": "true",
-			"INSTALL_DOCKER": "true",
-			"NODE_VERSION": "lts/*"
+			"TERRAGRUNT_VERSION": "0.28.1"
```
Chuxel, line 8: Why not make these options as well?

containers/perl/devcontainer-template.json
```diff
@@ -0,0 +1,24 @@
+{
+	"id": "perl",
+	"version": "1.0.0",
+	"name": "Perl",
+	"description": "Develop Perl applications on Linux",
+	"documentationURL": "https://github.com/microsoft/vscode-dev-containers/tree/main/containers/perl",
+	"publisher": "Community",
+	"licenseURL": "https://github.com/microsoft/vscode-dev-containers/blob/main/LICENSE",
+	"options": {
```
Chuxel, line 9: We can probably drop these given they're in the base image now. This predates the base I believe.

Chuxel (commented):
There are a few definitions where we have left build args in place, while in others they've been converted to options. Wouldn't it make sense to be consistent?

Chuxel (approved):
Actually - taking a second look, the issues I pointed out were in the original templates as well, which likely says that we should not address them in this PR which focuses just on just enabling OCI-based publishing.

eitsupi:
Sorry to keep asking this question, but would you be able to provide us a "template of templates"?
In other words, should I start building a template repository now based on https://github.com/devcontainers/templates?
Or should I wait for a template of devcontainer templates like https://github.com/devcontainers/feature-template?

bamurtaugh:
Thanks @eitsupi, this is a good topic! 

We've opened https://github.com/devcontainers/spec/issues/116 to track creating such a repo. However, we do hope the content in https://github.com/devcontainers/templates is sufficient to start self-hosting your templates now, so please feel free to start creating a repo, and you can always let us know if you have any questions or feedback.

eitsupi:
Thank you for the clarification!
Yes, no doubt your repositories provide me with enough information.

Will work on that as soon as possible.

cc @cboettig

containers/haskell/.devcontainer/Dockerfile
```diff
@@ -3,9 +3,9 @@ FROM debian:bullseye-slim
 ENV LANG C.UTF-8
 
 # [Option] Install zsh
-ARG INSTALL_ZSH="true"
+ARG INSTALL_ZSH="${templateOption:installZsh}"
```
josephcsible, line 6: The Stack Overflow question [unable to build VSC devcontainer for Haskell on OS X Ventura](https://stackoverflow.com/q/74929464/7509065) appears to be related to this change.

samruddhikhandale: Thanks! Replied my thoughts on the thread.

PR #1682, opened 2022-10-31
Fix: "E: Unable to locate package manpages-posix"

ZarTek-Creole:
If ${ADD_NON_FREE_PACKAGES} is true and Debian is "stable"

The replacement with sed does not work because it does not find the code name (VERSION_CODENAME) , it must find and replace the value "stable" instead of "${VERSION_CODENAME}" retrieve from /etc/os- release


ZarTek-Creole:
@microsoft-github-policy-service agree

PR #1708, opened 2023-01-03
fix: Add -e for echo newlines to resolve properly

killerHELIX:
Fixes a typo in the docker-from-docker docs.


Chuxel:
FYI - these have been moved to https://github.com/devcontainers/templates and https://github.com/devcontainers/features going forward.

But either way, thanks for the contribution!

Issue #1762, opened 2023-08-01
[Announcement] This repo will be archived November 2023


bamurtaugh:
> **TLDR; With the growing [dev container spec](https://containers.dev/), our core dev container work has moved to the [dev containers org on GitHub](https://github.com/devcontainers). Thus, we will archive this vscode-dev-containers repo in November 2023.**

👋 Hi dev container community! 

Opening the [dev container specification](https://containers.dev/) has been an incredible journey over the past year. We've implemented key changes to how and where we host and accept contributions to dev container resources, which we described in an earlier announcement issue: https://github.com/microsoft/vscode-dev-containers/issues/1589.

## Archiving this repo

We want to ensure everyone gets the latest resources (dev container images, Features, and Templates), which are all hosted in the [dev containers org on GitHub](https://github.com/devcontainers), or through our [community index](https://containers.dev/collections). The spec has reached a point where we will no longer use or reference this vscode-dev-containers repo.

**To avoid confusion and an additional maintenance burden, we will archive this repository in November 2023.**

## What does archiving mean to me?

If you've already been using the latest dev container resources in the dev containers org on GitHub, this change should have no impact on you!

If you're still using any of the resources in this repo, please migrate to using resources in the dev containers org instead by November 2023. If you'd like to contribute a definition from this repo that isn't yet in our [index](https://containers.dev/templates), please feel free to publish one and contribute it to our index. We will not be deleting this repo though, so you can still review it if necessary after November.

## Closing - thank you!

Your involvement, feedback, contributions to, and use of dev containers has been instrumental in us opening up the dev container spec - THANK YOU for everything! We can't wait to see how our community continues to grow and collaborate!

If you're looking to get started with dev containers or filing an issue / feature request, here are some places to get started:

* Dev Container images managed by the Dev Container spec maintainers: https://github.com/devcontainers/images
* Dev Container Features managed by the Dev Container spec maintainers: https://github.com/devcontainers/features
* Dev Container Templates managed by the Dev Container spec maintainers: https://github.com/devcontainers/templates
* For new Definitions/Templates, see https://github.com/devcontainers/template-starter to get started and add your Template to the community index
* For new Dev Container Features, see https://github.com/devcontainers/feature-template to get started and add your Feature to the community index

bamurtaugh:
Hi @papakpmartin, thanks for reaching out. The repos above are accurate and will provide the latest resources provided by the spec maintainers and community.

[...]

PR #1783, opened 2023-11-15
Stops publishing and supporting Templates

samruddhikhandale:
With respect to https://github.com/microsoft/vscode-dev-containers/issues/1762, we are archiving the [microsoft/vscode-dev-containers](https://github.com/microsoft/vscode-dev-containers) repo. However, it still contains code for [community contributed Templates](https://github.com/microsoft/vscode-dev-containers/tree/main/containers) which are published to the index. In https://containers.dev/templates, we do mark it as` Deprecated (No longer maintained)`, hence, now we will stop publishing them as they will no longer be supported.


bamurtaugh (approved):
Thank you!

lramos15 (approved):

